PR #1406, opened 2018-03-06
Regenerated allen atlases

cmungall:
Note there is a spurious diff in here - it makes it look like some IDs were deleted; just order shuffling (sorry, should have normalized ordering)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1360,14 +1360,8 @@ bto-anat.obo:
 # ABA
 # ----------------------------------------
 
-ALLENS = dmba hba dhba pba mba
+ALLENS = dmba hba dhba pba mba glioblastoma
 
-
-
-aba.obo: ABA-src.obo
-	./util/make-aba-part-ofs.pl $< > $@
-bridge/aba.owl: aba.obo
-	owltools $< -o file://`pwd`/$@
 
 allen_all: $(patsubst %,source-ontologies/allen-%.obo,$(ALLENS))
 source-ontologies/allen-dmba.json:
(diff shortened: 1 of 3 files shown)


matentzn:
Who would be an appropriate reviewer for this PR? @cmungall

cmungall:
@tgbugs or @dosumis. I simply forget what the current status is with IDs and the Allen Atlases

tgbugs:
The MBA changes look ok to me. The ids are more or less completely arbitrary at this point (much to the consternation of some atlas implementers), still no change to the fact that all the ontology ids share the same uri prefix in the Allen api (http://api.brain-map.org/api/v2/data/Structure/).

matentzn:
If to be moved forward needs to be aligned with new Makefile structure. Can help but needs driver.

dosumis:
ABA Uberon mappings are important for the Brain Data Standards grant, so work can be done @hkir-dev (+ Shawn & Allen representatives if needed). This are currently lower on our priority list.  We should organise a call to work out what needs to be done.  Can be after summer.

gouttegd:
Presumably we can close here?

For all 5 Allen atlases (HBA / DHBA / MBA / DMBA / PBA) that we include in `composite-metazoan`, we now use the pre-built ontologies provided by https://github.com/brain-bican, so there’s no more need to re-generate them ourselves.

As for the addition of the “glioblastoma” atlas:

* Is it really something that should be included in `composite-metazoan`? My feeling is that Uberon (and its composite products) should only be about “normal” / “wild-type” anatomy, and that describing pathological anatomical structures does not belong here.
* In this PR, `allen-glioblastoma.obo` is constructed exactly as `allen-hba.obo` – both files are constructed from the same JSON file: http://api.brain-map.org/api/v2/structure_graph_download/10.json – I assume this is a mistake?

PR #1429, opened 2018-05-13
Map worm anatomy to anatomical entity. 

balhoff:


balhoff:
Please make a release after.

PR #1467, opened 2018-11-20
Hyperlink DOIs to preferred resolver (safely possible within .obo files?)

katrinleinweber:
Hi! This is essentially the same PR as [OBOFoundry.github.io#769](https://github.com/OBOFoundry/OBOFoundry.github.io/pull/769), but I wanted to update the [perl code that generates new DOI links](https://github.com/obophenotype/uberon/compare/master...katrinleinweber:resolve-DOIs-securely#diff-d930c50cd28bdd548fc3de2156dae3e4).

Reviewing the changes in the `.obo` files now, I'm wondering whether updating the PIDs' URL prefix / resolver domain (not sure what the correct term is here) can be done like this, without problems. The [`EXACT` statements have me worried](https://github.com/obophenotype/uberon/compare/master...katrinleinweber:resolve-DOIs-securely#diff-84a8ee098f45ec7de5cd28cdfe9dc322L93223), but since I just started reading into this whole vocabulary & ontology topic, please excuse any noob-ness in the question.

Shall I exclude the `.obo` files from this PR again?


katrinleinweber:
Cool, thanks for explaining and merging :-)

PR #1493, opened 2019-04-16
UBERON:0013529 ! Broadmann area remove bad xref

tgbugs:
This is one of many terms that have been tainted with the madness
spawned from the NeuroNames id switch. I have corrected the NeuroNames
id to match the correct id at the present date, however the incorrect
mapping lives on in all the rest of NIFSTD's ontology holdings. We will
eventually get around to correcting all the mappings for these but for
the time being the best thing to do is probably to take a scorched
earth approach and remove any synonyms that have been pulled in from
a NeuroNames xref that came from NIFSTD.


cmungall:
Thank you! I thought I had corrected for everything involving the insane neuronames ID shift, but apparently not...

tgbugs:
Found another one in UBERON:0001869. @tmsincomb is working on hunting down all of these.

PR #1496, opened 2019-05-07
OBO 1.4 syntax and Quality-of-Life improvements

althonos:
Hi!

This PR addresses some issues in the uber-anatomy ontology to make it compliant with the OBO 1.4 specification. It also brings some suggestions to improve the contents and interoperability of the ontology itself. (cc @cmungall)

## Required changes

- [x] Fix invalid whitespaces in xrefs caused by definitions added as xrefs to some terms.

## Suggested changes

- [ ] Use PMID and ISBN for publicationsm using dbxref optional descriptions to add the informal name (e.g. `AAO:Pugener_and_Maglia_2008` would become `PMID:18946872 "Pugener and Maglia (2008)"`)
- [ ] Remove prefixes without semantic value that are only used as description, and use the description syntax instead (e.g. `Biology-Online_Dictionary:http\://www.biology-online.org/dictionary/` should be `http://www.biology-online.org/dictionary/` or `http://www.biology-online.org/dictionary/ "Biology Online Dictionary"`)
- [ ] Have a consistent way of managing DOI: currently some are being added as prefixed IDs (with either the `DOI` or `doi` prefix), while other are added as plain URLs to either `doi.org` or `onlinelibrary.wiley.com/doi`.
- [ ] Use IDspaces to convert some thoroughly used external resources to CURIES:
  - [ ] BrainInfo
  - [ ] Wikipedia (this is the GO convention)
  - [ ] Free Dictionary Medical Dictionary
  - [ ] Neurolex (they have their own URI system as well that would work fine with that)
  - [ ] Birnlex (same system, different URIs)
  - [ ] Medilexicon
  - [ ] PMID to replace `ncbi.nlm.nih.gov` links!
- [ ] Use an IDspace for Dublin Core annotations and add a `treat-xrefs-as-equivalent-to: dc` clause to the header for the `dc` typedefs to be mapped as `dc` annotations when converting to OWL
- [ ] Remove or fix dead `drugs.com` links (`drugs.com/dict` is not resolving anymore)
- [ ] Improve consistency of ISBNs: the advised way is to have all ISBNs under the `ISBN` prefix, in hyphenated form. This way, they can be expanded to `urn:isbn:<id>` which is the semantically correct ISBN URN.
- [ ] Figure out an alternative for the unreachable `NIFSTD` idspace.
- [ ] Use ORCIDs for people (*I know this won't happen because it's not practical to use them, but the current use of common term prefixes to refer to people can lead to invalid URI expansion*).

## Status

This is a *Work In Progress* PR, discussion is needed to add some of the suggested changes, but I will ultimately implement all of these.


althonos:
@cmungall : good thing I put a message stating that was a WIP, I wanted to add some more commits before the merge :sweat_smile:

tgbugs:
Very glad to see this. I updated our internal obo tools to match the 1.4 spec and we were inducing major changes in uberon when trying to use them to fix some issues that originally came from NIFSTD, so hopefully those won't be an issue now. @tmsincomb

PR #1502, opened 2019-07-04
Hyperlink DOIs to preferred resolver

katrinleinweber:
Same as https://github.com/obophenotype/developmental-stage-ontologies/pull/63 ;-)


cmungall:
These are ephemeral diffs that probably shouldn't even be checked in.. I will not merge this since the diff is an accurate record of the changes that happened.. but let me see if I can figure out where this is coming from, as we should be expanding to the proper/new DOI URLssss

PR #1537, opened 2019-11-26
DO NOT MERGE: Question about idspaces

matentzn:
This diff contains the declaration of three idspace declarations. When I add them to the obo file manually (with text editor, using the Protege55 normalised version) and then save it again as OBO, they disappear.

Is this a problem? A bug? @cmungall @balhoff


balhoff:
This doesn't seem to be implemented in the OBO library. It is in the 1.4 spec though.

cmungall:
This is a deeply unfortunate fact, working with non-OBO URIs in obo format files is hard, requiring really awful post-processing. The correct thing to do is just to fix this in the owlapi, we have not assigned anyone to this but the resource cost of the workarounds has probably surpasses what this iinvestment would have been

matentzn:
Uberon editors: This will probably never be merged, so you should close it. Its good to have a record of this though to know its been tried.

matentzn:
Closed as per comment above.

PR #1564, opened 2020-02-14
Add anatomical lines

megbalk:
added the following lines:
line from medial   condyle of femur to greater trochanter of femur
line from medial   condyle of femur to head of femur
line from trochlea   of humerus to proximal head of humerus
line from trochlea   of humerus to ventral tubercle of humerus
line from helix of   outer ear to intertragus incisure


megbalk:
added the following:
molar 4
lower fourth secondary molar tooth
upper fourth secondary molar tooth
premolar 3
lower third secondary premolar tooth
upper third secondary premolar tooth

PR #1567, opened 2020-02-18
seeing if changes show up now

megbalk:
Changes I had saved were not being shown in pull request for fovt-3


balhoff:
@ramonawalls @megbalk I think you're seeing a limitation of storing in OBO format. There is this equivalence axiom:

```
EquivalentClasses(
<http://purl.obolibrary.org/obo/UBERON_7500013> 
ObjectIntersectionOf(
	<http://purl.obolibrary.org/obo/UBERON_0006800> 
	ObjectSomeValuesFrom(
		<http://purl.obolibrary.org/obo/RO_0002176> 
		<http://purl.obolibrary.org/obo/UBERON_0009984>
	) 
	ObjectSomeValuesFrom(
		<http://purl.obolibrary.org/obo/RO_0002176> 
		ObjectIntersectionOf(
			<http://purl.obolibrary.org/obo/UBERON_0002503> 
			ObjectSomeValuesFrom(
				<http://purl.obolibrary.org/obo/BFO_0000050> 
				<http://purl.obolibrary.org/obo/UBERON_0000981>
			)
		)
	)
)
)
```

The second ObjectSomeValuesFrom is chained too deep for OBO to represent. If there is a simpler way to describe this, you could do that, or maybe @cmungall can suggest a different approach. We do have another Uberon component stored in OWL Functional syntax at https://github.com/obophenotype/uberon-phenoscape-ext. It could handle such expressions, but in general we are hoping to reunify these repositories in the near future. The uberon-phenoscape-ext is merged into Uberon when Uberon is built.

ramonawalls:
Thanks @balhoff. So, would it be okay for us to make our changes in the phenoscape extension, or should we start another one? We are coping patterns that are already in uberon, so someone must have made it work before.

PR #1568, opened 2020-02-19
GO:cell part is obsoleted, this commit suggests to use CL:cell

drseb:
If I understand it correctly, GO has obsoleted 'cell part'. GO suggests to consider CL 'cell'. If we want to do this, this PR does it


balhoff:
We replaced 'cell part' in GO with usages of the expression 'part_of some cell'. So that may be needed, depending on the situation here.

drseb:
@cmungall should this be replaced by 'part_of some cell' in both cases, as @balhoff mentions?

uberon_edit.obo
```diff
@@ -91212,7 +91212,7 @@ xref: MA:0003122
 xref: MBA:10703
 intersection_of: UBERON:0002304 ! layer of dentate gyrus
 intersection_of: bounding_layer_of UBERON:0001885 ! dentate gyrus of hippocampal formation
-relationship: composed_primarily_of GO:0044464 ! cell part
+relationship: composed_primarily_of CL:0000000 ! cell
```
cmungall: this is no longer correct. The layer need not consist of whole cells.

While this could be represented by an owl class expression `composed_primarily_of some (part-of some cell)`...

the value of the axiom doesn't really warrant the complexity

remove it

for specific layers we can actually describe what the parts are - e..g synapses, somas, axons, ....

uberon_edit.obo
```diff
@@ -98409,7 +98409,7 @@ xref: FBbt:00007060
 xref: FMA:83115
 is_a: UBERON:0000477 {source="FMA"} ! anatomical cluster
 relationship: composed_primarily_of GO:0005575 ! cellular_component
-relationship: has_component GO:0044464 {minCardinality="2"} ! cell part
+relationship: has_component CL:0000000 {minCardinality="2"} ! cell
```
cmungall: this is incorrect for same reasons as the above

I think the value of the axiom is even lower, so remove

cmungall:
@balhoff's advice is correct but we can't have nested expressions in obo file. We could keep in our supplemental owl file... but really the value of these are low, let's keep it simple

cmungall:
Related PR here: #1582 (that PR is about cell, this is about cell part)

PR #1605, opened 2020-05-26
uberon_edit.obo use oboformat v1.4 tag ordering

tgbugs:
Reserialize uberon_edit.obo using pyontutils.obo_io so that the tag
ordering in various blocks matches the v1.4 format spec.
https://owlcollab.github.io/oboformat/doc/obo-syntax.html

The primary changes are in the header ordering and in the ordering of
the property_value tag.

One other slight change is that trailing qualifiers are ordered
alphabetically. In theory this helps to ensure deterministic
serialization of qualifiers.


balhoff:
@tgbugs I think the current ordering is the output of the current Protégé, which is likely to be the most common way the file is edited. Is there a way we can align these? Does a Protégé resave undo your serialization?

tgbugs:
Will take a look. I have a bug in the the reserialization that breaks Protégé's ability to parse the file at all, so will fix that and look into it. If it doesn't match what Protégé has then I will add a Protégé compatibility mode and close this PR.

tgbugs:
After fixing the other issues I was able to determin that uberon_edit.obo cannot be edited using Protege because Protege does not currently roundtrip many tag types and also does not properly
roundtrip xrefs, failing to serialize the compact version of the xref as it was originally provided, so this reserializaztion will have no effect on any Protege related workflow.

cmungall:
I'd rather do this using `make roundtrip.obo` which uses robot synced to owlapi version which minimizes spurious diffs

tgbugs:
Ah, perfect. I'll make sure to match the robot output, so I will close this. Thanks!

Issue #1619, opened 2020-06-23
Missing connections between types of reproductive system secretions


ashfinckh:
There seems to be a missed connection between the finding "female reproductive gland secretion" (UBERON:0006537) and the general finding "reproductive gland secretion" (UBERON:0022293) 

In addition to the grouping of the reproductive gland secretion findings, there is no connection between the female reproductive gland secretion finding and the finding  "vaginal fluid" (UBERON:0036243).

Thanks very much for your help.

![Screenshot 2020-06-23 at 14 22 49](https://user-images.githubusercontent.com/67323234/85407205-342d5f80-b563-11ea-9474-7bf5185500c4.png)

cmungall:
@drseb - you are best to have Elk running all times + inferred view, you should see the inference

drseb:
Thanks @ashfinckh . This is now working. Thanks @cmungall for approving the PR

PR #1621, opened 2020-06-25
adding axioms to obtain same subclass-inference for male and female reproductive system gland secretion

drseb:
related to https://github.com/obophenotype/uberon/issues/1619


uberon_edit.obo
```diff
@@ -2268,6 +2268,7 @@ xref: VHOG:0000725
 xref: WBbt:0008423
 xref: Wikipedia:Male_reproductive_system_(human)
 xref: XAO:0000155
+is_a: UBERON:0000990 ! reproductive system
```
matentzn, line 2271: Is this the only change? The rest of the diff seems just spurious (I checked it all, nothing broke, just wanna be sure).

drseb: sorry. I just saw another inconsistency. so two changes are introduced in this branch.

cmungall: also this shouldn't be necessary as we have the equiv axiom

drseb: ok. but the other axiom is necessary? I think the first change makes the female and male axioms consistent, right?

drseb: should I make a new PR?

PR #1622, opened 2020-06-25
fix axiom for 'female reproductive gland secretion'

drseb:
related to https://github.com/obophenotype/uberon/issues/1619


drseb:
I still see this difference though:

Screenshot 2020-06-25 at 21 40 19
Screenshot 2020-06-25 at 21 40 14

drseb:
And this is why I suggested this other commit in the previous PR https://github.com/obophenotype/uberon/pull/1621

cmungall:
this is an annoying structural difference but it has no effect on the output (we run robot relax), and not effect on entailments (since the subclass is entaled by the equivalence)

need more dev docs!

drseb:
In HPO I would do everything to get rid of such annoying differences :-) But if it does not affect the inferences and you are okay with it, I leave it as it is.

PR #1629, opened 2020-08-10
Basic first (incomplete) fix of Makefile

matentzn:
Problems with the release will be solved here one by one but it needs active attention, probably another hack session. The most important stuff is towards the end of the Makefile, but this pipeline was so broken, if you just check in the old version [row 1040](https://github.com/obophenotype/uberon/blob/86bf51116f5f5487a051c977d8a995d3f5c5ba09/Makefile#L1040) as an example. No idea how running that would have been possible ever. :) 

Major changes:
- Some skipped goals (end of Makefile)
- Any command involving blip commented out
- ubuntu in odk did not like  `>& $@.LOG`, so I changed it to `2>&1 > $@.LOG` (not sure that is correct).

@cmungall

Running `make all`

- [ ] since `aminoacid.owl:` is commented out (in the Makefile with a warning that 'PR is now too big') and it is needed, I commented it back in and regenerated it.
- [ ] for pr.owl, the parameter `--split-ontology -d null` does not work (cant find the `null` directory). To work around the issue, I did a `mkdir null` but probably this will cause other issues later..

```
owltools $< --extract-mingraph --rename-entity $(OBO)/pr#has_part $(OBO)/BFO_0000051 --rename-entity $(OBO)/pr#part_of $(OBO)/BFO_0000050  --make-subset-by-properties -f BFO:0000050 BFO:0000051 // --split-ontology -d null -l snap --remove-imports-declarations  --remove-dangling --set-ontology-id $(OBO)/$@ -o $@ && touch $@
```

- [ ] the `update-stages:` changed entirely because it relied on developmental-stages being cloned already. Now its cloning, running make all (to generate, in particular ssso-merged.obo).

- [ ] no sparql/object-properties.sparql, which is needed for reports/uberon_edit-object-properties.csv, which is needed

- [ ] Missing file external-disjoints.owl

The syntax for piping to file has changed a lot
- Lots of stuff added to .gitignore

xref-to-equiv.pl does not exist, skipped
bridge-check-caro.txt reveals 55 unsatisfiable classes, so needs to be skipped.

There is no goal that generates cp composite-brain.{obo,owl} $(RELDIR) ;\

make release:
cp: cannot stat 'composite-brain.obo': No such file or directory
cp: cannot stat 'composite-brain.owl': No such file or directory
cp: cannot stat 'composite-vertebrate.owl': No such file or directory
cp: cannot stat 'composite-vertebrate-*.obo': No such file or directory
cp: cannot stat 'composite-vertebrate-*.owl': No such file or directory
cp: cannot stat 'composite-metazoan-*.owl': No such file or directory
cp: cannot stat 'reference/*[0-9]': No such file or directory


cmungall:
line 1040, lol yes that was a last minute addition for the next release, obv fix

all that faffing to get a minimal version of PRO can likely be better done by robot, using a pro base file... but ideally something PRO could themselves provide. 

sorry about the non-checked in files
```
$ more sparql/object-properties.sparql
prefix owl: <http://www.w3.org/2002/07/owl#>

SELECT DISTINCT ?p WHERE {?r owl:onProperty ?p}
```

cmungall:
external disjoints: https://github.com/obophenotype/uberon/blob/master/external-disjoints.obo

matentzn:
These have hopefully been addressed in #1798

PR #1638, opened 2020-09-08
Liver metabolic zones

mkeays:


uberon_edit.obo
```diff
@@ -193949,6 +193949,39 @@ intersection_of: in_left_side_of UBERON:0010709 ! pelvic complex
 property_value: implements_design_pattern https://github.com/obophenotype/uberon/wiki/Modeling-paired-structures-Design-Pattern xsd:anyURI
 created_by: http://orcid.org/0000-0001-9897-3238
 
+[Term]
+id: UBERON:8400001
+name: hepatic acinus periportal zone
+def: "The region of the hepatic acinus closest to the portal triad. Blood in this zone has the highest oxygen concentration of the three metabolic zones in the hepatic acinus." [PMID:28126520]
+synonym: "hepatic acinus zone I" EXACT []
+intersection_of: adjacent_to UBERON:0001279 ! portal triad
```
dosumis, line 193957: Needs a superclass.   I think 'organ subunit' http://purl.obolibrary.org/obo/UBERON_0000063 would work.  Not sure what the equivalentClass definition buys us and whether it will be safe (someone might define some sub-region that fits this description and you probably wouldn't want it auto-classified.  So I'd switch this all to SubClassOf

dosumis (changes requested):
Add genus + relax to SubClassOf  - see comment attached to commit.

dosumis:
Hi Maria, Could you also make sure defs and refs all have PMIDs.  Then happy to merge.   I'll start a separate ticket on adding a more specific genus term.  Do you think it would be reasonable to have 'metabolic zone of lliver' ?

mkeays:
@dosumis ahh thanks, I've added the missing PMID now. Possibly 'metabolic zone of liver', though I think more specifically they are zones of liver lobules or acinus, e.g. see  https://en.wikipedia.org/wiki/Lobules_of_liver

dosumis (approved):

PR #1647, opened 2020-09-28
Fixing TCs for slurped FBbts (UBERON:6nns)

cmungall:


dosumis (approved):
Looks reasonable.   Not so sure how useful the developmental terms will be, but good to have tagmata & arthropod nervous system terms.  What segments make up the head tagma in arthropods varies with taxon (and is controversial) so might consider going more specific with terms on that.

PR #1705, opened 2020-12-21
Protege 5.5 Normalisation PR

matentzn:
Someone must have used an old owltools, protege or ROBOT to edit uberon recently;

This PR adds nothing, just normalises the serialisation to Protege 5.5


matentzn:
Btw @cmungall seems like its your automatic process is the one that generates the different serialisation.. Wherever the HUMAN_PREFERRED obo annotations come from, and the FMA xrefs. Maybe run through ROBOT convert once after you run pipelines?

Issue #1657, opened 2020-10-23
duodeno-jejunal junction


ngeorgeebi:
**Preferred term label:**
duodeno-jejunal junction

**Synonyms**
duodenojejunal flexure, duodeno-jejunal flexure, duodenojejunal junction

**Definition (free text, please give PubMed ID)**
(Derived from textbook Gray's Anatomy)
The duodenojejunal (DJ) flexure or junction is the anatomical border between the duodenum and the jejunum.

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
Duodenum
http://purl.obolibrary.org/obo/UBERON_0002114

**Your nano-attribution (ORCID)**
0000-0003-4183-8865

paolaroncaglia:
cc @ngeorgeebi 

The requested term 'duodeno-jejunal junction' seems in scope for Uberon to me. It is referred to in the definition of UBERON:0000642. If it’s approved, I’d add it as follows:

duodeno-jejunal junction

“The anatomical border between the duodenum and the jejunum.”

subclass of organ part
part of small intestine (not duodenum)

Taxon constraint: the taxon notes for duodenum and jejunum suggest that the new term might potentially be applicable to some higher vertebrates beyond mammals. However PMID: 29064554 “…showed the anatomical similarities and dissimilarities of the duodenojejunal flexure among the mammals”. The term is used for mouse though (PMID: 23961897). I’d welcome feedback from others please. 

Dbxrefs:
https://en.wikipedia.org/w/index.php?title=Duodenojejunal_flexure&oldid=937307798
TA98:A05.6.02.009 (is this the correct format?)
TA2:2952 (is this the correct format?)
FMA:15957
https://orcid.org/0000-0003-4183-8865

synonyms: 
duodenojejunal flexure
duodeno-jejunal flexure
duodenojejunal junction
flexura duodenojejunalis (latin term)

Thanks,
Paola

paolaroncaglia:
@ngeorgeebi 
I created UBERON:8410000 duodeno-jejunal junction, it will be in the next Uberon release (there isn't a precise scheduled date yet, but I will ask at the editors' call on Monday). Thanks.

PR #1707, opened 2020-12-22
Add new class UBERON:8410000 duodeno-jejunal junction

paolaroncaglia:
Addresses #1657


paolaroncaglia:
@dosumis 
Could you please confirm if I used the 'Latin term' synonym property correctly? Thanks.

uberon_edit.obo
```diff
@@ -195227,6 +195227,27 @@ relationship: dc-creator http://orcid.org/0000-0003-2034-601X
...
 property_value: http://purl.org/dc/elements/1.1/date 2020-10-05T15:52:22Z xsd:dateTime
 
+[Term]
+id: UBERON:8410000
+name: duodeno-jejunal junction
+def: "The anatomical junction between the duodenum and the jejunum." []
+synonym: "duodeno-jejunal flexure" EXACT []
+synonym: "duodenojejunal flexure" EXACT []
+synonym: "duodenojejunal junction" EXACT []
+synonym: "flexura duodenojejunalis" EXACT [] {http://purl.obolibrary.org/obo/uberon/core#LATIN=""}
```
dosumis, line 195237: I think EXACT is fine, but that LATIN should be a synonym_type.  Did you check other examples in Protege?

dosumis (changes requested):
pubs in xrefs -> def xrefs

uberon_edit.obo
paolaroncaglia: @dosumis 

> I think EXACT is fine, but that LATIN should be a synonym_type. Did you check other examples in Protege?

I see two annotation properties, "LATIN" and "latin term" (only the latter is a subclass of synonym_type_property), and there are no annotation property usage examples for either (except for my term duodeno-jejunal junction). That's weird as I _think_ I've seen at least one of those properties used. I'll change the AP to "latin term" as a synonym_type_property, and I'll add to the agenda for the next CL/Uberon meeting.

paolaroncaglia:
@dosumis 
I addressed your comments, including the latin term, even though that's not showing up as a diff. Here's what it looks like in Protege for me:
Screen Shot 2021-01-14 at 11 28 20
I changed the Wikipedia version to what you suggested in https://github.com/obophenotype/uberon/pull/1718. It doesn't resolve as a live link, but I hope it's fine? There are no other examples of the same in Protege. Thanks.

uberon_edit.obo
dosumis: Yep - should be 'latin term' in Protege see
image - but this will show up as LATIN in OBO.

dosumis: There probably needs to be some clean up of cl-edit.owl to remove incorrect term - but we need to be careful:  Protege bug means usage is not shown.

paolaroncaglia: > There probably needs to be some clean up of cl-edit.owl to remove incorrect term - but we need to be careful: Protege bug means usage is not shown.

As discussed at the CL editors meeting today, this will be addressed in https://github.com/obophenotype/uberon/issues/1727. Note, the duplication is in Uberon but not CL, so no action required there.

Issue #1703, opened 2020-12-18
NTR: right colic artery


ngeorgeebi:
**Preferred term label:**
right colic artery

**Synonyms**

**Definition (free text, please give PubMed ID)**
The right colic artery may arise directly from the superior mesenteric artery (SMA) but often arises from a common trunk with the ileocolic artery or middle colic artery.  It supplies the ascending colon.  Paperback ISBN: 9780443103735

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
superior mesenteric artery: UBERON:0001182

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0003-4183-8865

ngeorgeebi:
please tag with HCA, @paolaroncaglia thanks!

paolaroncaglia:
@ngeorgeebi 
UBERON:8410023 right colic artery 
will be in the next Uberon release.

Issue #1702, opened 2020-12-18
NTR: left colic artery


ngeorgeebi:
**Preferred term label:**
left colic artery

**Synonyms**

**Definition (free text, please give PubMed ID)**
The first branch of the inferior mesenteric artery is the left colic artery. The left colic artery supplies blood flow to the distal third of the transverse colon and the entirety of the descending colon. Once diverging from the IMA, the left colic artery runs anterior to the left ureter, left internal spermatic vessels (males), and psoas major before branching into its ascending and descending segments. PubMed ID: NBK482156

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
inferior mesenteric artery : UBERON:0001183

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0003-4183-8865

ngeorgeebi:
@paolaroncaglia please tag with HCA, thank you!

paolaroncaglia:
@ngeorgeebi 
UBERON:8410022 left colic artery will be in the Uberon release - coming very soon.

PR #1714, opened 2020-12-23
Add classes UBERON:8410008 left colic artery and UBERON:8410009 right…

paolaroncaglia:
… colic artery

Addresses #1702 and #1703


paolaroncaglia:
@dosumis (for after the holidays!)
Not entirely sure about the diff. If it's any help, I duplicated the first term in Protege to obtain the second term (to keep the ORCID etc) and then edited the second term. Thanks.

dosumis (changes requested):
Def xrefs in regular xrefs.  Please move.  (check fail caused by missing definition ref)

dosumis (commented):
Also wondering how this got in: AnnotationAssertion(<http://www.geneontology.org/formats/oboInOwl#id> <http://purl.obolibrary.org/obo/UBERON_8410009> \"UBERON:8410008\"^^xsd:string)\n)

Probably doesn't do any harm, but could delete (will appear as annotation `id: UBERON:8410008` in Protege

paolaroncaglia:
AI for me: delete the id annotation UBERON:8410008

paolaroncaglia:
@dosumis 
I edited the dbxrefs in both classes, and removed the extra id in UBERON_8410009.

dosumis (approved):

Issue #1665, opened 2020-11-09
add term Fimbriae of uterine tube


mshadbolt:
**Preferred term label:**
fimbriae of uterine tube

**Synonyms**
fimbriae of fallopian tube

**Definition (free text, please give PubMed ID)**
"The fimbriae (singular fimbria) is a fringe of tissue around the ostium of the Fallopian tube, in the direction of the ovary. "

https://en.wikipedia.org/wiki/Fallopian_tube#Parts Not exactly sure how to find a good pubmed definition

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
I think fallopian tube http://purl.obolibrary.org/obo/UBERON_0003889 but perhaps it could be considered part of the 'uterine tube infundibulum'.

**Your nano-attribution (ORCID)**
0000-0001-7505-5418

tag @paolaroncaglia so she is aware

paolaroncaglia:
Hi @mshadbolt,

I agree that the term doesn’t seem to exist in Uberon. However, I believe that it would need a taxon restriction to mammals.
As for parentage, based on the same Wikipedia entry that you cite, I’d suggest part_of ‘uterine tube infundibulum’ and subclass_of ‘subdivision of oviduct’.
Not sure if fimbriae could be connected to the GO term ‘ovulation from ovarian follicle’, possibly by ‘involved_in’?

paolaroncaglia:
Note for self: left to do:
- [ ] Add subset tag "added_for_HCA" to UBERON:8410010 fimbria of uterine tube

paolaroncaglia:
Note for self:
- [ ] Fix synonym type property, see https://github.com/obophenotype/uberon/pull/1770#pullrequestreview-584219190

paolaroncaglia:
@mshadbolt 
UBERON:8410010 fimbria of uterine tube
(with synonyms fimbriae of uterine/fallopian tube)
will be in the next Uberon release.

PR #1717, opened 2021-01-08
Add UBERON:8410010 fimbria of uterine tube

paolaroncaglia:
Addresses #1665


dosumis (changes requested):
Missing def xref.  Wikipedia xref should be curie.

paolaroncaglia:
@dosumis 

> Missing def xref. Wikipedia xref should be curie.

Both should be fixed now, thanks.

dosumis (approved):

PR #1723, opened 2021-01-15
Add HCA subset tag to UBERON:8410010

paolaroncaglia:
Addresses #1665


paolaroncaglia:
@dosumis 
Now that the HCA subset tag is available in Uberon, I've added it to Marion's requested term

paolaroncaglia:
@dosumis 
Could you review please? Thanks.

Issue #1727, opened 2021-01-20
Duplicate 'latin' annotation properties


dosumis:
Uberon has two APs for latin synonym types.  Only one is a synonym type property.  Delete the one that is not.

See https://github.com/obophenotype/uberon/pull/1707

dosumis:
It turns out that these are coming from CL - where they have slightly different IRIs to the Uberon ones due to a defect in the OBO2OWL conversion.   The solution is, I think, to use the Uberon IRIs in both cases - I will make a ticket to this effect for CL.

paolaroncaglia:
@dosumis 

> It turns out that these are coming from CL - where they have slightly different IRIs to the Uberon ones due to a defect in the OBO2OWL conversion. The solution is, I think, to use the Uberon IRIs in both cases - I will make a ticket to this effect for CL.

Thanks for looking into this. I'll close this ticket then as a duplicate of https://github.com/obophenotype/cell-ontology/issues/893.

PR #1750, opened 2021-01-22
Delete duplicate APs: latin term and plural term

paolaroncaglia:
Addresses #1727


paolaroncaglia:
@dosumis 
A few notes:
I switched from fork to branch, let's see if I did it correctly;
Under the annotation property 'synonym_type_property', I noticed another duplicate, and I deleted that too;
I could not see (due the Protege bug you mentioned) if the deleted APs had been used anywhere;
I never deleted an AP before, so let's see if I did it correctly. Thanks!

dosumis:
Looks like these are coming from the cl import, so will need to be fixed in CL as the cl-import.owl file will just get regenerated.  I'll close this and add details of fix on ticket.  Apologies for not looking more closely - the issue would have been obvious to me if I had.

Issue #1722, opened 2021-01-13
Intestinal junction


dosumis:
Rationale: 

Intestinal junctions are currently classified as  'anatomical junction', which is correct, but a bit too generic.  There should be a better grouping term for intestinal junctions that indicates that they are subdivisions of the intestinal tract, with all the layers etc this structure has.
**Preferred term label:**
intestinal junction

(bit clunky? investigate if something more standard is available.

**Synonyms**
junction of intestine 

(Check whether there's some fancy medical latin for this

**Definition (free text, please give PubMed ID)**

A subdivision of the intestine that is an anatomical junction between two subdivisions of the digestive tract. 

(There may not be any good general ref for this.  If not, just use ORCIDs)

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**

Assert multiple parent classes:

SubClassOf: 'anatomical junction' 
SubClassOf: 'intestinal subdivision'
SubClassOf: connected_to some 'intestinal subdivision'

**Your nano-attribution (ORCID)**

https://orcid.org/0000-0002-7073-9172

cmungall:
Seems reasonable. We may need to do more clarification on what intestinal
subdivision is. The original intent was to have this be classic textbook
subdivisions, but there is an argument for it being any cross section. But
in that case how many junctions are there? Infinite?

cmungall:
current subclasses of subdivision of 'digestive tract':

https://www.ebi.ac.uk/ols/ontologies/uberon/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FUBERON_0004921

![image](https://user-images.githubusercontent.com/50745/104527018-9e1d5d00-55b8-11eb-97f2-3b9b72c2fd27.png)

pattern for layers:

https://github.com/obophenotype/uberon/blob/master/patterns/mucosal_layers.yaml

we need a pattern for junctions

dosumis:
How about: 

subdivision of digestive tract
.    segment of digestive tract
.    .    major segment of digestive tract
.    .   .   Ileum
.    .    junction between major segments of digestive tract' (also asserted subclassOf anatomical junction)
.   layer of digestive tract

Not sure about patterns.  This might be safe enough:

EquivalentTo: 'anatomical junction' and connected_to some 'major segment of digestive tract'
subClassOf: 'junction between major segments of digestive tract'

RDruzinsky:
I don't know if this makes sense or makes any difference at all, but an
[...]

PR #1774, opened 2021-02-02
Add UBERON:8410024 intestinal junction

paolaroncaglia:
Addresses #1722


paolaroncaglia:
@dosumis 
@cmungall 
Following up on the discussion in #1722 , here's a basic solution. It hopefully addresses the needs of the Single Cell Expression Atlas, without delving too deep into the anatomical boundary/intestinal subdivision discussion. The reasoner infers 5 terms to be children of 'intestinal junction' and they look correct to me. The ones that are left out are cases where one of the two parts is not strictly intestine, e.g. rectosigmoid junction. If we wanted those (and other existing classes) to be included, we'd have to create a broader term including other parts of the digestive tract. A few more comments:
I couldn't find good dbxrefs or xrefs, so left David's ORCID only;
Similarly, I couldn't find other wordings/labels; we might ask Nancy George;
Looking at the asserted children of 'anatomical junction', I wonder if 'perineal body' is indeed a junction (def says it's located at a junction, which doesn't necessarily make it a junction?); should 'angle of oral opening' be better axiomatised? ; 'esophagogastric junction' is axiomatised differently from its siblings; and there may be a few more classes needing work?
Thanks.

dosumis:
Got a bit carried away patternising.  Here's the result

image

@cmungall - can you review?  I think good for now.  Worth rolling into Doc (as DOSDP?) so we can standarise modelling choices.  I'm using connects through out, but in cleaning up I've see 'continuous with', overlaps and 'connected to' all used.  Probably matters less which choice we make than that we standardise.

uberon_edit.obo
```diff
@@ -191394,6 +191396,17 @@ xref: FMA:16029
 intersection_of: UBERON:0000353 ! parenchyma
 intersection_of: part_of UBERON:0002106 ! spleen
 
+[Term]
+id: UBERON:1100000
+name: digestive tract junction
+def: "An anatomical junction between two parts of the digestive tract." [ORCID:0000-0002-7073-9172]
+synonym: "junction of intestine" EXACT []
```
mkeays, line 191403: not sure if it makes sense to have this as a synonym -- doesn't "digestive tract" include non-intestine parts as well (e.g. https://www.cancer.gov/publications/dictionaries/cancer-terms/def/digestive-tract )

paolaroncaglia: @mkeays 
@dosumis 
"junction of intestine" should be a narrow synonym, not an exact one please.

paolaroncaglia: (And you may want to add "intestinal junction" also as narrow synonym if not there already)

dosumis: I'll remove.  I think it's there by accident because I clone a term to make this one.

dosumis (approved):

Issue #1653, opened 2020-09-29
Subserosa of liver


rich-brennan:
**Preferred term label:**
Liver subserosa

**Synonyms**
Subserous layer of liver, Subserosal tissue of liver, Tela subserosa (Hepar), Hepatic subserosa

**Definition (free text, please give PubMed ID)**
Subserosal tissue of liver (FMA:15812)

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
liver (UBERON:0002107)
or liver serosa (existing UBERON NTR)
and subserosa (UBERON:0012375)

**Your nano-attribution (ORCID)**
0000-0002-0449-2730

rich-brennan:
UBERON lacks a specific term for the subserosal region of the liver. A generic UBERON term for subserosa (UBERON:0012375) exists. FMA (Foundational Model of Anatomy Ontology) includes a term for Subserosa of liver (FMA:15812) which should be adopted for UBERON

mkeays:
Added UBERON:8400023.

Issue #1652, opened 2020-09-29
Liver Serosa


rich-brennan:
**Preferred term label:**
Liver serosa

**Synonyms**
Serosa of liver (FMA:15811)
Tunica serosa hepatis

**Definition (free text, please give PubMed ID)**
Hepatic serous coat; peritoneal covering of the liver, enclosing almost all except for a triangular area on its posterior surface (the "bare area of the liver") and a smaller area where the liver and gallbladder are in direct contact. (https://medical-dictionary.thefreedictionary.com/serosa+of+liver)

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
liver (UBERON:0002107)

**Your nano-attribution (ORCID)**
0000-0002-0449-2730

rich-brennan:
FMA ontology has term Serosa of liver (FMA:15811), suggest this term is adopted as “liver serosa” to UBERON

rich-brennan:
UBERON lacks a specific term for the subserosal region of the liver. A generic UBERON term for subserosa (UBERON:0012375) exists. FMA (Foundational Model of Anatomy Ontology) includes a term for Subserosa of liver (FMA:15812) which should be adopted for UBERON

mkeays:
Other serosa in UBERON are children of [serous membrane](https://www.ebi.ac.uk/ols/ontologies/uberon/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FUBERON_0000042) -- will use this as parent and add `part_of` relationship to [liver](https://www.ebi.ac.uk/ols/ontologies/uberon/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FUBERON_0002107).

mkeays:
Added UBERON:8400021.

PR #1775, opened 2021-02-03
adding three new liver terms: liver serosa, liver subserosa, subcapsu…

### LINKED ISSUE #1651: Liver subcapsular region
**Preferred term label:**
Subcapsular region of liver

**Synonyms**
Liver subcapsular region

Region of liver parenchyma immediately below the capsule (PMID:30992875)

liver (UBERON:0002107)

0000-0002-0449-2730

mkeays:
…lar region of liver (fixes #1651, #1652, #1653)

Requested for the upcoming release.


paolaroncaglia (changes requested):
Hi @mkeays,
A few minor notes:

synonym: "tunica serosa hepatis" EXACT
synonym: "tela subserosa (hepar)" EXACT [FMA:15821]
synonym: "tela subserosa hepatis" EXACT [https://medical-dictionary.thefreedictionary.com/subserosa+of+liver]
Could you please add an annotation: synonym type property = latin term

relationship: dc-contributor http://orcid.org/0000-0002-0449-2730
relationship: dc-creator http://orcid.org/0000-0003-2034-601X
(in all 3 terms)
These are a bit weird: your ORCID as a creator should be set up in Protege, as per CL training guidelines; and the contributor's ORCID should go as a definition dbxref please.

Thanks,
Paola

mkeays:
Hi @paolaroncaglia -- thanks very much for checking. I've added the 'latin term' synonym type property and moved the contributor's orcID to the definition as a database cross reference. It turns out that although I had my orcID in my Protege settings I didn't have the box checked to add my orcID as the creator (my laptop got wiped recently and had to set everything up again, must have forgotten this part!) -- is there a better way to do this now that the terms are already created?

paolaroncaglia:
@dosumis 

> It turns out that although I had my orcID in my Protege settings I didn't have the box checked to add my orcID as the creator (my laptop got wiped recently and had to set everything up again, must have forgotten this part!) -- is there a better way to do this now that the terms are already created?

Any suggestion please? Possibly less error-prone than hand-editing the obo file? ;-)

mkeays:
@paolaroncaglia @dosumis it seems that previous terms I've added, before my laptop was wiped and I did have the correct setup, my orcID URL was added to the `creator` Annotation field by Protege -- shouldn't that be the same as choosing the "+" by "Annotations" and selecting "creator" from there? It looks the same in my previous terms in Protege, but I can also see "created_by" which it looks like some terms have in UBERON...

paolaroncaglia:
@mkeays 

> @paolaroncaglia @dosumis it seems that previous terms I've added, before my laptop was wiped and I did have the correct setup, my orcID URL was added to the `creator` Annotation field by Protege -- shouldn't that be the same as choosing the "+" by "Annotations" and selecting "creator" from there? It looks the same in my previous terms in Protege, but I can also see "created_by" which it looks like some terms have in UBERON...

I think that created_by is older, and that we should use creator going forward. I also think that both may be acceptable, but I strongly suspect that the "relationship" bit displayed in your diff wouldn't... @dosumis please?

dosumis:
> I think that created_by is older, and that we should use creator going forward. I also think that both may be acceptable, but I strongly suspect that the "relationship" bit displayed in your diff wouldn't... @dosumis please?

I'm still using OboInOWL:created_by. @matentzn @nicolevasilevsky - Have we moved on from this?
The relationship thing is an OWL2OBO conversion issue.  The tag for non-OBO annotation properties should be property_value.  This is likely a problem with creator not being typed as an OWL annotationProperty.  We should fix this in the master.

Re bulk changes - it is possible to change all uses of an OWL entity in an OWL file in one go using Protege.  I can show you how in a quick screenshare.  Otherwise find and replace on text file.  Just be careful of the diffs whatever you do.

mkeays:
It looks like all my entries have been added with `relationship: dc-creator`, having had Protege settings as in the attached image; I can change them to `created_by` in a text editor, but I'm not sure how to get the Protege settings correct so that it doesn't do this in future. These three I did manually because my settings had been wiped, but the seven other UBERON terms I have added before that also have this issue, done by Protege.
Screenshot 2021-02-04 at 09 43 57

paolaroncaglia:
@mkeays 
In my Protege preferences, in the new entities metadata tab, the creator property field is filled out as follows, based on the CL training material:
http://www.geneontology.org/formats/oboInOwl#created_by
Hope that helps.

mkeays:
Thanks @paolaroncaglia -- have changed mine to that and updated `uberon_edit.obo` to correct it.

nicolevasilevsky:
I use created_by

image

dosumis (approved):

dosumis (approved):

PR #1778, opened 2021-02-03
Update uberon_edit.obo

megbalk:
deprecating lines and points made that will be in the uberonfovt module. I will put in the new IDs to the obsolete values as soon as they are made.


dosumis (changes requested):
Would be better to do this once you can add replacement IDs pointing to fovt (as replaced_by axioms)

megbalk:
@dosumis I will do that once I get those new terms created in the uberonfovt module. I'll close this for now and request again later.

PR #1780, opened 2021-02-04
Vasculature template + compiled OWL

dosumis:
This currently takes all assertions of connections between Uberon terms for elements of the vasculature and adds connected_to relationships between them.  

Template generated from Vasculature ASCT-b table by

https://github.com/obophenotype/CCF_tools/blob/master/src/ccf_tools.py#L32
+
https://github.com/obophenotype/CCF_tools/blob/master/src/template_generation_tools.py#L52

Still to do:  
 - Decide whether connected_to is the right relationship to use.
 - Add verification step that checks for the presence of asserted or inferred connections before adding.  (Verification via UberonGraph triple store queries is already in place: https://github.com/obophenotype/CCF_tools/blob/master/src/uberongraph_tools.py)


cmungall (commented):
what's the status of this?

was there intended to be a yaml here?

dosumis:
superceded.

PR #1798, opened 2021-02-17
Uberon ODK migration

matentzn:
Some comments from last meeting: 

- [x] Make sure  metazoan-composite diff and human-view are ok
- [x] Erro1 broken rdf
- [x] Track http://purl.obolibrary.org/obo/FBbt_00004910 through the whole pipeline
- [x] First time users of the pipeline suffer from the non-existence of `tmp`, `mirror` and `reports` directories
- [x] Can't open ../tmp/cl-zfa-xrefs.obo: No such file or directory at ../make-bridge-ontologies-from-xrefs.pl line 25, <> line 84756. after running make stages on a new machine
- [x] ext.owl has Uberon terms floated at the top


matentzn:
Some comments from last meeting: 

- Make sure  metazoan-composite diff and human-view are ok
- nucleus under wbbt go
- Erro1 broken rdf
- Track http://purl.obolibrary.org/obo/FBbt_00004910

matentzn:
- [ ] The stage pipeline does not require imports to be refreshed first
- [ ] First time users of the pipeline suffer from the non-existence of `tmp` and `reports` directory

dosumis:
Done lots of spot checking or Euarchontoglires.  Looks good - classes marker as being outside the taxon are eliminated as expected.  Quite a lot of work to do on deepening and extending taxon restrictions - but these are not relevant to getting this release out.  Will make a general ticket.

dosumis:
still on draft so can't approve, but need to make release.  Key users can't wait longer

dosumis:
On further inspection - taxon slims still have problems 

From spot checking - it looks like only_in_taxon axioms are sufficient to remove axioms from outside of earchontoglires, but never_in_taxon APs to taxonomic groupings including euarchontoglires are not.

So pectoral fin should be stripped because it is a 'paired fin' - which has never_in_taxon tetrapoda. antler is stripped because has 'in taxon' some Cervidae - which is outside or Euarch

dosumis:
Looks great.   The stripping on never_in_taxon seems to be working as intended.  One issue (I think minor) - we seem have have ended up with a set of classes sitting outside of the of the Uberon classification:

image

Perhaps this is an indicator of some over-stripping.  I'd really like to be able to trace through some examples.  Is it possible to get access to intermediate files to investigate?  However,  the numbers are small enough and the need for the file great enough that I'm happy to release anyway. 

I was also surprised to see that these classes get some classification from running the reasoner.  Does that mean that this release is not pre-reasoned?  I'd like to check this before release & correct if necessary.

dosumis (changes requested):
Please check that end product is pre-reasoned.  See comments for reason to believe that it is not.  Once we're clear that it is, I'm happy to release.

dosumis:
It may be that over-stripping means that another round of inference instantiation is needed to get the optimum/complete reasoned graph.

dosumis (approved):

cmungall (approved):

PR #1805, opened 2021-03-16
Resave to prevent spurious diff.

balhoff:
This is just a resave—no ontology changes. I was making another edit and noticed all these extra changes in my diff.


matentzn:
Some of these changes seem odd to me (significant label change?):

image

Maybe have @dosumis review this instead?

balhoff:
The taxon name changes are consistent with the ncbitaxon_import update in this commit: https://github.com/obophenotype/uberon/commit/9af94555ce4179026260eeae199390aaebfeb9ca#diff-7c597415e022211f3ba53705a3d86edbc39caec88d42d3ff233d4931c87d1ea9

matentzn:
Looks fine to me :)

PR #1807, opened 2021-03-17
Resave to prevent spurious diff.

balhoff:
@matentzn I updated from master after you merged #1805, and now I get a few more diffs! Not sure why this is happening. I'm using Protégé 5.5.0.


balhoff:
Actually if I open this one fresh, and save again, it flips back to the previous. This must be an ordering bug in the OBO serializer.

balhoff:
These are all cases where there is a "duplicate" annotation axiom (same property and value) but with different axiom annotations.

matentzn:
I added the Mondo normalisation pipeline to master just now.. you can run `make NORM` to see what it does; unfortunately, it does not seem to be able to handle secondary OBO format qualities:

image

image

Not sure how important this is, or whether this is actually working as intended?

matentzn:
Will not fix, so closing..

PR #1808, opened 2021-03-17
Remove property chains inferring develops_from over part_of.

balhoff:
This PR removes two property chains:

- `part_of o develops_from -> develops_from`
- `develops_from o part_of -> develops_from`

We noticed these causing some unexpected inferences in GO, for example `protein-containing complex` being inferred to develop from an `anatomical entity`. These should be in RO anyway, if they were to be accepted.

I also deleted these chains from the `cl_import.owl`. I looked at the CL source, and I think they got these via Uberon, so hopefully that bit of ontology autophagy will be cleaned up when both have releases.

I computed a reasoned file using this command: `robot reason --catalog catalog-v001.xml --exclude-duplicate-axioms true --remove-redundant-subclass-axioms true -i uberon_edit.obo -o reasoned-new.ofn`

and compared before and after the changes using `robot diff`. Somewhat unexpectedly, I didn't see any classification differences.


cmungall (approved):

dosumis:
Just spotted this.  I think the property chains are kosher and should be in RO.

balhoff:
@dosumis I think that chain leads to weird statements:

'neuron projection bundle' develops_from 'multicellular organism'
'flask cell' develops_from 'organ'
'interrenal gland' develops_from 'abdomen'

Issue #1872, opened 2021-05-26
NTR: anorectum


paolaroncaglia:
As requested by @sfexova after offline discussion, for use in the Single Cell Expression Atlas anatomograms:

**Preferred term label:**
anorectum

**Synonyms**
anorectal canal [narrow]
anorectal [relational adjective]

**Definition (free text, please give PubMed ID)**
A subdivision of the digestive tract in humans that includes the rectum, the anal canal and the anus. 
https://link.springer.com/chapter/10.1007/978-3-319-26938-2_8 ( DOI https://doi.org/10.1007/978-3-319-26938-2_8 , ISBN 978-3-319-26936-8)
https://pubs.rsna.org/doi/full/10.1148/rg.2015150037 (https://doi.org/10.1148/rg.2015150037)
PMID:29742860

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
equivalent to 'subdivision of digestive tract' and 'has part' some rectum and 'has part' some 'anal canal' and 'has part' some anus
present in taxon Homo sapiens 

**Your nano-attribution (ORCID)**
Silvie's ORCID: https://orcid.org/0000-0003-3440-1876

Optional: editor note: Although a few resources seem to only include the distal part of the rectum in 'anorectum', this grouping term is meant to refer more broadly to the whole of rectum, anal canal and anus.

paolaroncaglia:
@cmungall 
Would you or others have any objection to the creation of a grouping term for 'anorectum' as described in https://github.com/obophenotype/uberon/issues/1872#issue-902530452? Sorry for pestering, I'd like to get the edits in for SCA, in time for the next Uberon release. Thanks!

sfexova:
Hi @paolaroncaglia - thank you! Just one question, sorry if it's a dumb one - I know you mentioned offline that rectum cannot be made part of anorectum because that wouldn’t be true for species that don’t have an anal canal but where exactly does the current definition and relationships place this term in the ontology? - Is this term going to be under 'large intestine' and have children 'rectum', 'anal canal' and 'anus'?

paolaroncaglia:
Hi @sfexova ,

[...]

PR #1875, opened 2021-05-27
Add UBERON:8410015 anorectum

paolaroncaglia:
Addresses #1872


paolaroncaglia:
Please see discussion in #1872. There are 2 commits in this PR, but the second commit is a test and should not be approved unless anus is part of large intestine.

paolaroncaglia:
> This is a very complex issue, because the anorectal canal is certainly part of the large intestine but the anus, proper, is really just an opening in the body wall. The tissues of the external part of the anus, including the external anal sphincter, develop from the body wall, distinct from the gi tract. 

Thanks @RDruzinsky for confirming!
@sfexova , if you need the anatomogram zoom-in of "anorectum" to link up to the large intestine image(s), please consider if instead of 
anorectum = rectum + anal canal + anus
you could use
anorectal canal = rectum + anal canal
I'm fairly optimistic that it would still represent your intents correctly, but please confirm. Thanks.

paolaroncaglia:
Hi @sfexova ,

> Hi @paolaroncaglia thank you for investigating and for your feedback @RDruzinsky - let me check first whether perhaps I made a mistake in the definition of anorectum in the first place - I'm going to double-check the references I have found to see whether they actually include anus proper or whether that was an extension that I made in my head

Thanks for looking into this. I will hold on to creating any new Uberon terms potentially related to the large intestine anatomograms until I hear from you. Best,
Paola

paolaroncaglia:
I'll close this PR as it's old and will re-do it based on recent discussion, to avoid potential mistakes.

PR #1913, opened 2021-06-07
Some annotation property fixes needed for merging phenoscape-ext into OBO file

balhoff:
Mainly - unobsoleted 'provenance notes' and declared as annotation property (there is another provenance_notes with an non-obo-friendly IRI in ext.

Also fixed an annotation property typo. Some of these changes are spurious line reorderings.


paolaroncaglia (commented):
Hi @balhoff ,
Looks good to me, but note I haven't come across obsoletion of properties before, so I'm not 100% sure of the SOP.
The spurious line reorderings occur in my PRs too.
Thanks,
Paola

balhoff:
Thanks, I will go ahead and merge. Once we combine phenoscape-ext into this repo, we can migrate the provenance notes to some other annotation property if we want to.

PR #1935, opened 2021-06-28
Comparing FBbt bridge files after ODK migration

gouttegd:
/!\ Not for merging /!\

This PR is solely intended to provide an easy way to compare the FBbt bridge files in the course of the migration to ODK.


matentzn:
@gouttegd ODK now is in master, what do you want to do with this here PR?

gouttegd:
This was solely intended to provide an online diff of the bridges between the `master` and `odk` branches, so with the ODK now in master (sorry, I had missed the fact that the ODK migration was over) this is no longer needed.

PR #1956, opened 2021-07-06
Main Uberon release files

matentzn:


dosumis (approved):

dosumis:
Files in release only:
image

Merging to Master now as release is urgent.  

However, we have not yet made a tag release.  @balhoff - would like your input on this.  Simple tag release now would => version IRI resolution using latest version IRI to old versions taxon slims, composite etc.  Do you think this is acceptable?  Maybe these should => 404?

matentzn:
Basically you answered it. As a power user of versioned ontologies, we wanted to ask whether making a tagged release for a partial release can cause havoc or not. Sounds like not :D

PR #1989, opened 2021-07-26
adding copepodite stages for issue 136 in ecocore

diatomsRcool:
This is my first PR to uberon, so I'm sure there will be changes requested.
Added copepod developmental stages, copepodid stage 1 thru 6 with definitions and axioms. All are children of life cycle stage.


dosumis (approved):

dosumis:
To pass checks you will need to add references to definitions:

image

matentzn:
Look at other examples of terms for this;

- If the definitions came from your brain,  use your orcid; 
- If the definition came from a paper, use the pmid
- If the definition came from Wikipedia, use that

PR #1993, opened 2021-07-28
Fixing owl-axiom issues

matentzn:
see #1991


src/ontology/uberon-edit.obo
```diff
@@ -225642,18 +225646,6 @@ is_a: UBERON:0001744 ! lymphoid tissue
 relationship: part_of UBERON:0010417 ! lymph node T cell domain
 created_by: http://orcid.org/0000-0002-2825-0621
 
-[Typedef]
-id: RO:0002473 ! composed primarily of
-alt_id: UBREL:0000002
```
dosumis, line 225647: Do we need to make sure these alt_Ids are preserved with declarations + replaced_by annotation axioms?   - see also BFO IDs below.
(CC @cmungall )

src/ontology/uberon-edit.obo
```diff
@@ -116472,7 +116469,6 @@ property_value: seeAlso https://github.com/obophenotype/uberon/issues/539 xsd:an
 [Term]
 id: UBERON:0006720
 name: pterygoid muscle
-alt_id: UBERON:3010730
```
dosumis, line 116475: WHy are we losing these axioms?

matentzn: Because these terms are obsolete! Which caused the uberon to be outside of OBO.

src/ontology/uberon-edit.obo
```diff
@@ -184,7 +184,6 @@ property_value: doap-GitRepository https://github.com/cmungall/uberon/ xsd:anyUR
...
 treat-xrefs-as-has-subclass: EHDAA
```
dosumis, line 187: Doesn't look like all of the axioms that ended up i the header are accounted for by new content in OBO.  Usually there's some simple mistake causing divergence from OBO.  Occasionally we have logical axioms that need to be in an extension OWL file but don't see that here.

src/ontology/uberon-edit.obo
dosumis: There's a much bigger issue here in that case!  We should be committed to retaining chains of IDs through obsoletion and merges.  There should have been a step at some point where we converted all alt_ids into class declarations + replaced_by mappings.  Without this, updating legacy annotation sets will become impossible.   This is something we need to keep a close eye on across ontologies formerly or currently in OBO format.

src/ontology/uberon-edit.obo
matentzn: Are you sure, I think they are all accounted for.. mainly the alt-id ones. Can you give me an example of an axiom unaccounted for?

src/ontology/uberon-edit.obo
matentzn: What is the purpose here of `alt_id`? I could see that the other way around, you may want to keep a replaced_by annotation.. Is this maybe what was meant here?

dosumis: > So in this case UBERON:3010730 is obsolete.

Ahh - I see it is in Uberon after all and replaced_by is there:

image

As long as this always happens, it's fine to drop the alt_id (although some legacy software may be affected)

src/ontology/uberon-edit.obo
dosumis: Was expecting to see obsolete stanzas in OBO in the diff, but I guess these are already in place, 

e.g. 

image

matentzn: I made a ticket here:
https://github.com/obophenotype/uberon/issues/2018

I don't think any of it is critical, but if you think so, let me know which ones are really important.

src/ontology/uberon-edit.obo
dosumis: BTW - there's nothing in OBO format that prevents axiomatisation AFAIK.  Might be some other pipeline or some weirdness with conversion.

Issue #1669, opened 2020-11-11
Typo in definition of several "epithelium" terms 


paola-scibite:
**Uberon term**
Several, e.g. 
gustatory epithelium UBERON:0002926
ectocervical epithelium UBERON:0012251
ear epithelium UBERON:0015807
and more

**Bug description**
Definition (possibly template ones) start with "A epithelium..." instead of "An epithelium...".
Thanks.

paolaroncaglia:
@dosumis @matentzn 
Assigning to you as possibly pattern-related, please.

matentzn:
Note to self:

Regex:
```
"A ([aeio])
```

Replace:

```
"An $1
```

I have fixed all these (and more, see regex above) now in #1995 but I am not aware of patterns generating anything automatically.

matentzn:
Thanks a ton @paolaroncaglia your tickets have really just helped me quickly made thousands of improvements to Uberon. You can assign everything like mass typos etc always to me as well!

matentzn:
@paolaroncaglia I think this can be closed - if you want to give it a quick check..

paolaroncaglia:
@matentzn 

> @paolaroncaglia I think this can be closed - if you want to give it a quick check..

Actually, please don't close, and don't shoot the messenger... :-D we should probably instruct OBOL:automatic to use "an" instead of "a" for the parts_of as well when they start with a vowel

![Screen Shot 2021-10-01 at 18 38 58](https://user-images.githubusercontent.com/4982516/135656898-26a6481d-89d9-4ca7-8b6c-d15d6822483a.png)
 
Though I guess we can live with this. And I promise I'll stop looking for bugs for today.
Thanks Nico!

matentzn:
I wonder what OBOL is :D

PR #1995, opened 2021-07-28
Uppercasing definitions and adding periods

### LINKED ISSUE #1677: Some definitions don't start with uppercase and don't end with a period
I presume you're aware of this issue, but I don't know if it's ticketed. Can't say how widespread it is and/or if it occurs more often in some branches than in others, but here are a few random examples:

cervical gland
UBERON:0012247
mucus-secreting glands in the mucosa of the uterine cervix

gland
UBERON:0002530
an organ that functions as a secretory or excretory organ

liver lobule
UBERON:0004647
the polygonal structure of the liver that consists of hepatocytes radiating outward from a hepatic vein

liver perisinusoidal space
UBERON:0006729
the extravascular space in the liver between the parenchymal cells and a sinusoid that contains the blood plasma

matentzn:
Fixes #1677 #1669

This PR:
- Adds periods to all definitions where missing
- Makes all definitions upper case where they are not
- Replaces a few false double quotes with single quotes
- Replacing A.. with An.. where appropriate


matentzn:
Note this revealed a duplicate definition, which I fixed like this:
https://github.com/obophenotype/uberon/commit/b6f6fde6b3e9d58a4bfc462e3ce0a5bde4473b31

PR #2007, opened 2021-07-29
Issue 1623

megbalk:


megbalk:
@matentzn please approve

balhoff (changes requested):
@megbalk does this term have a genus concept? What should its superclass be?

megbalk:
@balhoff should i change the equivalencies to subclasses to fix that issue?

balhoff:
If any possible thing that has these three relations (has_part dentine, has_part dental pulp, part_of calcareous tooth) could be called 'endodontium', then continue using equivalentClass. But if you can imagine something with those three relations that is not an endodontium, then make those subclass relations.

But in either case I think you need to pick a parent term (some kind of tissue?). For the first case, that term would be part of the equivalence intersection, for the subclass case, it would be another superclass. I hope that's not too confusing. :-).

RDruzinsky:
These are developmental derivatives of the dental papilla. Does that help?

Robert E. Druzinsky, Ph.D.
Clinical Associate Professor
Dept. of Oral Biology
College of Dentistry
University of Illinois at Chicago
801 S. Paulina
Chicago, IL 60612
***@***.***

Office: 312-996-0406
Lab: 312-996-0629
Website: www.peerj.com/RobertDruzinsky

cmungall (changes requested):
I'm surprised the automated checks did not complain about lack of genus

I agree with @balhoff this should probably not be expressed with an equivalence axiom, these should be subclasses. A major red flag is when the OWL definition and the text definition don't match. See https://douroucouli.wordpress.com/2019/07/08/ontotip-write-simple-concise-clear-operational-textual-definitions/

Another red flag is the text definition references a concept that is not in the ontology (endodont)

For selecting the is_a parent, you have a choice of "odontoid tissue" or "multi-tissue structure". We should go back and revisit some of these early CARO decisions, I don't think this is a good choice as it leads to a ragged hierarchy. I would place under odontoid tissue.

cmungall:
> If any possible thing that has these three relations (has_part dentine, has_part dental pulp, part_of calcareous tooth) could be called 'endodontium', then continue using equivalentClass. But if you can imagine something with those three relations that is not an endodontium, then make those subclass relations.

I can: a calcareous tooth.

(or, if we interpret part-of as being non-reflexive, a calcareous tooth minus some atoms)

megbalk:
I think I made the necessary changes. I apologize if I'm not understanding something.

balhoff (changes requested):
@megbalk relationships look good; you just need to add an is_a parent (a superclass term). @cmungall suggested a couple.

megbalk:
ok I think I did it!

balhoff:
@megbalk that looks good! Last question: the def says "pulp-dentine complex or endodont". Is "endodont" another word for "pulp-dentine complex"? Should there be a term in Uberon for endodont? Are there other kinds of endodonts that are not endodontiums? If we don't need a separate endondont term, should "endodont" be a BROAD synonym for this term?

megbalk:
@RDruzinsky, I may defer to you here: does there need to be a term "endodont"? I am thinking not, and making "endodont" a broad synonym.

balhoff:
Or maybe it's even an EXACT synonym.

RDruzinsky:
Both "endodont" and "pulp-dentin complex" are exact synonyms.

megbalk:
Great, added the synonyms. I'll assign someone to review.

dosumis (approved):

PR #2011, opened 2021-07-30
Fix issues found by fastobo-validator.

balhoff:
@matentzn the ontology passes fastobo-validator now. Can we re-enable the check in the build?


matentzn (approved):
This is awesome thanks. Can you merge this, then merge master I to pr #1994 and see wether all is green, then merge that? I am on vacation :P thanks

PR #2019, opened 2021-08-03
NEVER MERGE: illustrating OBO serialisation issues

matentzn:
1. OBO normalisation using ROBOT ([commit](https://github.com/obophenotype/uberon/pull/2019/commits/66d7bd2c3a7c6f95ccb4e5ab6efae169e1210f0e))
2. Protege save ([commit](https://github.com/obophenotype/uberon/pull/2019/commits/79064dd5eb9bb3e623c3a4dfd9100515bf1f6dda))
3. OBO normalisation using ROBOT ([commit](https://github.com/obophenotype/uberon/pull/2019/commits/bee86c10e026d7fe886500b772069196654b90c1))


dosumis:
@matentzn - is the normalization via `robot convert` ?

matentzn:
```
normalise_robot: .FORCE
	$(ROBOT) convert -i $(SRC) -f obo --check false -o $(SRC)
```

You can try it yourself:

```
sh run.sh make normalise_robot
```

dosumis:
Let's see what Chris says.  Suspect it will be easy to fix in Robot.

cmungall:
I get the same behavior on mondo, go.

So it looks like we are using owltools to do roundtrips for mondo. This means it is either (1) an owlapi version thing (2) some kind of flag setting difference.

cmungall:
OK, it looks like OWLTools bypasses the main owlapi writer interface as it needs to provide the obowriter with a NameProvider:

https://github.com/owlcollab/owltools/blob/05524b3c6b5effeddffa8d0c5f53b8e537ff7a33/OWLTools-Runner/src/main/java/owltools/cli/CommandRunner.java#L6155-L6181

Here is the code for nameprovider:

https://github.com/owlcollab/owltools/blob/05524b3c6b5effeddffa8d0c5f53b8e537ff7a33/OWLTools-Core/src/main/java/owltools/io/ParserWrapper.java#L297-L364

However, we shouldn't need to do all this now. Clearly the OWLAPI has the logic for providing the names for referenced fields, as comments are preserved when doing a save from Protege, as is evidenced by the fact that GO and Mondo use obo format for the edit file and have many classes in the import closure that are referenced. So rather than re-implement the owltools logic we should just figure out the owlapi magic.

cmungall:
__UPDATE__

ok, this is very odd

in mondo, the comments are in fact preserved:

```bash
robot --catalog catalog-v001.xml convert -i mondo-edit.obo -f obo -o z.obo
$ grep UBERON: z.obo|more
relationship: disease_has_location UBERON:0001235 ! adrenal cortex
intersection_of: disease_has_location UBERON:0001729 ! oropharynx
relationship: disease_has_location UBERON:0002048 ! lung
relationship: disease_has_location UBERON:0002048 ! lung
intersection_of: disease_has_location UBERON:0010167 ! beard
```

in go, they are **not**

```bash
$ robot --catalog catalog-v001.xml convert -i go-edit.obo -f obo --check false -o z.obo
$ grep UBERON: z.obo|more
intersection_of: results_in_development_of UBERON:0001434
intersection_of: results_in_formation_of UBERON:0001474
intersection_of: results_in_formation_of UBERON:0001981
intersection_of: results_in_development_of UBERON:0001305
intersection_of: results_in_growth_of UBERON:0000035
intersection_of: results_in_growth_of UBERON:0000036
intersection_of: results_in_growth_of UBERON:0000037
intersection_of: results_in_growth_of UBERON:0001306

cmungall:
**DIAGNOSED**

OK, it looks like `--check false` causes the comments to be dropped. I have no I idea why, but that is the cause.

The solution in some cases is to not provide this flag. Note this only works for ontologies that conform to OBO structural cardanlity constraints. But IMO all ontologies should anyway

cmungall:
apologies for the stream responses...

Here is the logic error in robot (I may well have written this myself...):

https://github.com/ontodev/robot/blob/5689f6f4db0efc3e85a5180f2e1270e370e8ba9b/robot-core/src/main/java/org/obolibrary/robot/IOHelper.java#L1715

dosumis:
So can we fix Robot?

cmungall:
making a ticket

cmungall:
https://github.com/ontodev/robot/issues/896

matentzn:
I have added a potential fix here: https://github.com/ontodev/robot/pull/903

anitacaron:
I am closing this PR because it's not relevant anymore.

Issue #2022, opened 2021-08-04
add calcaneus terms


megbalk:
**Preferred term label:**
1. posterior facet for talus of calcaneus
2. middle articular facet for talus of calcaneum
3. anterior facet for talus of calcaneum
4. calcaneal body
5. tubercle of calcaneus
6. lateral tubercle of calcaneus
7. medial tubercle of calcaneus

**Synonyms**
1. proximal facet of calcaneum
2. 
3. distal facet of calcaneus
4. corpus calcanei
5. trochlear process of calcaneus; calcaneal tubercle

**Definition (free text, please give PubMed ID)**

1. Posterior-most facet on the proximal surface of the calcaneus for the articulation of the talus bone.
2. Middle facet on the proximal surface of the calcaneus for the articulation of the talus bone.
3. Anterior-most facet on the proximal surface of the calcaneus for the articulation of the talus bone.
4. The calcaneus body is the part of the bone not used for articulation with other bones.
5. A raised projection on the calcaneus.
6. A raised projection on the lateral side of a calcaneus.
7. A raised projection on the medial side of a calcaneus.

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
1. part of' some   calcaneus; facet
2. part of' some calcaneus; facet
3. part of' some calcaneus; facet
4. part of' some calcaneus; 'zone of bone organ'
5. tubercle and 'part   of' some 'calcaneus'
6. tubercle of calcaneus
7. tubercle of calcaneus

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0003-2699-3066

megbalk:
I will create a branch and add them

matentzn:
Check my last comment:
![image](https://user-images.githubusercontent.com/7070631/128178500-f1528efb-b2f8-45f5-a230-3ea984a59b61.png)

Wrong file, perhaps outdated repo (`src/ontology/uberon-edit.obo`

megbalk:
made pull request from branch fovt-calcaneus2 #2028

paolaroncaglia:
Hi @megbalk ,
Should this ticket be closed, please? 
Thanks,
Paola

PR #2028, opened 2021-08-04
Fovt calcaneus2

megbalk:
#2022


balhoff (approved):

cmungall (approved):
This is great, thanks!

It would be good to come up with some DPs here, but spatial DPs are non-trivial, it's easy to think you are saying something in OWL and actually be saying something else

why doesn't have 7500092 have "articular" in the name?

would it not be good to have a grouping class for AFfToC?

these aren't blockers though

megbalk:
@cmungall are you suggesting the name be "anterior articular facet..."?

megbalk:
added articular to the terms.

cmungall:
Replaced by #2052

PR #2030, opened 2021-08-10
Add a workflow to automatically close stale issues.

gouttegd:
Add a new GitHub Actions workflow to be executed daily to:
- mark issues that have not been updated in 5 months as 'Stale';
- close 'Stale' issues that have not been updated in 1 month.

This workflow will only apply to issues that were created or updated *after* March 1st, 2021, to avoid auto-closing all the issues from the past 3 years.

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -15,4 +15,4 @@ jobs:
           days-before-issue-close: 30
           close-issue-label: 'autoclosed-unfixed'
           close-issue-message: 'This issue has been closed automatically because it has not been updated in 6 months. Please re-open if you still need this to be addressed.'
-          start-date: '2018-09-01T00:00:00Z'
+          start-date: '2021-03-01T00:00:00Z'
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -1,0 +1,18 @@
+name: 'Mark and close stale issues'
+
+on:
+  schedule:
+    - cron: '0 1 * * *'
+
+jobs:
+  stale:
+    runs-ons: ubuntu-latest
+    steps:
+      - uses: actions/stale@v4
+        with:
+          days-before-issue-stale: 150
+          days-before-pr-stale: -1
+          days-before-issue-close: 30
+          close-issue-label: 'autoclosed-unfixed'
+          close-issue-message: 'This issue has been closed automatically because it has not been updated in 6 months. Please re-open if you still need this to be addressed.'
+          start-date: '2018-09-01T00:00:00Z'


matentzn:
Awwwesome, thank @gouttegd 

However, waaaay too tight.

I would prefer issues older than 2 years being closed by this in general. Lets put this on the Uberon meeting agenda.

matentzn:
Action items Uberon editorial team:

- [ ] Decide suitable parameters for closing (date frames, how to treat the last 3 years issues)
- [ ] Review message and labels

gouttegd:
I put arbitrary thresholds just to have something we can discuss about. :) Happy to change those parameters once we reach a consensus about them.

gouttegd:
The way this would work is as follows:

Every day:
* any issue that has not seen any activity (somebody commenting, adding/removing a label, adding/removing an assignee) for the past _X₁_ days will be flagged with a _stale_ label, marking it as a candidate for automatic closure;
* any issue with a _stale_ label that has not seen activity for the past _X₂_ days will be flagged with the label _autoclosed-unfixed_, a comment "This issue has been closed automatically because it has not been updated in _X₁_ + _X₂_ days. Please re-open if you still need this to be addressed." will be appended, and the issue will be closed;
* conversely, if an issue that was flagged with a _stale_ label sees some activity, the label is removed and the issue will no longer be considered a candidate for closure.

Assuming we agree on the basic idea, what needs to be decided are the following parameters:
- _X₁_, the duration of the period of inactivity after which an issue is to be considered "stale"; current proposition is 5 months, which is most likely way too short;
- _X₂_, the duration of the "grace" period after which a stale issue is to be closed in the continued absence of any activity; current proposition is one month.

The phrasing of the comment added when closing an issue may also be discussed. It is also possible to add such a comment when an issue enters the "grace" period, should we decide it (e.g., "This issue has not seen  any activity in the past _X₁_ days; it will be closed automatically in _X₂_ days if no action is taken.").

matentzn:
This is great, loving it!

My personal sense is:
- 6 months no reaction = stale
- 1 year stale = close

With ontologies we generally need to have a slightly longer breath - usually, we operate in pushes, so 1 year delay is pretty normal. If it looks like 2-3 curators go real wild on the Uberon issue tracker, maybe 6 months and 6 months is also ok.

@cmungall @dosumis should probably have the most experience here.

paolaroncaglia:
Thanks @gouttegd, this will be very helpful! 
I couldn't attend yesterday's meeting. @matentzn 's suggestion (6 months no reaction = stale, 1 year stale = close) is reasonable, but I'd add some extra steps in the process (not necessarily in Damien's workflow script), because in my experience items that are even slightly stale in any to-do list will continue to stay stale unless a conscious effort is made to look at them. (Either that, or some angry yelling from submitters.) I think we basically need a strategy to prevent staleness. So perhaps some or all of the following, in addition to Damien's workflow: 

- [ ] Plan a Uberon issue-aton (?) as Chris suggested, even just a couple of hours, at least every quarter. In my experience with GO and EFO, this strategy works well with backlogs. Duration and frequency of hackatons can be adjusted. Due to time zone jungle, we could schedule sub-groups and post comments/questions on the tickets, for people on the opposite hemisphere to address in their own time zone.   
- [ ] At every monthly Uberon call, set aside 10' to scan tickets that are open and unassigned. We could start with the oldest tickets of course (so the stale ones in Damien's workflow). The aim would be to quickly assign someone appropriate, or close with a quick comment if applicable, rather than fully address the request.
- [ ] If/when we get more funding/developers, shorten the time after which a ticket is labelled as stale (5-6 months would feel quite long to any submitter if the ontology is to be considered active).   
- [ ] If/when a curator goes wild on the Uberon tracker, contact them to discuss needs and manage expectations ;-)

Note, 1 year could still go by before a ticket is addressed if we work in batches/pushes, but at least we'd make the submitter aware of that. 

Thanks,
Paola

matentzn:
@dosumis how do you want to continue on this?

shawntanzk:
got thumbs up, will merge

PR #2048, opened 2021-09-08
Fix RO_import.owl

shawntanzk:
fix double def for "overlaps"
1) deleted all defs
2) refreshed RO


matentzn (approved):

matentzn:
Looks good, diff is noise. FMA xrefs lost their labels, which we don't really care about (they will return if they are so inclined).

PR #2049, opened 2021-09-10
Refreshed PATO

shawntanzk:
I refreshed PATO, wanted to run a release, but ran into error, not sure if it is because things are still moving around? 
checked uberon-edit and pato_imports.owl and did not find any obsolete so looks good. 
Happy to retry release and all, and work out what the error is if we should run a release, lmk @matentzn 
related to https://github.com/obophenotype/uberon/issues/1992


matentzn (approved):
The diff is huge, and there is an enormous amount if non-pato terms being imported through PATO.. Nothing to worry about now. If QC passes its fine.

PR #2054, opened 2021-09-23
Update uberon-idranges.owl

dosumis:
Adding ID range for Ray S.  (Note -  this file really should have be moved as part of ODK conversion - @gouttegd, could you move it?)


dosumis:
@gouttegd - sorry - don't worry .  I'll do it.

PR #2063, opened 2021-09-27
New release 27.09.21

matentzn:
I know these are tough ones to review, so here a reminder:

- [ ] clone branch
- [ ] open uberon.owl in top-level and inspect for "reasonableness" (does it appear more or less ok)
- [ ] check the diffs of uberon.obo, uberon-base.owl at least in some detail if they look ok, most of the other ones will be fine too
- [ ] Skim through all the files in the diff, and check that there are no major major issues (like all axioms deleted, etc).


cmungall:
where is this ENVO class coming from?

image

matentzn:
I have a better idea for this kind of work.

PR #2066, opened 2021-09-27
Deleting unnecessary files that create issues with cloning in Windows Environment

JoshuaFortriede:
I found 2 more seemingly meaningless files that contain colons. These are causing a problem with cloning the repo on a Windows environment.


matentzn:
Unfortunately these files are not irrelevant. They are part of some deeply hidden part of the uberon pipeline. What problems do they cause you?

dosumis:
Colons in filenames are incompatible with windows.

dosumis:
> Unfortunately these files are not irrelevant. They are part of some deeply hidden part of the uberon pipeline.

Does that mean I ***really*** shouldn't have made this hot fix ?

https://github.com/obophenotype/uberon/commit/7e6940aaf4dab168698af46e9380b5e0613cee75

JoshuaFortriede:
@matentzn, The problem that I am having is that on a windows machine, a colon is not a valid character for a filename. As such, I am not able to clone the repo in my windows environment. I am able to clone it in my WSL, but not in an area that I can access it from windows programs (i.e. Protege). So the main request is to either remove the files if not necessary, or to rename the files so they do not have a colon in them.

matentzn:
I did this in a different way not, should be fine.

PR #2098, opened 2021-10-13
issue 1686 adding greater and lesser palatine foramen

meghalithic:


matentzn:
@megbalk after you get an approval you can merge it yourself :)

Issue #2050, opened 2021-09-15
Refining the Uberon Euarchontoglires subset


paolaroncaglia:
Stemming from https://github.com/HumanCellAtlas/ontology/issues/84#. I scanned the Uberon Euarchontoglires subset to search for non-primate, non-rodent, non-rabbit terms. Note, this was a quick search and not meant to be exhaustive. Here are a few possible strategies to identify and filter out unwanted classes in a time-effective way. These would improve the subset quality and should help make the subset more manageable (it currently has >20k classes):

- [x] Search for annotation property ‘never in taxon’ ‘Homo sapiens’ (e.g. ‘ampullary gland’, ‘bone of reproductive organ’). Some terms may need double-checking in case they are found in non-human Euarchontoglires.

- [x] I spotted many fish-specific terms. Searching for created_by teleost_anatomy_curators should retrieve them all (e.g. 'fin fold pectoral fin bud’). Their IDs all start with UBERON:2

- [x] Same for amphibians: search by created_by amphibian_anatomy_curators (e.g. 'pars amphibiorum’). Their IDs all start with UBERON:3

- [x] Insect terms may be retrieved by searching for database_cross_reference contains FBbt (e.g.: 'egg chorion’). This will also bring up a dozen terms whose label contains “insect” (e.g. 'insect ring gland’).

- [x] Other random finds:
exoskeleton 
shell
'open circulatory system'
feather
rhinarium?
honey

- [x] Investigate adding taxon constraints en masse (I recall there was a similar comment from Chris some time ago somewhere; found it: https://github.com/obophenotype/uberon/issues/1824#issuecomment-817959064)

Thanks,
Paola

dosumis:
Hi @paolaroncaglia  - are you clear on how to add taxon constraints in these cases, or would you like to discuss?

paolaroncaglia:
@dosumis 

> Hi @paolaroncaglia - are you clear on how to add taxon constraints in these cases, or would you like to discuss?

I might ask @matentzn for help please with adding constraints in bulk, but I'll take a closer look and will then contact him offline if he's available. I'll update the ticket as we go - aiming to have the slim ready to integrate in the HCAO release pipeline for the Oct 19th release.

paolaroncaglia:
@cmungall wrote in https://github.com/obophenotype/uberon/issues/1824#issuecomment-817959064:

"UBERON:6x --> these are all specific to arthropods, and we can add TCs to 6x roots en masse
[...]

PR #2105, opened 2021-10-18
Add taxon constraints 'in taxon' Arthropoda or not in Mammalia

rays22:
This commit intends to
add taxon constraint 'in taxon' some Arthropoda
using a ROBOT template.
**NOTE** that there are lots of *diffs* in the `uberon-edit.obo` file that are not explicitly related to the intended taxon constraints.
This commit addresses #2050 in part.


dosumis:
Assume diff issue is XSD string bug again?

No idea why build is failing now.  I've set to re-run.  Ask Nico if fails again.

matentzn:
Ok, that is the problem. 

Now run ROBOT explain in unsatisfiability mode:

http://robot.obolibrary.org/explain

perhaps with `random:10` parameter. This should give you a sense what wrong

PR #2108, opened 2021-10-19
Add taxon constraints

rays22:
This commit intends to
1.  add taxon constraint `in taxon` *some* **Arthropoda** to some terms
using a ROBOT template,
2. add taxon restriction annotation `never in taxon` **Mammalia** (NCBITaxon:40674) to some terms using another ROBOT template.

If applied, this commit will address #2050.


rays22:
### ROBOT report
 
    robot reason --input uberon-edit.obo --output tmp/unsatisfiable.owl
    robot explain -i tmp/unreasoned.owl -M unsatisfiability --unsatisfiable random:10 --explanation tmp/explain.md

[explain.md](https://github.com/obophenotype/uberon/files/7375043/explain.md)

dosumis:
That certainly looks wrong:

Display name http://purl.obolibrary.org/obo/UBERON_3000961 UBERON:3000961 'external integument structure'
never_in_taxon Mammalia

dosumis:
Happy to approve if we can get a non-huge diff. Still too large to display.

cmungall:
I can see the diff if I click "load diff"

it's not optimal as sometimes the name is truncated, and there are many redundancies.

Coming up with the perfect entailment diff is not trivial (I am not sure if ontobot will work out the box here since we need to transform the APs to TBox axioms)

If it is not hard what about a query for all TCs in this batch:

 - only-in where there is not an only in to the same taxon or descendant taxon to the same term or an isa-partof-develops-from ancestor
 - never-in where there is not an never in to the same taxon or ancestor taxon to the same term or an isa-partof-develops-from ancestor

(I have prolog queries for all this but weaning us off them...)

If either the query or the review is too hard, I am OK to sign off on merging this, my intuition is that there will be very few false positives, and any such will be relatively low utility and can be added back later.

However, provenance *must* be added so that we aren't perplexed by these 3 years later...

matentzn:
You have correctly addressed the @cmungall issue, so you don't need to wait for his approval again (just for QC to pass).

Note that I changed your provenance a bit to use the "Mondo way". Confusingly, the Mondo way is:

1) we only use rdfs:seeAlso on term level, i.e. directly as annotation assertions on Terms
2) on axiom or annotation level, like here, we use oio:source instead of rdfs:seeAlso, because it looks nicer in OBO format


we should not change that system now, if we do not like it, we can fix it in bulk when the time is right. @rays22 can you add a note to our CL/Uberon/RO board to discuss this at the next call 

Once QC is passed, this is ready to be merged.

PR #2125, opened 2021-10-22
Check for cycles in basic.owl using souffle.

### LINKED ISSUE #1829: QC: Need cycle detection on basic.owl
We need to check basic for cycles ideally in basic file. High priority. Maybe ontobio. The old command is deprecated:

```
$(REPORTDIR)/%-cycles: %.obo
	echo "STRONG WARNING: $@ skipped, because there is no more blip." && touch $@
#	owltools --no-debug $< --list-cycles -f > $@
	#blip-findall -i $< "subclass_cycle/2" -label > $@
```

balhoff:
Fixes #1829.

Creates `tmp/basic-cycles.tsv`. ~This check is simple, but doesn't explain the path causing the cycle.~

### DIFF
--- a/src/scripts/cycles.dl
+++ b/src/scripts/cycles.dl
@@ -4,10 +4,22 @@
 #define OWL_ON_PROPERTY "<http://www.w3.org/2002/07/owl#onProperty>"
 #define OWL_SOME_VALUES_FROM "<http://www.w3.org/2002/07/owl#someValuesFrom>"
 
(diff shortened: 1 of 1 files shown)


balhoff:
I updated the output so that the paths are provided.

balhoff:
@matentzn I have been using a similar script for go-basic for quite a while now. So if you still want it, it works.

Issue #2116, opened 2021-10-21
Overhaul how mappings with FBbt (and possibly other organism-specific ontologies) work


gouttegd:
The following is a proposal, based on a discussion with @dosumis, to change how the mappings between Uberon/CL and FBbt are maintained and represented.

Currently, those mappings are maintained on the Uberon/CL side and represented by cross-references (`<http://www.geneontology.org/formats/oboInOwl#>:hasDbXref`) on the Uberon/CL term pointing to the FBbt term (though for some historical reasons, some mappings between CL and FBbt are made the other way round, with the cross-reference being carried by the FBbt term and pointing toward the CL term, as discussed in obophenotype/cell-ontology#1199). A custom logic in the Uberon repository takes care of extracting those cross-references and generates the corresponding bridge files `src/ontology/bridge/uberon-bridge-to-fbbt` and `src/ontology/bridge/cl-bridge-to-fbbt`.

The proposal here is:

1) To shift the maintenance of those mappings entirely to the FBbt side. That is, FBbt editors decide which terms should be mapped to which (though input from Uberon/CL editors would of course be welcomed).

2) To distinguish between “exact mappings” and “debatable mappings” (for lack of better terms). Basically, “exact mappings” are between terms that are indisputably equivalent, where the FBbt terms refers to exactly the same structure or cell type than the Uberon or CL term, with the only difference being that the FBbt term is specific to _Drosophila_ (e.g., the mapping between FBbt’s `epithelial cell` [FBbt:00000124] and CL’s `epithelial cell` [CL:0000066] is “exact”). By contrast, “debatable mappings” are between terms whose equivalence is not as clear-cut and may be argued in favour or against (e.g., it is debatable whether FBbt’s `subperineurial glial sheath` [FBbt:00007091] should be mapped to Uberon’s `glial blood brain barrier` [UBERON:0000202], because while the subperineurial glial sheath has been proposed as a possible model for the mammalian blood-brain barrier, it could be a stretch to state that the two are _equivalent_). “Debatable mappings” would probably be best represented using SSSOM, which would allow for more “nuanced” mappings.

3) FBbt would provide the “exact mappings” as a supplementary TSV file containing simple pairs {FBbt_ID, Uberon_ID} or {FBbt_ID, CL_ID}, and the “debatable mappings“ as SSSOM files.

[...]

PR #2132, opened 2021-10-26
Remove FBbt xrefs and get them from the FBbt-provided SSSOM file

gouttegd:
This PR implements the proposal in #2116 according to the discussion in the October 25th meeting of the Technical Board.

Roughly, this PR:

1. Removes all cross-references to FBbt terms from the UBERON edit file (the same thing will also need to be done in CL, but of course this is another repository so another PR).
2. Uses the FBbt-provided SSSOM file to generate a small OWL file containing cross-reference statement for all the exact mappings described in that SSSOM file.
3. Merges the OWL file produced at step 2 into the `$(TMPDIR)/seed.obo` file.

From step 3, we are back to the pre-existing pipeline for generating the bridge files: the `$(TMPDIR)/seed.obo` file still contains the FBbt xrefs (they have been added back at step 3), so the `make-bridge-ontologies-from-xrefs.pl` script can extract them and produce the bridge files.

A few things to note and discuss either here or at the next meeting of the Technical Board:

a. For now, the FBbt SSSOM file is committed to the repository. This is because we (FlyBase) have not done a FBbt release with the SSSOM file yet, so that file is not available online for download. Once we have done that, we can remove the SSSOM file from the UBERON repo and download it on the spot when generating the bridges.

b. With that pipeline, the cross-references are _not_ part of the released ontology: they are only included in the `$(TMPDIR)/seed.obo` file, which is only used to make the bridge files. If the intention is that the cross-references _should_ be part of the released ontology, then the OWL file generated in step 2 above should be made into a standard _component_, to be imported as any other components (instead of a temporary file). @matentzn Not sure what your intentions were on that point

c. The process that generates the bridge files from the cross-references (through the `make-bridge-ontologies-from-xrefs.pl` script) is somewhat broken is in that all bridge axioms are generated into the `uberon-bridge-to-fbbt` file, even the axioms about CL terms.


src/ontology/fbbt-mappings.sssom.tsv
```diff
@@ -0,0 +1,464 @@
+#curie_map:
```
matentzn, line 1: move to src/ontology/mappings directory and update makefile accordingly

matentzn (approved):
One tiny change

PR #2134, opened 2021-10-28
Rename and re-classify 'panarthropod tagma`

### LINKED ISSUE #1990: insect specific class is parent of panarthropod class
**Uberon term**

insect anterior-posterior subdivision of organism
Go to external page http://purl.obolibrary.org/obo/UBERON_6057001

-> arthropod (or pan arthropod?)  anterior-posterior subdivision of arthropod (or pan arthropod)?

Alternatively, just obsolete term - how useful it is as a grouping?

**Bug description**

insect specific class is parent of panarthropod class

image

Also - if we keep this grouping, shouldn't it include segments?

image

gouttegd:
This PR:

* Obsoletes the term `insect anterior-posterior subdivision of organism` (UBERON:6057001), on the grounds of its redundancy with the taxon-neutral term `subdivision of organism along main body axis` and its incorrect use as a parent term for `panarthropod tagma` (UBERON:6000002), which has a broader taxon restriction (#1990).
* Consequently re-classify `panarthropod tagma` under `subdivision of organism along main body axis`.
* Narrows the taxon restriction of `panarthropod tagma` to `Arthropoda` only instead of `Panarthropoda` (and updates the label accordingly), on the grounds that (i) it is unclear whether the term _tagma_ is commonly used for broader panarthropods and (ii) this is one of only two terms pertaining to panarthropods that we have in UBERON (the other one is `panarthropod head`, which is also of dubious usefulness).
* Fixes the definition of the previous term to remove the mention of “_three_ main subdivisions” (insects have three tagmata but this is not true for all arthropods).

closes #1990


matentzn:
@gouttegd can you fix this PR, there are some pretty severe conflicts which I cant fix by myself!

I am doing a release tonight, would be nice to get this in.

PR #2135, opened 2021-10-28
Replace to be obsoleted CL:1000416

shawntanzk:
see https://github.com/obophenotype/cell-ontology/pull/1298 for related issues


shawntanzk:
done

PR #2141, opened 2021-11-01
Delete obsolete class UBERON:3010730

### LINKED ISSUE #2140: UBERON:3010730 has no label
@rays22 to figure out what this is

rays22:
This commit intends to remove the obsolete class
UBERON:3010730 that also lacks a label.
If applied, this commit will fix #2140.


shawntanzk:
Should this term be obsoleted properly as per https://ontology-development-kit.readthedocs.io/en/latest/ObsoleteTerm.html or just removed?
From reading the past tickets, I think you should be right to remove, but maybe double check before you do?

rays22:
> Should this term be obsoleted properly as per https://ontology-development-kit.readthedocs.io/en/latest/ObsoleteTerm.html or just removed? From reading the past tickets, I think you should be right to remove, but maybe double check before you do?

I had checked what the state of the ontology was one commit before I re-introduced this term in my pull request <https://github.com/obophenotype/uberon/pull/2108>, and my present commit is in agreement with that state. More specifically, the class **UBERON:3010730 was already absent** from the ontology at that point and only my subsequent commit (re-)introduced it.

I think the ontology version where I re-introduced this class has not made it to the public yet, so it is very unlikely that it has seen much usage by the outside world to warrant a repeated obsoletion process.

shawntanzk (approved):

Issue #2145, opened 2021-11-01
Output of the expand-dbxref-literals.pl script is discarded


gouttegd:
The rule to generate the `$(TMPDIR)/uberon-edit.owl` file is broken:

```make
$(OWLSRC): $(SRC) $(COMPONENTSDIR)/disjoint_union_over.ofn $(REPORTDIR)/$(SRC)-gocheck $(REPORTDIR)/$(SRC)-iconv $(SCRIPTSDIR)/expand-dbxref-literals.pl
	echo "STRONG WARNING: issues/contributor.owl needs to be manually updated."
	owltools --no-logging $(UCAT) $< $(COMPONENTSDIR)/disjoint_union_over.ofn issues/contributor.owl --merge-support-ontologies --expand-macros -o  $@.tmp &&  $(SCRIPTSDIR)/expand-dbxref-literals.pl $@.tmp
```

The output of the `owltools` command is sent to `$@.tmp` (so `$(TMPDIR)/uberon-edit.owl.tmp`). Then that file is given as input to the `expand-dbxref-literals.pl` script, but then the result of that script is sent to the standard output without being written anywhere.

This has two consequences:
* First, and obviously, whatever that script does is lost;
* Second, it artificially increases the duration of that step of the pipeline, because of the time spent in needlessly dumping the entire contents of the ontology to the standard output (it also has the effect of cluttering the output).

What the script does is replacing lines such as this one:

```
<oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">PMID:20973947</oboInOwl>
```

by this:

```
<oboInOwl:hasDbXref rdf:resource="http://www.ncbi.nlm.nih.gov/pubmed/20973947"/>
```

If this is something we want to keep, it is trivial to fix the rule above to restore the proper behaviour and make sure the output of the script is not lost. If this is something that is no longer needed, we can simply remove the call to that script.

Issue #2144, opened 2021-11-01
Reference to inexistent ftr.obo file


gouttegd:
The `make-bridge-ontologies-from-xrefs.pl` script tries to open a file named `ftr.obo`:

```perl
open(F,"ftr.obo");
while(<F>) {
    print $fh $_;
}
close(F);
```

This file does not exist and is not referenced anywhere else in the repository (there’s no rule to generate it).

Does anyone know what is, or rather _was_, the purpose of this file? Should that file exist?

The above code causes a bunch of Perl warnings when generating the bridge files, because it tries to read from the file handle even when the handle is not valid due to the file not being present.

gouttegd:
I suspect `ftr` might stand for _footer_, and that the point of the above snippet is to allow to include an arbitrary piece of OBO code at the end of all the generated bridge files, simply by putting the piece into a `ftr.obo` file.

We’re not using that feature right now but it could become handy in the future, so I think the proper thing to do here is to keep the snippet, but with proper handling of the condition where the `ftr.obo` is absent to avoid cluttering the output with warnings.

cmungall:
```yaml
[Typedef]
id: foaf-depicted_by
name: depicted_by
is_metadata_tag: true
xref: http://xmlns.com/foaf/0.1/depicted_by

[Typedef]
id: foaf-homepage
name: homepage
is_metadata_tag: true
xref: http://xmlns.com/foaf/0.1/homepage

[Typedef]
id: dc-title
name: title
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/title

[Typedef]
id: dc-contributor
name: contributor
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/contributor

[Typedef]
id: dc-creator
name: creator
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/creator

[Typedef]
id: dc-publisher
name: publisher
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/publisher

[Typedef]
id: dc-description
name: description
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/description

[Typedef]
id: dc-source
name: derived from resource
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/source

[Typedef]
id: dcterms-isReferencedBy
name: is referenced by
is_metadata_tag: true
xref: http://purl.org/dc/terms/isReferencedBy

[Typedef]
id: dcterms-references
name: references
[...]

Issue #2143, opened 2021-11-01
Generation of CL bridge files is broken


gouttegd:
For each foreign ontology _X_ with which we have mappings, the pipeline is supposed to generate two bridge files in `src/ontology/bridges`: one for mappings with Uberon (`src/ontology/bridges/uberon-bridge-to-X`) and one for mappings with CL (`src/ontology/bridges/cl-bridge-to-X`).

What is currently happening is that only the `uberon-bridge-to-X` file is generated, containing both the Uberon mappings _and_ the CL mappings. The `cl-bridge-to-X` is not generated.

Not sure I fully understand why that is the case, but here are some paths to investigate:

The bridge files are generated by the `make-bridge-ontologies-from-xrefs.pl` script, which takes as input an OBO file and produces bridge files according to instructions of the form `treat-xrefs-as...` that it finds in the source file (for example, a line`treat-xrefs-as-reverse-genus-differentia: FBbt part_of NCBITaxon:7227` instructs the script to generate a bridge file from FBbt cross-references). Such instructions are present in `uberon-edit.obo`, but _not_ in CL, which must be why the CL bridges are not generated.

As for why the Uberon bridges contain axioms for both Uberon and CL (instead of Uberon only), this is seemingly because the file given as input to the `make-bridge-ontologies-from-xrefs.pl` script (`$(TMPDIR)/seed.obo`) is ultimately the result of merging `uberon-edit.obo` with a version of CL (`$(TMPDIR)/cl-core.obo`). I am not sure I understand why this merging is necessary.

gouttegd:
Update: the reason why the CL bridges are not generated is in this rule:

```make
$(TMPDIR)/cl-with-xrefs.obo: $(TMPDIR)/cl-core.obo $(SCRIPTSDIR)/expand-idspaces.pl
	if [ $(BRI) = true ]; then egrep '^(idspace|treat-)' $(SRC) > $@ && cat $< >> $@.tmp && $(SCRIPTSDIR)/expand-idspaces.pl $@.tmp > $@; fi
```

That rule is intended to create a version of CL containing all the `treat-xrefs-as...` stanzas copied over from `uberon-edit.obo`. However, this is not what it does, because the `treat-xrefs-as...` lines are sent to `$@` instead of `$@.tmp`, and `$@` is then completely overwritten by the result of the `expand-idspaces.pl` script.

This is an easy one to fix and I’ll have a PR for that tomorrow (once I have checked my fix does not break anything else…).

[...]

PR #2147, opened 2021-11-02
Fix generation of bridge files

gouttegd:
This PR addresses several small problems recently highlighted in the pipeline that generates the bridge files:

* CL bridge files (`bridges/cl-bridge-to…`) are not generated (#2143).
* Lots of Perl warnings about a “closed filehandle” when generating the bridges, due to attempts at reading a non-existing `ftr.obo` file (#2144).
* The result of the `expand-dbxref-literals.pl` is sent to stdout without ever being used (#2145).


matentzn:
The world is lucky to have you. I will take care of this later this week, awesome.

src/ontology/uberon.Makefile
```diff
@@ -90,7 +90,7 @@ checks: $(REPORTDIR)/uberon-edit-xp-check $(REPORTDIR)/uberon-edit-obscheck.txt
 # TODO: Huge number of printouts that pollute the general logs
 $(OWLSRC): $(SRC) $(COMPONENTSDIR)/disjoint_union_over.ofn $(REPORTDIR)/$(SRC)-gocheck $(REPORTDIR)/$(SRC)-iconv $(SCRIPTSDIR)/expand-dbxref-literals.pl
 	echo "STRONG WARNING: issues/contributor.owl needs to be manually updated."
-	owltools --no-logging $(UCAT) $< $(COMPONENTSDIR)/disjoint_union_over.ofn issues/contributor.owl --merge-support-ontologies --expand-macros -o  $@.tmp &&  $(SCRIPTSDIR)/expand-dbxref-literals.pl $@.tmp
+	owltools --no-logging $(UCAT) $< $(COMPONENTSDIR)/disjoint_union_over.ofn issues/contributor.owl --merge-support-ontologies --expand-macros -o  $@ &&  $(SCRIPTSDIR)/expand-dbxref-literals.pl $@ > $@.tmp
 	$(ROBOT) query -i $@.tmp --update $(SPARQLDIR)/taxon_constraint_never_in_taxon.ru --update $(SPARQLDIR)/remove_axioms.ru -o $@
```
matentzn, line 94: Seems the update query (row 94, using $@.tmp) may not be correctly applied if row 93 results in $@?

src/ontology/uberon.Makefile
```diff
@@ -1189,7 +1189,7 @@ $(BRIDGEDIR)/bridges: $(TMPDIR)/seed.obo $(TMPDIR)/cl-with-xrefs.obo $(TMPDIR)/c
 	if [ $(BRI) = true ]; then cd $(BRIDGEDIR) && perl ../make-bridge-ontologies-from-xrefs.pl ../$(TMPDIR)/seed.obo && perl ../make-bridge-ontologies-from-xrefs.pl -b cl ../$(TMPDIR)/cl-with-xrefs.obo ../$(TMPDIR)/cl-zfa-xrefs.obo && touch bridges; fi
 
 $(TMPDIR)/cl-with-xrefs.obo: $(TMPDIR)/cl-core.obo $(SCRIPTSDIR)/expand-idspaces.pl
-	if [ $(BRI) = true ]; then egrep '^(idspace|treat-)' $(SRC) > $@ && cat $< >> $@.tmp && $(SCRIPTSDIR)/expand-idspaces.pl $@.tmp > $@; fi
+	if [ $(BRI) = true ]; then egrep '^(idspace|treat-)' $(SRC) > $@.tmp && cat $< >> $@.tmp && $(SCRIPTSDIR)/expand-idspaces.pl $@.tmp > $@; fi
```
matentzn, line 1192: OMG!!! Great find.

src/ontology/make-bridge-ontologies-from-xrefs.pl
```diff
@@ -116,12 +116,14 @@
     print $fh "id: $prefix#develops_from\n";
     print $fh "equivalent_to: RO:0002225\n\n";
 
-    open(F,"ftr.obo");
```
matentzn, line 119: @cmungall please check

matentzn (changes requested):

src/ontology/uberon.Makefile
gouttegd: Before the patch, the `owltools` command in line 93 would send its result to `$@.tmp`. The `expand-dbxref-literals.pl` script (on the same line, after the `&&`) would then read that `$@.tmp` file, and send its own result to the standard output (i.e. on the terminal), where it is not used at all (there’s no redirection to capture the output into a file and use it later). Then on line 94, the update query would read again the output of the previous `owltools` command (still in `$@.tmp`). In other words, `expand-dbxref-literals.pl` is invoked for nothing at all, its output is lost.

With the patch applied, the `owltools` command in line 93 will now send its result to `$@`. Those results will then be read by the `expand-dbxref-literals.pl` script (on the same line), which will send its own result to `$@.tmp`. Then in line 94 the update query will work from the result on the `expand-dbxrefs-literals.pl` taken from `$@.tmp`.

I would have preferred to pipe the output of the `owltools` command directly to the `expand-dbref-literals.pl` script (_i.e._, `owltools ... | $(SCRIPTSDIR)/expand-dbxref-literals.pl > $@.tmp`), but unfortunately it seems that `owltools` does not allow to send its output to stdout, only to a real file (the classical Unix-y convention of using `-` to represent stdout does not work).

matentzn (approved):
Sorry didnt read it right! Thanks for the clarification.

src/ontology/make-bridge-ontologies-from-xrefs.pl
cmungall:
```yaml
[Typedef]
id: foaf-depicted_by
name: depicted_by
is_metadata_tag: true
xref: http://xmlns.com/foaf/0.1/depicted_by

[Typedef]
id: foaf-homepage
name: homepage
is_metadata_tag: true
xref: http://xmlns.com/foaf/0.1/homepage

[Typedef]
id: dc-title
name: title
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/title

[Typedef]
id: dc-contributor
name: contributor
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/contributor

[Typedef]
id: dc-creator
name: creator
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/creator

[Typedef]
id: dc-publisher
name: publisher
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/publisher

[Typedef]
id: dc-description
name: description
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/description

[Typedef]
id: dc-source
name: derived from resource
is_metadata_tag: true
xref: http://purl.org/dc/elements/1.1/source

[Typedef]
id: dcterms-isReferencedBy
name: is referenced by
is_metadata_tag: true
xref: http://purl.org/dc/terms/isReferencedBy

[Typedef]
id: dcterms-references
name: references
is_metadata_tag: true
xref: http://purl.org/dc/terms/references

[Typedef]
id: doap-GitRepository
name: Git repository
is_metadata_tag: true
xref: http://usefulinc.com/ns/doap#GitRepository

[Typedef]
id: doap-SVNRepository
name: SVN repository
is_metadata_tag: true
xref: http://usefulinc.com/ns/doap#SVNRepository

[Typedef]
id: doap-wiki
name: wiki
is_metadata_tag: true
xref: http://usefulinc.com/ns/doap#wiki

[Typedef]
id: doap-bug-database
name: bug database
is_metadata_tag: true
xref: http://usefulinc.com/ns/doap#bug-database

[Typedef]
id: seeAlso
name: see also
is_metadata_tag: true
xref: http://www.w3.org/2000/01/rdf-schema#seeAlso

[Typedef]
id: example_of_usage
name: example of usage
is_metadata_tag: true
xref: IAO:0000122
```

src/ontology/uberon.Makefile
matentzn: Thanks @gouttegd all great.

src/ontology/make-bridge-ontologies-from-xrefs.pl
matentzn: @gouttegd can you add this file to the repo and double check that the resulting behaviour is still ok?

gouttegd: If I add a `ftr.obo` file in `src/ontology/bridge` with the above contents, as expected I obtain the same OBO bridge files as before, with the added contents at the end of the file.

The `Typedef`s also end up as `owl:AnnotationProperty` stanzas in the corresponding OWL bridge files (which again is expected, since they are generated from the OBO bridge files).

Is that the intended result? Shall I commit that file?

matentzn: Just looking at a single bridge, are any of the properties used in ftr.obo _actually_ used in the bridge files?

gouttegd: Some of them are used in the prologue, namely *seeAlso* and all the *dc-* properties (here with the Uberon/FBbt bridge):

```
ontology: uberon/bridge/uberon-bridge-to-fbbt
data-version: 2021-11-11
property_value: seeAlso "http://purl.obolibrary.org/obo/uberon/references/reference_0000026" xsd:anyURI
property_value: dc-title "Uberon bridge to fbbt" xsd:string
property_value: dc-description "Taxonomic equivalence axioms between the drosophila anatomy ontology and Uberon. Note that currently Uberon focuses on chordate anatomy so there are only a minimal set of equivalencies for high level classes. In future Uberon may bridge via the arthropod anatomy ontology" xsd:string
property_value: dc-creator "Uberon editors" xsd:string
property_value: dc-contributor "David Osumi-Sutherland" xsd:string
property_value: dcterms-references http://purl.obolibrary.org/obo/fbbt.owl
property_value: dcterms-references http://purl.obolibrary.org/obo/uberon.owl
```

matentzn: ok then, commit ftr.obo please and make sure its included! Thank you! :)

gouttegd: Done, and the conflict introduced by the #2163 is fixed as well. Ready to merge everything if you’re happy.

matentzn (approved):
Ready to merge.

Issue #2137, opened 2021-10-28
Decide on graph walking vs EL inference for using taxon constraints to make subsets


cmungall:
Current strategy to make a taxon subset:

 - Add axiom Thing subClassOf part-of some NCBITaxon:nnnn
 - Remove any inter-species existentials
      - homology represented as reciprocal existentials
      - inter-species edges (less relevant for uberon)
 - Ensure all TCs are EL-ified
     - never_in becomes disjointWith in-taxon some X
     - Ensure NCBITaxon GCIs are added (in-taxon some X disjoint with in-taxon some Y for all sibs)
 - Reason with Elk
 - Eliminate all unsatisfiables

Note this has issues if we have:

 * A subclass part-of some B
 * B never-in-taxon

See 
https://github.com/geneontology/go-annotation/issues/3942

@balhoff says he has a solution

Some additional issues with the approach

 - hard to mentally reason over - logic is separated across ad-hoc filters on OPs, various pre-processing steps, property chains stored centrally in GO
 - Can require 10s of gigs of memory
      - this gets worse the further away from human we go
 - Hard to 'customize'. E.g. GO may want to include occurs_in when making their subsets

Outline alternative strategies in this ticket

matentzn:
@dosumis if you want @shawntanzk, @anitacaron and me to act on this, we need some specific instructions as it will fall to @anitacaron to take the bulk of this work, and it will occupy her for a few weeks (given she only has a few hours a week to dedicate to this project I mean). 

From the meeting I gather we need to:

- [ ] Decide who should document, and where the documentation lives
- [ ] Decide on the curation strategy (how do we get the taxon constraints into the ontologies)
- [ ] Decide on the technical strategy to materialise the logical constraints (DOSDP vs SPARQL)
- [ ] Decide on the technical strategy on extracting taxon views on the basis of these constraints

@shawntanzk I think we can handle this after all, its just going to be a slow process. If you want, you can put it up on the board again for next week.

cmungall:
For the graph strategy, it is easy to explore this using the existing ubergraph instance, which includes relation-graph inferences

See this query:
https://api.triplydb.com/s/8hs8rvxj3

Which is hardcoded to return classes EXCLUDED from a human view

Scroll up for an explanation of the query

[...]

PR #2160, opened 2021-11-09
Steps to get taxon exclusions from relation graph

cmungall:


matentzn:
Is this ready to merge? what are the associated steps, i.e. whats next?

cmungall:
Let's keep this unmerged for now until we have reached more clarity on #2137. If we decide to go a pure TBox strategy then we can close this.

However, I suspect at the least the graph strategy will be a highly useful adjunct to a TBox strategy. In which case we should tidy this PR:

 - decide on appropriate file names for products
 - which directories they should live in
 - how it should be integrated into the main flow

I think for these steps it should be someone more familiar with current ODK idioms that does this work

PR #2163, opened 2021-11-10
Merge the mappings component early in the pipeline.

gouttegd:
Make sure the cross-references for mappings (that have been moved to the components/mappings.owl file) are merged back early in the pipeline (in the first file derived from the -edit file), so that they end up in all generated artefacts.

This is the least disruptive option, as it allows to mimic closely the situation that existed before the mappings were moved to a separate component.

Without that, the mappings component is only merged at the end of the pipeline (when producing `uberon.owl`), when all other imports are merged.

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -92,7 +92,7 @@ checks: $(REPORTDIR)/uberon-edit-xp-check $(REPORTDIR)/uberon-edit-obscheck.txt 
 # TODO: Huge number of printouts that pollute the general logs
 $(OWLSRC): $(SRC) $(COMPONENTSDIR)/disjoint_union_over.ofn $(REPORTDIR)/$(SRC)-gocheck $(REPORTDIR)/$(SRC)-iconv $(SCRIPTSDIR)/expand-dbxref-literals.pl
 	echo "STRONG WARNING: issues/contributor.owl needs to be manually updated."
-	$(OWLTOOLS) --no-logging $< $(COMPONENTSDIR)/disjoint_union_over.ofn issues/contributor.owl --merge-support-ontologies --expand-macros -o  $@.tmp &&  $(SCRIPTSDIR)/expand-dbxref-literals.pl $@.tmp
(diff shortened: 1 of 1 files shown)


gouttegd:
@matentzn Note that this PR will most likely conflict with #2147, since both PRs update the same Makefile goal (the one making `tmp/uberon-edit.owl`). If you are happy with both this PR and #2147 (which is ready to be merged, I just need confirmation on what to do with the `ftr.obo` file), then I’d suggest dropping this PR and add the above commit to #2147 instead (where the conflict can be fixed), and then merge the result.

matentzn (approved):

matentzn:
This will work as intended, but for completeness, let me tell you the proper solution. 

In ODK, we have a concept of `components`. Components are like imports, only that they are "conceptually part" of the ontology. For example, FBCV has DPO as a simple "component" - which is not quite right for FBcv. Other components include src/patterns/definitions.owl, which contains all the axioms generated by DOSDP.  This is automatically classified as a component.

Mappings.owl here should be such a component. So this is what needs to happen:

1. Add mappings.owl as a component to `uberon-odk.yaml`, see [FBCV example](https://github.com/FlyBase/flybase-controlled-vocabulary/blob/master/src/ontology/fbcv-odk.yaml#L28)
2. run `sh run.sh make update_repo` to add the component formally (the component must be located in `src/ontology/components`)
3. As you will see you component will then appear in $(OTHER_SRC), see [example](https://github.com/FlyBase/flybase-controlled-vocabulary/blob/master/src/ontology/Makefile#L44).
4. Instead of merging the ontology as you do here by URI, you merge it using $(OTHER_SRC).

This will have the advantage that next time we add a component, it will just work and we don't have to edit the Makefile manually.

gouttegd:
Got it.

So how do you want to proceed? Do you want to merge the PR as it is and to postpone making a proper component for after the release?

matentzn:
Ok, since we need this here for the release, lets merge it (I am assuming you tested `.run.sh make IMP=false PAT=false uberon-base.obo` to check that the FBbt xrefs are indeed in there), but also make a new issue for building the components pipeline. This could also be a good thing for Anita to do, but we can discuss this next Monday.

PR #2184, opened 2021-11-22
Added root nodes

### LINKED ISSUE #2180: Set preferred roots Uberon
See for example mondo

https://github.com/monarch-initiative/mondo/blob/master/src/ontology/mondo-edit.obo

This is an annotation to the ontology itself.

IAO:0000700

shawntanzk:
Added anatomical entity and life cycle as preferred roots in uberon-edit file
Fixes #2180


matentzn (approved):
Very nice, thanks shawn!

PR #2189, opened 2021-11-23
Replace CL:1000416 with CL:0002324

### LINKED ISSUE #2188: Replace temrs that use CL:1000416 with CL:0002324
Related to merging terms in https://github.com/obophenotype/cell-ontology/pull/1298

shawntanzk:
Fixes #2188
Note: the rest of the usage are in CL import file (which I think will get updated when CL gets released and we do a refresh here - though not 100% sure about cyclicity)
For now I will leave them be - happy to remove them if need be


src/ontology/uberon-edit.obo
```diff
@@ -157956,7 +157956,7 @@ xref: DHBA:10431
 xref: FMA:76989
 xref: PBA:128013078
 intersection_of: UBERON:0013605 ! layer of lateral geniculate body
-intersection_of: composed_primarily_of CL:0011003 ! magnocellular neurosecretory cell (duplicate class)
+intersection_of: composed_primarily_of CL:0011003 ! magnocellular neurosecretory cell
```
shawntanzk, line 157959: not 100% sure why this change suddenly happened, but doesn't seem like a bad one

PR #2191, opened 2021-11-25
Import CL

anitacaron:


matentzn (approved):

matentzn:
@shawntanzk please check out this branch

- [ ] Refresh CL import
- [ ] Double check that the inference error we had is dealt with

shawntanzk:
With new CL release + this implementation, the problem is overall fixed with this PR -> which means if this PR gets released, we would have finally slain the zombie! 🥳

PR #2202, opened 2021-11-30
Remove spatial disjointness AP on PNS/CNS.

### LINKED ISSUE #2174: Double specification of CNS-PNS disjointness
Uberon considers that the `central nervous system` and the `peripheral nervous system` are spatially disjoint, that is nothing can be simultaneously part of both the CNS and the PNS.

This “disjointness” is expressed at two different places using two different mechanisms:

* There is a general class axiom `'part of' some 'peripheral nervous system' DisjointWith 'part of' some 'central nervous system'`. This axiom is stored in the file `src/ontology/components/disjoint_union_over.ofn`, which is merged with the main ontology at the very beginning of the pipeline, when generating `src/ontology/tmp/uberon-edit.owl` from the edit file.
* There is also an annotation `mutually_spatially_disjoint_with` carried by the term `peripheral nervous system` in the main edit file: `'peripheral nervous system' mutually_spatially_disjoint_with 'central nervous system'`. When generating the `src/ontology/tmp/uberon-edit.owl` file from the edit file, that annotation is expanded (using `owltools --expand-macros`) into the equivalence axiom `owl:Nothing EquivalentTo ('part of' some 'peripheral nervous system') and ('part of' some 'central nervous system')`.

My understanding is that these two mechanisms have _in fine_ the same result, so what is the rationale for having both?

Can we remove one of them? If so, any reason to prefer the `DisjointWith` general class axiom over the “equivalence-to-nothing” axiom, or the other way round?

gouttegd:
`peripheral nervous system` carries a `spatially_mutually_disjoint_with` annotation to express the spatial disjointness between the PNS and the CNS.

That spatial disjointness is already expressed as a general class axiom (in `src/ontology/components/disjoint_union_over.ofn`) though, so the annotation is redundant.

closes #2174


gouttegd:
@dosumis This is what has been decided in the last Uberon call, however I note that the `mutually_spatially_disjoint_with` annotation carries some extra informations – specifically the fact that the disjointness is not valid for the ZFA ontology:

```
relationship: mutually_spatially_disjoint_with UBERON:0001017 {exceptions="ZFA", source="https://github.com/obophenotype/uberon/issues/378", status="pending"} ! central nervous system
```

Removing the annotation to only keep the general class axiom would mean losing this bit of information. Is that desirable?

It does not seem that the `exceptions` value is used anywhere, so I assume it is only here for the benefit of Uberon editors, so that they are reminded of the conflict between ZFA and Uberon on the CNS/PNS disjointness issue.

dosumis:
> It does not seem that the exceptions value is used anywhere, so I assume it is only here for the benefit of Uberon editors, so that they are reminded of the conflict between ZFA and Uberon on the CNS/PNS disjointness issue.

I think that's correct. The logic does not allow for exceptions.  You could always retain this annotation on the GCI if you want to preserve it.

dosumis (approved):
If you wish to retain the annotation on the axiom, you could do so as a comment on the GCI.  Your call.

PR #2204, opened 2021-12-07
UBERON base approach

### LINKED ISSUE #2094: Creating procedures for imports
Current solution to messy imports should be to refresh imports in related ontologies when a release is made. This requires knowledge of import links so we can come up with the order in which is best to refresh import.

@shawntanzk to create an ontology dependency list and figure out a procedure on what should get imported first etc.

### LINKED ISSUE #2738: Switch CHEBI import to slim
After #2204 is merged, we should switch the unnecessarily enormous CHEBI import to the OBO slim, like we do for all other ontologies.

anitacaron:
Fixes #2094 
Fixes #2738


matentzn:
This is very good work @anitacaron 

Unfortunately, we need to really dig into `uberon.Makefile` as well, and I wont have time to do that before I leave now. 

Please add this PR to the project board and make sure we revisit it again on Feburary next year.

Action items:

- [ ] Understand which special "imports" goals from `uberon.Makefile` can be ditched. Right now, all of them, AND the mirrors, are overwritten by ideosyncratic code. Dealing with this will lead to a significant cleanup of Uberon Makefile, but is not easy.
- [ ] Create a "published" module, as the merged module as is will likely make `uberon.owl` too large for GitHub.

Please remind me of this PR in our first meeting in February! Thank you!

anitacaron:
This morning I added the property chain and generated another diff https://github.com/obophenotype/uberon/issues/2664#issuecomment-1328986014

matentzn:
@anitacaron when you tried adding the role chain to the role chains component, also make sure that uberon-simple.owl actually depends on tmp/materialized! Maybe this `materialised` goal should mostly be moved to tmp/uberon-edit.owl..

anitacaron:
Hmm, maybe that's why when I moved the property chain from the edit to the component, it got the same diff as previously.

matentzn:
#gogoeditdiff

anitacaron:
We imported the component with the properties chains for `develops_from` in the edit file. This will have the same reasoning result, but they will not be included in the base file.

anitacaron:
Finally, it's ready for final review and to be merged.

src/ontology/catalog-v001.xml
```diff
@@ -1,201 +1,11 @@
...
-        <uri name="http://purl.obolibrary.org/obo/uberon/mod/uberon-combined-embryonic-mammal.owl" uri="bridge/uberon-combined-embryonic-mammal.owl"/>
-        <uri name="http://purl.obolibrary.org/obo/uberon/mod/uberon-combined-eukaryote.owl" uri="bridge/uberon-combined-eukaryote.owl"/>
-        <uri name="http://purl.obolibrary.org/obo/uberon/mod/uberon-combined-mammal.owl" uri="bridge/uberon-combined-mammal.owl"/>
-        <uri name="http://purl.obolibrary.org/obo/uberon/mod/uberon-combined-metazoa.owl" uri="bridge/uberon-combined-metazoa.owl"/>
-        <uri name="http://purl.obolibrary.org/obo/uberon/mod/uberon-combined-opisthokont.owl" uri="bridge/uberon-combined-opisthokont.owl"/>
-        <uri name="http://purl.obolibrary.org/obo/uberon/mod/uberon-combined-teleost.owl" uri="bridge/uberon-combined-teleost.owl"/>
-        <uri name="http://purl.obolibrary.org/obo/uberon/mod/uberon-combined-tetrapod.owl" uri="bridge/uberon-combined-tetrapod.owl"/>
-        <uri name="http://purl.obolibrary.org/obo/uberon/mod/uberon-combined-vertebrate.owl" uri="bridge/uberon-combined-vertebrate.owl"/>
```
matentzn, line 187: Are you sure the above are not needed for composite metazoan? What is the reason to delete them all?

src/ontology/imports/chebi_terms.txt
```diff
@@ -1,2 +1,6 @@
 CHEBI:24431
+CHEBI:3306
+CHEBI:24943
+CHEBI:18246
```
matentzn, line 4: Why? I thought we deleted chebi import?

src/ontology/imports/envo_terms.txt
```diff
@@ -1,2 +1,5 @@
-ENVO:00001998
-
+http://purl.obolibrary.org/obo/ENVO_00001998
+http://purl.obolibrary.org/obo/ENVO_00000446
+http://purl.obolibrary.org/obo/ENVO_00000447
+http://purl.obolibrary.org/obo/ENVO_00002030
+http://purl.obolibrary.org/obo/ENVO_01000048
```
matentzn, line 5: Why is this needed? Does it even work?

matentzn: (in particular since you deleted all the other term files)

src/ontology/imports/ro_import.owl
```diff
@@ -239,6 +239,12 @@ WHERE {
     
 
 
+    <!-- http://purl.obolibrary.org/obo/RO_0002579 -->
```
matentzn, line 242: Why is this file still neeeded?

matentzn (commented):
Some observations:

- [ ] some term files are kept, others deleted
- [ ] are you 100% sure we need the edits to catalog? I am afraid this will change the composite metazoan etc file to using URLs for imports rather than local files?

src/ontology/imports/ro_import.owl
anitacaron: It's needed in the pipeline, to create OWLSRC.

Line 110 on uberon.Makefile: 
```
	(...)
		merge -i imports/ro_import.owl -o $@
```

src/ontology/imports/envo_terms.txt
anitacaron: The terms files that I deleted are from the imports that are not in the config file.

matentzn: Ok cool.

src/ontology/imports/ro_import.owl
matentzn: Does this still make sense? Is it even kept up to date by the OWLSRC goal?

anitacaron:
I returned the imports in the catalogue needed for the composite-metazoan and composite-vertebrate.

anitacaron:
I ran an offline release to ensure all imports were in the catalogue.

matentzn (approved):
Lets go for it!

PR #2209, opened 2021-12-14
Add vasculature template

anitacaron:
Related to https://github.com/hubmapconsortium/ccf-validation-tools/issues/120


dosumis:
Converting to draft for now as we need to discuss whether adding connected_to relationships is the best approach for these.

dosumis (changes requested):
* I don't think vasculature.owl should be committed
* Maybe robot templates need to be enabled in ODK to build components/vasculature_class.owl?
* Value for >AI should be full IRI (or CURIE might also work) - this looks like just string

anitacaron:
I followed this doc to include the template. https://github.com/obophenotype/uberon/tree/master/src/templates 
Enabling `use_custom_import_module` will create `external_import.tsv` with a list of terms to import when the imports are refreshed. (found here the doc https://github.com/OBOFoundry/COB/blob/53368763cd61e871543bfb8836aee8917cab8bc2/docs/odk-workflows/UpdateImports.md#using-the-custom-import-template) Is this what you hope to get?

dosumis:
I think this is happening because the import statement is looking for a file in components that you've committed in templates
image

I guess the make command doesn't move it into place? @matentzn  - is there a gap in the docs?

> Enabling use_custom_import_module will create external_import.tsv with a list of terms to import when the imports are refreshed. (found here the doc https://github.com/OBOFoundry/COB/blob/53368763cd61e871543bfb8836aee8917cab8bc2/docs/odk-workflows/UpdateImports.md#using-the-custom-import-template) Is this what you hope to get?

This is not relevant (we're not doing anything with imports).   I was just speculating that it might be possible to configure ODK to rebuild from robot templates & save the file to components automatically.

matentzn:
Didnt read all your comments, just fixed the catalog (see last commit). Was that all you need from me?

matentzn:
Remember @anitacaron catalog.xml is relative path from `src/ontology`, not `src/`.

anitacaron:
Thanks, @matentzn! 

Do we need to add `vasculature_class.owl`, or it will be generated when running `prepare_release`?

matentzn:
You need it, if you want it to be seen when opening uberon-edit.obo

src/ontology/uberon-edit.obo
```diff
@@ -53,6 +54,7 @@ treat-xrefs-as-equivalent: OG
 treat-xrefs-as-equivalent: VSAO
 treat-xrefs-as-is_a: VHOG
 import: http://purl.obolibrary.org/obo/uberon/components/mappings.owl
+import: http://purl.obolibrary.org/obo/uberon/components/vasculature_class.owl
```
matentzn, line 57: @anitacaron if you add this, the owl file has to exist as well! So yes, leave it checked in here as is!

src/templates/vasculature_class.owl
```diff
@@ -81,7 +81,7 @@
                 <owl:someValuesFrom rdf:resource="http://purl.obolibrary.org/obo/UBERON_0001142"/>
             </owl:Restriction>
         </owl:annotatedTarget>
-        <oboInOwl:inSubset rdf:resource="http://purl.obolibrary.org/obo/uberon/core#human_reference_atlas"/>
+        <oboInOwl:inSubset rdf:resource="&lt;http://purl.obolibrary.org/obo/uberon/core#human_reference_atlas&gt;"/>
```
anitacaron, line 84: @dosumis it was already good before changing to the complete URI; somewhere it's already defined (?)

anitacaron: I'll revert the commit.

dosumis (approved):

PR #2217, opened 2021-12-17
Added part_of cerebral cortex to laminar subdivision of the cortex

### LINKED ISSUE #2216: laminar subdivision of the cortex is currently not part_of neocortex
**Uberon term**
UBERON:8440004 - laminar subdivision of the cortex

**Suggested revision of logical definition**
Add: 'part of' some 'cerebral cortex'

shawntanzk:
Fixes #2216


paolaroncaglia (commented):
Hi @shawntanzk ,
'laminar subdivision of the cortex' is defined as "One or a subset of adjacent layers of the _neocortex_.", and 'cerebral cortex' is defined as "The thin layer of gray matter on the surface of the cerebral hemisphere that develops from the telencephalon. It consists of the _neocortex (6 layered cortex_ or isocortex), the hippocampal formation and the olfactory cortex.", so shouldn't 'laminar subdivision of the cortex' be part of neocortex more specifically? If there are parts of the cortex other than the neocortex that have layers, or if the neocortex is not identifiable in some species but there are still layers, could this be made more explicit?
Thanks!

shawntanzk:
@paolaroncaglia thanks for the catch, you are definitely right, it really should be neocortex and not cerebral cortex - will make the change

paolaroncaglia (approved):

Issue #1886, opened 2021-05-28
antorbital notch


megbalk:
**Preferred term label:**
antorbital notch

**Synonyms**

**Definition (free text, please give PubMed ID)**
The antorbital notch is a prominence of bone located anteriolaterally of the antorbital fenestra.

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
'bone fossa'
'part of' some skull

**Your nano-attribution (ORCID)**

megbalk:
I will create a branch and add this term. I suspect this term also needs taxon constraints.

megbalk:
@paolaroncaglia does this need a taxon constraint too?

paolaroncaglia:
Hi @megbalk ,

> @paolaroncaglia does this need a taxon constraint too?

Good point; when creating new terms, it's always useful to consider that. You don't provide references or usage examples, and I'm not sure what species have an 'antorbital notch'. However, it may be that the _parent_ terms you suggest already have the appropriate taxon constraints. In this case, the new term would inherit the taxon rule "only in Vertebrata" via 'bone fossa' and its ancestor(s), plus it would inherit the broader 'only in Craniata' via 'skull' (you can see all this in Protege). If 'antorbital notch' occurs in some vertebrates but not others, it'd be useful if you could add more stringent constraints. 
Also, in general, please consider adding definition database cross references  (papers, books, reliable internet resources...).
Thanks!
Paola

paolaroncaglia:
@megbalk 
Just a note to say that details of how to formalize taxon constraints are currently under discussion (there's a whole GitBub project board [here](https://github.com/orgs/obophenotype/projects/2#card-58876023)). In particular, 'only in taxon' is being considered for obsoletion, possibly in favor of 'in taxon', but I wasn't involved in the most recent discussion. So perhaps best to make a note in your tickets of what taxonomic groups the new term a) would never apply to, and/or b) is (only?) found in, and come back to apply those rules when a resolution is reached? I'll make a note in the agenda for the next CL/Uberon editors' meeting.

megbalk:
Thank you!

megbalk:
See "[better doc](https://obofoundry.github.io/obo-patterns.github.io/key-topics/taxon_constraints/)"

paolaroncaglia:
Hi @megbalk ,
Hope all's well. Assigning to you please. As far as I can see, the term hasn't been created yet. Thanks.
Paola

PR #2221, opened 2021-12-20
adding antorbital notch

megbalk:
from issue #1886


megbalk:
What is exit code 2?

paolaroncaglia:
Could anyone assist @megbalk with her question please: 

> What is exit code 2?

(Meghan, you may also wish to try doing the pull request again if the error came out of the blue)

megbalk:
I tried doing re-doing the pull request and got the same error. @cmungall or @nicolevasilevsky could you help? I don't know what "exit code 2" error is.

nicolevasilevsky:
where do you see that error code?

megbalk:
Clicking on details under "all checks have failed".
Screen Shot 2022-01-06 at 11 22 20 AM

balhoff:
You need to look at the next section in the checks output—here is the real error:

> OBO STRUCTURE ERROR Ontology does not conform to OBO structure rules:
17
multiple name tags not allowed. in frame:Frame(UBERON:7500070 id( UBERON:7500070)name( antorbital notch)property_value( http://purl.org/dc/elements/1.1/date 2021-05-28T18:26:38Z xsd:dateTime)namespace( uberon)name( sulcus tali)def( The antorbital notch is a prominence of bone located anteriolaterally of the antorbital fenestra.)def( A depression that runs between the two facets of a talus that interact with a calcaneum.[https://en.wikipedia.org/wiki/Talus_bone ])property_value( term:editor http://orcid.org/0000-0003-2699-3066)property_value( present_in_taxon Equus xsd:string)property_value( present_in_taxon Cetacea xsd:string)synonym( sulcus for flexor hallucis longus tendon EXACT)property_value( http://purl.org/dc/elements/1.1/date 2021-05-28T01:09:52Z xsd:dateTime)relationship( part_of UBERON:0002395)is_a( UBERON:0006846)relationship( part_of UBERON:0003129)is_a( UBERON:0000093))
18
For details see: http://robot.obolibrary.org/errors#obo-structure-error

megbalk:
Thanks @balhoff, I think I've fixed the issue

megbalk:
@paolaroncaglia can you review this?

paolaroncaglia (commented):
Hi @megbalk ,
In the definition, is "anteriolaterally" a misspelling of "anterolaterally"? If it's correct but just less used, it's ok to leave. 
Ideally, you'd want a reference in support of a new term, e.g. a PMID or ISBN or Wikipedia link as an annotation property (dbxref) of the definition. But if none is easily available, your ORCID will serve the purpose.
I see your Protege preferences are set to use term:editor, mine use a different property; I remember there was some discussion to adopt a standard, but I'm just back from sick leave and can't remember exactly. Perhaps @nicolevasilevsky may comment please if term:editor should not be used - otherwise I think it's acceptable. Thanks.

megbalk:
@paolaroncaglia thank you for catching those!
@nicolevasilevsky could you help if I should use editor or creator?

nicolevasilevsky:
I'm using this creator property:
http://www.geneontology.org/formats/oboInOwl#created_by

image

I think that is the standard we agreed on.

megbalk:
fixed! @paolaroncaglia let me know if it looks good.

paolaroncaglia:
@megbalk PR looks good to me, but the checks are failing again :-(

megbalk:
it says "Makefile:563: recipe for target 'sparql_test' failed". @balhoff @dosumis could you help me understand this?

balhoff:
The problem is `FAIL Rule ../sparql/illegal-annotation-property-violation.sparql: 1 violation(s)`. But it doesn't print which property. Maybe it's your `date` property?

megbalk:
@paolaroncaglia it looks like it's passed! could you please approve?

megbalk:
@paolaroncaglia it looks like it passed! Could you please approve?

src/ontology/uberon-edit.obo
```diff
@@ -224519,6 +224519,16 @@ is_a: UBERON:7500094 ! tubercle of calcaneus
...
 property_value: term:editor http://orcid.org/0000-0003-2699-3066
 
+[Term]
+id: UBERON:7500101
+name: antorbital notch
+def: "The antorbital notch is a prominence of bone located anterolaterally of the antorbital fenestra." [doi:10.5479/si.00810282.627]
+is_a: UBERON:0006846 ! surface groove
+relationship: part_of UBERON:0003129 ! skull
+property_value: present_in_taxon "Cetacea" xsd:string
+property_value: present_in_taxon "Equus" xsd:string
```
dosumis, line 224529: These should link to the terms in NCBI_Taxon, rather than being strings.

dosumis (changes requested):
See note on present_in_taxon axioms

src/ontology/uberon-edit.obo
megbalk: @dosumis I don't see an option for that in the annotation properties.

megbalk: @dosumis how do I do that?

dosumis: Just realised that Protege doesn't make it easy as you can't use the class expression editor for an AP value.
- you can  navigate to the term by hand like this - which is not very practical
image
- much easier is to search for the term 
image
copy the IRI (use COMMAND U to display) 
image
then paste into the IRI editor on the term you're editing:
image
=>
image

dosumis (approved):

Issue #2198, opened 2021-11-29
Taxon reporting pipeline


matentzn:
We want to inform curators about the consequences of adding taxon constraints. 

- [ ] generate a report that looks at the difference between a number of key subsets (human-view, euroglires-basic, mouse-view, potentially more, i.e. configurable). The report should just be class ID and label. In other words: a list should be produced with ID, label pairs indicating which classes are being excluded as a consequence of the current edit. The diff should be between the _version of uberon-edit.obo on master_ and the version of _uberon-edit.obo on the branch_ that performs the taxon constraint.
- [ ] When a user creates a pull request that changes taxon constraints (or for now, edits uberon-edit.obo) generate the reports and concat them into one file. If the file is not empty, paste a comment on the pull request with the report (i.e. the affected classes).

shawntanzk:
@rays22 to make smaller test file for @ubyndr to test on a branch

rays22:
@ubyndr ,
I have created two test files using [robot](https://oboacademy.github.io/obook/tutorial/robot_tutorial_1/) that differ only in taxon constraints. They could be used to generate the list of class IDs and labels that are excluded for the human-specific view.
1. [`test-master.obo`](https://raw.githubusercontent.com/obophenotype/uberon/issue2198t/src/ontology/test-master.obo) : This file is the mock 'master' version.
2. [`test-branch.obo`](https://raw.githubusercontent.com/obophenotype/uberon/issue2198t/src/ontology/test-branch.obo) : This is the edited ontology file that has one additional taxon restriction `property_value: never_in_taxon "Euarchontoglires" xsd:string` for term UBERON:0000022 (label: feather).

You can use them to test full versus human-views.

PR #2232, opened 2022-01-11
DO NOT MERGE with master

rays22:
This commit intends to add two test files related to #2198.

I have created two test files using [robot](https://oboacademy.github.io/obook/tutorial/robot_tutorial_1/) that differ only in taxon restrictions. They could be used to generate the list of class IDs and labels that are excluded for the human-specific view.
1. `test-master.obo` : This file is the mock 'master' version.
2. `test-branch.obo` : This is the edited ontology file that has one additional taxon restriction `property_value: never_in_taxon "Euarchontoglires" xsd:string` for term UBERON:0000022 (label: feather).

You can use them to test full versus human-views.


rays22:
The test files are ready so I am closing this PR.

PR #2239, opened 2022-01-20
Change vertebrate_skeletal_anatomy_curators from created_by to oboInOwl:source

### LINKED ISSUE #2213: changing created_by: [curators' group] to oboInOwl:source http://purl.obolibrary.org/obo/[ontology]
Comes from suggestion in PR https://github.com/obophenotype/uberon/pull/2206

anitacaron:
Fixes \#2213


shawntanzk:
Look back into historic version to see if there are xrefs to aao 
If we cant find them - we tag them with purl of the ontology that they came from

anitacaron:
Found out when started having anphibian_anatomy_curators terms on the [2013-07-13 release](http://svn.code.sf.net/p/obo/svn/uberon/releases/2013-07-03/uberon.obo)
I didn't see any xref AAO/VSAO/TAO in the terms

matentzn:
Ok.. so let's go as we said. Forget the xrefs for now - just at the provenance the way you were doing and we are fine. If there never were xrefs it does not matter now.

matentzn (approved):
Very nice! Ready to merge!

PR #2241, opened 2022-01-21
Add sparql check for taxon range

anitacaron:
This is a test for this https://github.com/INCATools/ontology-development-kit/issues/520

At the moment, there're 18 violations:

```
FAIL Rule ../sparql/taxon-range-violation.sparql: 18 violation(s)
term,taxon
http://purl.obolibrary.org/obo/UBERON_0006611,Euarchontoglires
http://purl.obolibrary.org/obo/UBERON_8410062,Homo sapiens
http://purl.obolibrary.org/obo/UBERON_8410010,Mammalia
http://purl.obolibrary.org/obo/UBERON_8410003,Mammalia
http://purl.obolibrary.org/obo/UBERON_8410025,Homo sapiens
http://purl.obolibrary.org/obo/UBERON_8410026,Homo sapiens
http://purl.obolibrary.org/obo/UBERON_8410027,Homo sapiens
http://purl.obolibrary.org/obo/UBERON_8410050,Homo sapiens
http://purl.obolibrary.org/obo/UBERON_0001154,Mammalia
http://purl.obolibrary.org/obo/UBERON_0000022,Euarchontoglires
http://purl.obolibrary.org/obo/UBERON_8400003,Mammalia
http://purl.obolibrary.org/obo/UBERON_8400002,Mammalia
http://purl.obolibrary.org/obo/UBERON_8400001,Mammalia
http://purl.obolibrary.org/obo/UBERON_8400007,Mammalia
http://purl.obolibrary.org/obo/UBERON_8400008,Mammalia
http://purl.obolibrary.org/obo/UBERON_8400005,Mammalia
http://purl.obolibrary.org/obo/UBERON_8400006,Mammalia
http://purl.obolibrary.org/obo/UBERON_0010053,Euarchontoglires
```


dosumis:
Looks good.  I think we fixed the errors in this PR: https://github.com/obophenotype/uberon/pull/2231

Do we have a way to tailor the error text in these cases, or some doc to help editors interpret the errors?  If not I'd like to add some text, if not I'm happy to approve and make this a more general issue for discussion.

matentzn:
This is awesome thank you! 

@dosumis this is odd - I would have expected a proper summary of the SPARQL verify command. 

@anitacaron can you check that if you run `sh run.sh make sparql_test` locally you don't see the error summary (listing all the sparql checks performed? If so, can your remove the conditionals https://github.com/obophenotype/uberon/blob/master/src/ontology/Makefile#L562 and see whether it has something to do with that `ifeq` thing?

anitacaron:
@matentzn it's not generating the reports either way. I think the problem is with the command `verify`.

anitacaron:
Ah no, sorry. It's normally generating the report, but the `src/ontology/reports/*` folder is in the `.gitignore`.

dosumis:
My question was more about whether/how it is possible to specify some more specific, editor friendly error message.  Something more than just the name of the test / sparql file or details of the sparql query?

matentzn:
@dosumis @anitacaron my bad. Because of the enormous size of the Uberon Log, I have piped it to a log file:

https://github.com/obophenotype/uberon/blob/master/.github/workflows/qc.yml#L33

Whose last rows are shown in the QC report:

image

We need to prime curators to look in there, sorry about that. That is standard output and should be sufficient to understand the problem. Sorry @anitacaron to send you on a goose chase, all good!

matentzn (approved):
Excellent!

matentzn:
The way we do it at Mondo is this:

When the curator sees this error, they can go here:
https://mondo.readthedocs.io/en/latest/editors-guide/quality-control-tests/

They can then find the check that has failed. If you include actually human readable descriptions of the SPARQL query inline as a comment on the query, that should do that trick.

A more hacky way is to grep the last lines of log for the failing SPARQLS and cat grep the Human readable description from out there. All that said - I think its better to train people to look at the SPARQL queries when in doubt and include good human readable explanations in the metadata.

dosumis:
One other thing:  I think this SPARQL check should be built into ODK and configurable per repo.  It's needed by CL and GO as well as Uberon.

balhoff:
We have a check in GO for IRI values: https://github.com/geneontology/go-ontology/blob/master/src/sparql/non-IRI-value-violation.sparql

We do use non-NCBITaxon_ IRIs: the taxon union classes.

dosumis:
> We do use non-NCBITaxon_ IRIs: the taxon union classes.

Good point.  We could extend to check both patterns, but maybe it's better to use a generic IRI check here then (in the absence of a way of specifying range semantically).  I guess the same check could be used with in_subset too and will work for any other AP shortcut.

Maybe we should just add the generic check to ODK?

src/sparql/taxon-range-violation.sparql
```diff
@@ -0,0 +1,12 @@
...
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX never_in_taxon: <http://purl.obolibrary.org/obo/RO_0002161>
+PREFIX present_in_taxon: <http://purl.obolibrary.org/obo/RO_0002175>
+
+SELECT ?term ?taxon
+WHERE {
+  ?term never_in_taxon:|present_in_taxon: ?taxon .
+
+  FILTER (!isIRI(?taxon) && (!STRSTARTS(STR(?taxon), "http://purl.obolibrary.org/obo/NCBITaxon_")))
```
shawntanzk, line 11: should be:
!isIRI(?taxon) || (isIRI(?taxon) && ((!STRSTARTS(STR(?taxon), "http://purl.obolibrary.org/obo/NCBITaxon_")))

shawntanzk:
> Do we fix the rest in this PR and update that PR when merging this PR?

ok think its probably less messy if we merged #2260 first then update this to check

src/sparql/taxon-range-violation.sparql
balhoff: @shawntanzk with Ubergraph you can specify particular named graphs to limit which inferences are returned. Also, much faster to use `?term rdfs:isDefinedBy <http://purl.obolibrary.org/obo/bfo.owl>` instead of FILTER. But these tips are only if you are querying only Ubergraph.

balhoff: See https://github.com/INCATools/ubergraph#graph-organization

anitacaron: @shawntanzk thanks, but no need to verify again if `isIRI` on the second side :)
`!isIRI(?taxon) || !STRSTARTS(STR(?taxon), "http://purl.obolibrary.org/obo/NCBITaxon_")`

shawntanzk: @anitacaron yeah I guess makes sense, if it starts with "http://purl.obolibrary.org/obo/" it is a IRI lol

matentzn: Are you sure? What if it's a string? I think you need isIRI. What makes sure that it is an iri rather than a string looking like an IRI?

anitacaron: It will fail in the first part `!isIRI(?taxon) || !STRSTARTS(STR(?taxon), "http://purl.obolibrary.org/obo/NCBITaxon_")`

matentzn (approved):

Issue #2233, opened 2022-01-12
Error on uberon.makefile ln100


shawntanzk:
```
$(OWLSRC): tmp/uberon-merged.owl $(COMPONENTSDIR)/disjoint_union_over.ofn $(REPORTDIR)/$(SRC)-gocheck $(REPORTDIR)/$(SRC)-iconv $(SCRIPTSDIR)/expand-dbxref-literals.pl
	echo "STRONG WARNING: issues/contributor.owl needs to be manually updated."
	$(OWLTOOLS) --no-logging $< $(COMPONENTSDIR)/disjoint_union_over.ofn issues/contributor.owl --merge-support-ontologies --expand-macros -o  $@ &&  $(SCRIPTSDIR)/expand-dbxref-literals.pl $@ > $@.tmp
	$(ROBOT) query -i $@.tmp --update $(SPARQLDIR)/taxon_constraint_never_in_taxon.ru --update $(SPARQLDIR)/remove_axioms.ru -o $@ \
           query --update ../sparql/delete-definition-dot.ru merge -i imports/ro_import.owl -o $@
```

The above is the command in the makefile where it is failing.

tmp/uberon-merged.owl seems to be built alright, but seems to be failing at the line that starts with $(OWLTOOLS)

`make: *** [uberon.Makefile:100: tmp/uberon-edit.owl] Error 1`

This is the same failure with trying to build imports and mirrors and trying to make tmp/uberon-edit.owl 

making tmp/uberon-merged.owl is fine though

shawntanzk:
full documentation of this: https://github.com/obophenotype/uberon/pull/2231#issuecomment-1011094937

dosumis:
This is sufficient to prompt the error:

./run.sh owltools tmp/uberon-merged.owl  --expand-macros
(assuming  tmp/uberon-merged.owl is built).

hypothesis - this is caused by an issue in never_in_taxon AP axioms that have changed since the last release

@shawntanzk can you do a diff of  uberon-edit.owl since the last release and grep the results for never_in_taxon to see if any look odd?

shawntanzk:
@dosumis not sure how to grep with old version and all, but the above is the diff between the last uberon release and current, the following are the ones that have never in taxon:

echolocation organ http://purl.obolibrary.org/obo/UBERON_0010053
Added
echolocation organ never_in_taxon "Euarchontoglires"^^string
exoskeleton http://purl.obolibrary.org/obo/UBERON_0006611
Added
[...]

PR #2248, opened 2022-01-24
Fix string on euarch

shawntanzk:
Related to #2233
Think it fixes the error - need to try before I close


dosumis:
Also fixed in Meghan's pattern PR.

dosumis (approved):

shawntanzk:
> Also fixed in Meghan's pattern PR.

sorry, was impatient, was trying to do another fix and came to that error again, thanks for the review :) will merge once the checks are completed!

Issue #2249, opened 2022-01-25
PR import broken


shawntanzk:
Can't seem to get PR import to work in UBERON.
Did manage to RO import to work though so it seems specific to PR
could use some help, thanks :) 

```
2022-01-25 08:56:43,476 ERROR (CommandRunner:4310) Could not find specified ontology http://purl.obolibrary.org/obo/pr.owl for --source
Exception in thread "main" java.lang.NullPointerException
	at uk.ac.manchester.cs.owlapi.modularity.SyntacticLocalityModuleExtractor.asAxiomSet(SyntacticLocalityModuleExtractor.java:287)
	at uk.ac.manchester.cs.owlapi.modularity.SyntacticLocalityModuleExtractor.<init>(SyntacticLocalityModuleExtractor.java:282)
	at owltools.cli.CommandRunner.runSingleIteration(CommandRunner.java:4357)
	at owltools.cli.CommandRunnerBase.run(CommandRunnerBase.java:76)
	at owltools.cli.CommandRunnerBase.run(CommandRunnerBase.java:68)
	at owltools.cli.CommandLineInterface.main(CommandLineInterface.java:12)
make: *** [uberon.Makefile:416: imports/pr_import.owl] Error 1
```

matentzn:
@shawntanzk do you want to change PR import in Uberon to the PRO slim? This will make the process much slimmer as well.

Steps:

- [ ] change uberon-odk.yaml to refer to the slim (see PCL, CL)
- [ ] open src/ontology/imports/pr_terms_combined.txt (or src/ontology/tmp/seed.txt) and copy all PR references in there and add them to the PRO slim like you did last time (make sure no duplicates, use `cat seed.txt | sort | uniq > new_seed.txt` to get a unique list).
- [ ] Make sure that in `uberon.Makefile`, the default `import/pr_import.owl` is used, i.e. comment out the one made by uberon.Makefile. 
- [ ] Have me review the PR

shawntanzk:
will do :)

shawntanzk:
@matentzn 
Currently commenting out PRO stuff in uberon.makefile (assume that the same has to be done with mirror/pr.owl stuff too) and notice that mirror/aminoacid.owl will be affected - do you know what I should do with it? Thanks

PR #2250, opened 2022-01-25
Added RP3V

### LINKED ISSUE #2238: [NTR] rostral periventricular region of the third ventricle
**Preferred term label:**
rostral periventricular region of the third ventricle

**Synonyms**
RP3V

a functional anatomical construct containing the key estrogen-regulated inputs to GnRH neurons necessary for estrogen positive feedback. [PMID:17604108]

also in that paper: 

A population of neurons scattered through the anteroventral periventricular (AVPV), periventricular preoptic (PVpo) and median preoptic (MEPO) nuclei project directly to GnRH neurons and express ERα. Together, they are referred to as
the rostral periventricular area of the third ventricle (RP3V) population of neurons.

(not sure if that means it belongs in UBERON or CL?) 

Brain? honestly need help with this.

**Notes**
In rats and mice, the anteroventral periventricular nucleus plus the periventricular nucleus make up the rostral periventricular region of the third ventricle (RP3V). WikipediaVersioned:Anteroventral_periventricular_nucleus&oldid=966137187

shawntanzk:
Fixes #2238

@dosumis & @paolaroncaglia have some questions that I would like input on: 

1) Questions on adding axioms:
'synapsed to' some 'gonadotropin releasing neuron'
expresses some 'estrogen receptor'
The structure is defined by neurons that do this, but the neurons are the one that do it rather than the structure

2) is overlaps correctly used here?

Note: currently can't get PR import to work properly, see #2249

Related to: https://github.com/obophenotype/cell-ontology/pull/1424


paolaroncaglia:
@shawntanzk My understanding from https://obophenotype.github.io/cell-ontology/relations_guide/ is that "synapsed to" should be used for cells, not anatomy (though tricky as your Oberon term represents a group of those cells, but still). And based on the def phrasing, I suspect that "overlaps" should be replaced by "has part", provided that the link is true for all species that have RP3V.
@dosumis ?

paolaroncaglia (commented):
See comment

shawntanzk:
thanks @paolaroncaglia that was super helpful, I will leave out the axioms (I think it's a similar concept for expresses). 

I was struggling between overlaps and the more specific 'has part', but used overlap here as I saw that 'secondary visual cortex' uses overlaps some 'Brodmann (1909) area 18' - though I do get that that isn't exactly what this case is. But yes, would like some input on this from @dosumis

src/ontology/uberon-edit.obo
```diff
@@ -225644,6 +225644,18 @@ def: "One or a subset of adjacent layers of the neocortex." [ISBN-10:0-87893-742
 is_a: UBERON:0011215 ! central nervous system cell part cluster
 relationship: part_of UBERON:0001950 ! neocortex
 
+[Term]
+id: UBERON:8440005
+name: rostral periventricular region of the third ventricle
+def: "A functional anatomical construct containing the key estrogen-regulated inputs to gonadotropin-releasing hormone expressing (GnRH) neurons necessary for estrogen positive feedback. The rostral periventricular region of the third ventricle contains a population of neurons scattered through the anteroventral periventricular (AVPV), periventricular preoptic (PVpo) and median preoptic (MEPO) nuclei project directly to GnRH neurons and express ERα." [PMID:17604108]
+comment: In rats and mice, the anteroventral periventricular nucleus plus the periventricular nucleus make up the rostral periventricular region of the third ventricle (RP3V). {xref="WikipediaVersioned:Anteroventral_periventricular_nucleus&oldid=966137187"}
```
shawntanzk, line 225651: make clear on interpretation

shawntanzk:
@paolaroncaglia, I've just talked to @dosumis about this, and given cells seem to be scattered in the region, overlaps seems safer to use. I've also added reciprocal overlaps on the 3 structures. Thanks :)

paolaroncaglia (commented):
@shawntanzk thanks for clarifying. Looks good to me then. Just one thing: are any words missing between "nuclei" and "project" below please? 
def: "...The rostral periventricular region of the third ventricle contains a population of neurons scattered through the anteroventral periventricular (AVPV), periventricular preoptic (PVpo) and median preoptic (MEPO) nuclei project directly to GnRH neurons and express ERα."

shawntanzk:
@paolaroncaglia it is supposed to refer to the nuclei of the "anteroventral periventricular (AVPV), periventricular preoptic (PVpo) and median preoptic (MEPO)", so technically no word in between, but I can see how that can be confusing - will figure a way to make it more clear later :) thanks

shawntanzk:
@paolaroncaglia ok updated the def - thinks it reads better now :) thanks

paolaroncaglia (commented):
@shawntanzk yes, much clearer now! Thanks, go for it ;-)

paolaroncaglia (approved):

Issue #1904, opened 2021-06-03
NTR: levatores costarum


paola-scibite:
**Preferred term label:**
levatores costarum

**Synonyms**
levatores costarum muscle
levatores costarum muscles
musculi levatores costarum [latin term]

**Definition (free text, please give PubMed ID)**
Adapt from Wikipedia: "The Levatores costarum (/ˌlɛvəˈtɔːriːz kəˈstɛərəm/), twelve in number on either side, are small tendinous and fleshy bundles, which arise from the ends of the transverse processes of the seventh cervical and upper eleven thoracic vertebrae. They pass obliquely downward and laterally, like the fibers of the Intercostales externi, and each is inserted into the outer surface of the rib immediately below the vertebra from which it takes origin, between the tubercle and the angle (Levatores costarum breves). Each of the four lower muscles divides into two fasciculi, one of which is inserted as above described; the other passes down to the second rib below its origin (Levatores costarum longi). They have a role in forceful inspiration."
WikipediaVersioned:Levatores_costarum_muscles&oldid=876138804

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
UBERON:0002426 chest muscle

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0002-2825-0621

**dbxrefs**
FMA:71312

paolaroncaglia:
@dosumis please confirm: I'll fix the "present in taxon" as per https://github.com/INCATools/ontology-development-kit/blob/b5ac0030c7523dccd57eab567fd85d6a0e807027/docs/TaxonRestriction.md. Correct? :-) Ta.

PR #2256, opened 2022-01-27
Added UBERON:8410070 'levatores costarum' for #1904

paolaroncaglia:


paolaroncaglia:
@dosumis could you please confirm that I've entered 
present_in_taxon "Homo sapiens"
in the correct way? The usage/syntax in Uberon seems inconsistent, I see both taxon labels and IDs being used. Thank you.

src/ontology/uberon-edit.obo
```diff
@@ -225615,6 +225615,20 @@ relationship: in_taxon NCBITaxon:8782 ! Aves
...
+synonym: "levatores costarum muscle" EXACT []
+synonym: "musculi levatores costarum" EXACT LATIN []
+xref: FMA:71312
+is_a: UBERON:0002426 ! chest muscle
+relationship: has_muscle_insertion UBERON:0002228 ! rib
+relationship: has_muscle_origin UBERON:0001077 ! transverse process of vertebra
+property_value: http://purl.org/dc/elements/1.1/date 2022-01-27T10:41:44Z xsd:dateTime
+property_value: present_in_taxon "Homo sapiens" xsd:string
```
dosumis, line 225629: Should use NCBI_taxon IRI of Homo Sapiens as value.

paolaroncaglia:
I'll merge for now; if I've entered present_in_taxon "Homo sapiens" incorrectly, I can fix it later, but it seems that it wouldn't be the only one in need of editing.

dosumis (changes requested):

src/ontology/uberon-edit.obo
paolaroncaglia: Just saw this... will fix, but what about all the others?

dosumis:
Oops - see you've merged.  We should now have a test for this.  I think all others have been fixed elsewhere now.

PR #2257, opened 2022-01-27
Fix annotation for #1904

paolaroncaglia:


dosumis:
Beat me to the merge again!  

You need to enter full IRI like this:
image

I think we have doc.  Will dig it out.

dosumis:
@shawntanzk - sorry to trouble you. I'm sure I was looking at some doc on this when we worked on it the other day.  Any idea where it might be?

Issue #2293, opened 2022-02-09
There are 6 unsatisfiable classes in the ontology


anitacaron:
When running `prepare_release` got this error when generating `materialized.owl` that uses `unreasoned.owl` as input.

```
ERROR There are 6 unsatisfiable classes in the ontology.
ERROR     unsatisfiable: http://purl.obolibrary.org/obo/UBERON_0039421
ERROR     unsatisfiable: http://purl.obolibrary.org/obo/UBERON_0009046
ERROR     unsatisfiable: http://purl.obolibrary.org/obo/UBERON_0009047
ERROR     unsatisfiable: http://purl.obolibrary.org/obo/UBERON_0009048
ERROR     unsatisfiable: http://purl.obolibrary.org/obo/UBERON_0018253
ERROR     unsatisfiable: http://purl.obolibrary.org/obo/UBERON_0009049
```

The result of `robot explain`:

## [inferior external pudendal vein](http://purl.obolibrary.org/obo/UBERON_0009046) SubClassOf [Nothing](http://www.w3.org/2002/07/owl#Nothing) ##

  - [inferior external pudendal vein](http://purl.obolibrary.org/obo/UBERON_0009046) SubClassOf [external pudendal vein](http://purl.obolibrary.org/obo/UBERON_0018253)
    - [external pudendal vein](http://purl.obolibrary.org/obo/UBERON_0018253) SubClassOf [pudendal vein](http://purl.obolibrary.org/obo/UBERON_0009029)
      - [pudendal vein](http://purl.obolibrary.org/obo/UBERON_0009029) SubClassOf [part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [pelvic region of trunk](http://purl.obolibrary.org/obo/UBERON_0002355)
        - [pelvic region of trunk](http://purl.obolibrary.org/obo/UBERON_0002355) SubClassOf [part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [abdominal segment of trunk](http://purl.obolibrary.org/obo/UBERON_0002417)
          - [abdominal segment of trunk](http://purl.obolibrary.org/obo/UBERON_0002417) SubClassOf [subdivision of trunk](http://purl.obolibrary.org/obo/UBERON_0009569)
[...]

PR #2295, opened 2022-02-09
removed reflexive develops from

### LINKED ISSUE #2294: develops_from reflexive cause unsats
Related to https://github.com/obophenotype/uberon/issues/2293

Trying to figure out the above, we realised that unsats only came in when components/reflexivity_axioms.owl is merged in
unstats are all due to develops_from reflexive (when removed unsats go away) 

Questions

- Why does making develops_from reflexive cause unsats
- Do we really need develops_from to be reflexive

shawntanzk:
Fixes #2294
Related to https://github.com/obophenotype/uberon/issues/2293

will definitely wait for @cmungall to comment on this too

### DIFF
--- a/src/ontology/components/reflexivity_axioms.owl
+++ b/src/ontology/components/reflexivity_axioms.owl
@@ -1,4 +1,3 @@
 Ontology(<http://purl.obolibrary.org/obo/uberon/reflexivity_axioms.owl>
  ReflexiveObjectProperty(<http://purl.obolibrary.org/obo/BFO_0000050>)
- ReflexiveObjectProperty(<http://purl.obolibrary.org/obo/RO_0002202>)
 )


shawntanzk:
error:

```
Replaced: 25304
Cloning into 'developmental-stage-ontologies'...
cannot open tmp/mirror-emapa.obo at ../scripts/obo-grep.pl line 62.
Killed
make: *** [reports/basic-allcycles] Error 137
```

shawntanzk:
I've checked and `tmp/mirror-emapa.obo ` is in my tmp folder, and I can open it with protege without an issue

shawntanzk:
not sure if it is `make basic-allcycles` that is failing instead?

dosumis:
Force re-run in case this is a network issue.

matentzn:
That was not a network issue. That sounds like a memory issue. Also, checking this change like this is pretty irrelevant as the relevant bit of code is not executed by QC.. Unless @anitacaron has added the check.

anitacaron:
> error:

```
Replaced: 25304
Cloning into 'developmental-stage-ontologies'...
cannot open tmp/mirror-emapa.obo at ../scripts/obo-grep.pl line 62.
Killed
make: *** [reports/basic-allcycles] Error 137
```

Where is coming to this error from?

matentzn:
Seems to have been a fluke checks passed now.. Lets ignore it :)

shawntanzk:
If it's all good then I'll just wait for @cmungall to give a thumbs up and ill merge in :) thanks!

matentzn:
You will have to share it with him on slack, he is in a big meeting this week and wont read GitHUb emails.

cmungall:
you can go ahead and merge this, I won't block it

It sounds like it was doing work and catching potential issues (but there wasn't examples posted in the issue). I could mount a defense of local reflexivity biologically ontologically and pragmatically (it at least seems odd to leave the part-of in) but the priority here should be simplicity of the overall pipeline, you might want to remove this file altogether, but I leave it up to you all.

shawntanzk:
From the thread, my understanding is we should remove adding reflexive.owl from the pipeline altogether yeah?

matentzn:
Not in the PR, lets just remove the one causing the issue for now.. We are still not sure what the problem is.

shawntanzk:
> lets just remove the one causing the issue for now

In that case, if everyone's good, please give a thumbs up :) thanks!

matentzn (approved):

PR #2299, opened 2022-02-10
Closes #2230 'glomerular visceral epithelium' should be part of (or merged with) glomerular visceral layer

### LINKED ISSUE #2230: 'glomerular visceral epithelium' should be part of (or merged with) glomerular visceral layer
'glomerular visceral epithelium' currently has no definition.

The only reason I can see to keep both terms would be if we need to include basement membrane as part of the layer but not part of the epithelium, however, the basement membrane is currently separate and appears to be a composite structure coming from visceral epithelium and underlying endothelium, so probably not:

https://www.ebi.ac.uk/ols/ontologies/uberon/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FUBERON_0005777

bvarner-ebi:
Closes #2230 'glomerular visceral epithelium' should be part of (or merged with) glomerular visceral layer.

Outcome: Manually 'merged' UBERON:0005751 ‘visceral layer of glomerular capsule’ with UBERON:0006852 ‘glomerular visceral epithelium’ by updating the label of the former to 'glomerular visceral epithelium’ and obsoleting the latter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -174,4 +174,5 @@ src/ontology/basic.json
 src/ontology/basic.obo
 src/ontology/basic.owl
 src/ontology/test_obo_serialiser.sh
+src/ontology/uberon-edit.properties
 src/ontology/.token
--- a/src/ontology/uberon-edit.properties
+++ b/src/ontology/uberon-edit.properties
@@ -1,5 +1,0 @@
-#Fri Feb 04 10:12:43 CET 2022
-jdbc.url=
-jdbc.driver=
-jdbc.user=
-jdbc.password=


balhoff:
I think there's a flaw in this check. Where is it defined?

src/ontology/uberon-edit.obo
```diff
@@ -106123,8 +106123,9 @@ property_value: external_definition "Simple squamous epithelium that lines the c
 
 [Term]
 id: UBERON:0005751
-name: visceral layer of glomerular capsule
+name: glomerular visceral epithelium
 def: "The inner, or visceral, layer of the renal glomerular capsule, overlying the capillaries, composed of podocytes, and separated from the outer, or parietal, layer by the capsular (Bowman's) space[BTO]. Lies just above the thickened glomerular basement membrane and is made of podocytes. Beneath the visceral layer lie the glomerular capillaries[WP]. inner layer of the glomerular capsule surrounding the capillary endothelial cells and composed of highly specialized epithelial cells (podocytes) that are separated by filtration slits; the visceral layer is separated from the outer parietal layer by the urinary space (also known as the Bowman's or capsular space)[MP]." [BTO:0001515, MP:0011497]
```
paolaroncaglia, line 106127: Please check capitalisation and phrasing starting at "inner layer of the glomerular capsule..."

paolaroncaglia (commented):
@bvarner-ebi 
This looks overall ok to me, but to be sure I'd like a second opinion/review.

src/ontology/uberon-edit.obo
bvarner-ebi: I agree this definition should be refined. Do you think the [WP] is a typo and should be [MP] instead? I don't see any reference for WP elsewhere in this class.

paolaroncaglia: @bvarner-ebi [WP] is definitely not a typo, when those [xx] appear in definitions they were usually derived automatically from other ontologies or resources. I strongly suspect that WP stands for Wikipedia. FYI, a list of abbreviations used in Uberon is in the 2012 paper here: https://genomebiology.biomedcentral.com/articles/10.1186/gb-2012-13-1-r5#abbreviations. WP definitely does not refer to the Worm Phenotype Ontology ;-)

paolaroncaglia (commented):
@bvarner-ebi the updated definition reads ok to me. Not sure if the contributor/error issue was addressed though. Thanks.

src/ontology/uberon-edit.obo
bvarner-ebi: Thank you, @paolaroncaglia! :)

bvarner-ebi:
@paolaroncaglia, yes, I deleted the contributor from the obsoleted term (a term unrelated to this PR), and the most recent commit passed the check. I will go ahead and merge.

bvarner-ebi:
@paolaroncaglia, yes, I deleted the contributor from the obsoleted term (a term unrelated to this PR), and the most recent commit passed the check. I will go ahead and merge.

matentzn:
Thank you for fixing the properties  file issue @bvarner-ebi !

bvarner-ebi:
> Thank you for fixing the properties file issue @bvarner-ebi !

Thank you for the instructions, @matentzn.

bvarner-ebi:
Sorry, @paolaroncaglia, requesting one more review since I made revisions to fix the properties file issue, and a review is required before I can merge.

PR #2302, opened 2022-02-11
adding upper tooth tsv

megbalk:
@dosumis could you check these?


rays22 (approved):
I have changed the indentation and white space syntax as suggested by `yamllint`, but otherwise OK.

PR #2316, opened 2022-02-21
Remove restriction to latin-9 character set.

### LINKED ISSUE #2292: Remove character coding restrictions on Uberon
@gouttegd noted:

> Even though the edit file is encoded in UTF-8, we forbid all characters outside of the Latin-9 character set (see uberon.Makefile, line 638)

We should be able to drop this restriction.  I think it dates back to OBO-Edit days and support for legacy DBs.  The world should all be UTF-8 compliant now, and if they're not, it's their problem.

Any objections?  @cmungall ?

gouttegd:
We had a historical restriction that all files in the ontology should only use characters from the Latin-9 character set, even if is encoded in UTF-8. Such a restriction is no longer warranted.

Besides, if the intention behind this restriction was to avoid characters coded on more than 1 byte, then it failed at doing so, as the ontology already contains at least one character outside of the ASCII character set: “ö“ (in the definition of `atlanto-occipital joint`, UBERON:0000220), which is encoded on two bytes in UTF-8 (0xC3 0xB6). The fact that nobody seemingly ever complained about it reinforces the case for dropping the character set restriction.

We keep a check intended to verify that the ontology only contains *valid* UTF-8.

close #2292

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -669,7 +669,7 @@ $(REPORTDIR)/%.obo-OWL-check: %.obo
 
 # test any file for non UTF-8 characters
 $(REPORTDIR)/%-iconv: %
-	iconv -f UTF-8 -t ISO-8859-15 $< > $@
+	iconv -f UTF-8 -t UTF-8 $< > $@
 
 # run subset of syntax and structure checks used by GO
 # (see .travis.yml for dependencies)


gouttegd:
Note: All checks passed, but please do _not_ merge yet!

I will soon submit another PR on top of this one, in which I will randomly introduce non-ASCII characters at several places in the ontology to check more thoroughly that such characters don’t break anything downstream.

shawntanzk:
seems to be looked at here: https://github.com/obophenotype/uberon/pull/2328 is this ready @gouttegd ?

shawntanzk:
100% for merging this

anitacaron (approved):

gouttegd:
OK, then given that nobody voiced any objection in the 20 days since @dosumis raised the issue in #2292, let’s move on and merge. :)

PR #2328, opened 2022-02-25
[DO NOT MERGE] Randomly introduce non-ASCII characters.

gouttegd:
FOR TESTING PURPOSES ONLY. Introduces non-ASCII characters at several places in the ontology to check that such characters don‘t break anything.


gouttegd:
Closing here since the test is done (but keeping the branch alive for a bit, in case someone wants to use it for their own tests).

PR #2331, opened 2022-03-01
[DO NOT MERGE] Add automated diff machine

shawntanzk:
As discussed in CL/Uberon call - implementing automated diff for uberon


shawntanzk:
Error caused by not enough memory

shawntanzk:
oh actually its the classify step - ill up the memory there

matentzn:
Vey interesting. Can you withdraw the diff tool from Uberon for now, and then we will talk during the tech call whether we can actually solve the diff problem - at all!

Issue #2333, opened 2022-03-03
depreciate FOVT lines


megbalk:
**Uberon term**
line from caundal-sacral region of the vertebral column to the last vertebra
line from helix of outer ear to intertragus incisure
line from the distal-most end of the longest nail to the posterior-most part of the foot
line from the distal-most end of the longest phalax of the foot to the posterior-most part of the foot
line from tip of nose to the caudal-sacral region of the vertebral column
line from tip of nose to the posterior-most part of the last caudal vertebra
line from trochlea of humerus to proximal head of humerus
line from trochlea of humerus to ventral tubercle of humerus
line of medial condyle of femur to greater trochanter
line of medial condyle of femur to head of femur

**Reason for deprecation**
@dosumis and I came up with a different pattern such that lines are no longer needed. This is not a common thing people would want to have or search on.

**Suggested replacement or term to consider**
I have noted that lengths of these lines will be in OBA or FOVT.

tail length (http://purl.obolibrary.org/obo/OBA_VT0002758)
ear length to notch (FOVT:0000005)
pes length with nail (FOVT:0000003)
pes length without nail (FOVT:0000004)
body length without tail (FOVT:0000002)
body length with tail (FOVT:0000001)
humerus length from trochlea to caput (FOVT:0000008)
humerus length from trochlea to ventral tubercle (FOVT:0000009)
femur length from greater trochanter to medial condyle (FOVT:0000006)
femur length from caput to medial condyle (FOVT:0000007)

paolaroncaglia:
Hi @megbalk ,
I assigned this ticket to you as I understand that you plan to do the obsoletions, but if not, please assign to others as appropriate.
Unassigned tickets tend to get lost ;-) 
Thanks,
Paola

megbalk:
depreciating more terms:
tip of tail (already exists)
tip of digit (not needed)
posterior of pes

PR #2334, opened 2022-03-03
Fovt depreciate

megbalk:
depreciating lines in uberon and directing to the length of the lines created by FOVT.
Issue #2333


nicolevasilevsky:
@megbalk this PR is failing because the obsoleted terms are referenced in logical axioms.
For example: 'distal-most point of medial condyle of femur' is used in the logical definition for 'distal-most point of medial condyle of femur'.

I think you'll either need to update these logical definitions or remove them.

src/ontology/uberon-edit.obo
```diff
@@ -221742,39 +221739,35 @@ consider: UBERON:0009551
...
 [Term]
 id: UBERON:7500042
-name: line from the distal-most end of the longest phalax of the foot to the posterior-most part of the foot
-def: "An anatomical line of a hindfoot that connects the end of the longest digit to the posterior-most part of a foot." [https://orcid.org/0000-0003-2699-3066]
-intersection_of: UBERON:0006800 ! anatomical line
-intersection_of: connects UBERON:0009551 ! distal segment of digit
-intersection_of: connects UBERON:7500039 ! posterior of pes
+name: obsolete line from the distal-most end of the longest phalax of the foot to the posterior-most part of the foot
+def: "OBSOLETE An anatomical line of a hindfoot that connects the end of the longest digit to the posterior-most part of a foot." [https://orcid.org/0000-0003-2699-3066]
+comment: No longer part of the pattern design for lengths of lines. Can be found in FOVT as pes length without nail (FOVT:0000004).
```
nicolevasilevsky, line 221744: Uberon uses the 'consider' annotation property for terms to consider on obsoleted terms. I think you should add those using that annotation property, rather than mentioning them in the comment. It's okay if the consider points to another ontology term outside of Uberon.

See for example: 'obsolete atrium'

FYI - if you want to consider more than one term, you must have two consider annotations, like in the example above.

megbalk:
added "consider" annotation

megbalk:
@nicolevasilevsky I don't think I understand the errors

nicolevasilevsky:
From what I can tell, the error is caused by an annotation that uses your ORCID. Maybe you are not allowed to have a contributor annotation on obsoleted terms?
I am going to try to remove one and see if that works- standby.

nicolevasilevsky:
Yep, that is the problem, I can fix this for you b/c I have the file open at the moment

src/ontology/uberon-edit.obo
```diff
@@ -221690,30 +221690,27 @@ relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
...
-def: "An anatomical line along a tail that extends from the first caudal vertebra to the last cuadal vertebra." [https://orcid.org/0000-0003-2699-3066]
-intersection_of: UBERON:0006800 ! anatomical line
-intersection_of: connects UBERON:0011271 ! caudal-sacral region of vertebral column
-intersection_of: connects UBERON:7500035 ! tip of tail
-relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
+name: obsolete line from caundal-sacral region of the vertebral column to the last vertebra
+def: "OBSOLETE An anatomical line along a tail that extends from the first caudal vertebra to the last cuadal vertebra." [https://orcid.org/0000-0003-2699-3066]
+comment: No longer part of the pattern design for lengths of lines. Can be found in OBA as tail length (http://purl.obolibrary.org/obo/OBA_VT0002758).
+is_obsolete: true
+consider: http://purl.obolibrary.org/obo/OBA_VT0002758
```
nicolevasilevsky, line 221697: One more minor change is needed:
the full URI for the consider should be added in the IRI editor, not in the Literal tab.

here is a screenshot:
image

Could you please fix these?

thanks!

nicolevasilevsky: the diff should look like this: 
consider: OBA:VT0002758

nicolevasilevsky (changes requested):
One more minor change needed, thanks!

megbalk:
can do! Thanks for the help!

nicolevasilevsky (approved):
looks great!

PR #2336, opened 2022-03-07
Fixes #2335 major and minor kidney calyces miss-classified as renal v…

### LINKED ISSUE #2335: major and minor kidney calyces miss-classified as renal venous blood vessel
image

This is inferred from the domain of 'drains':

image

It's understandable why this relationships was added.  

Short term fix: remove 'drains' relationship.  
Potential longer term fix:  remove  domain of 'drains' in RO.  If the domain of drains is to be kept, we should make more effort to add disjoints that could catch these.

ghost:
Fixes #2335 major and minor kidney calyces miss-classified as renal venous blood vessel

Short term fix:

- [x] remove 'drains' relationships from major and minor calyx terms.

Longer term fixes (for consideration outside of this PR):

- [ ] Check for other potential examples - remove range of drains (for testing purposes only); find terms that are the subject of a drains relationships that are not subclasses of vein.
- [ ] Add high level disjoints to catch issues like this in future
- [ ] TBD - remove domain of 'drains' in RO?


dosumis (approved):

dosumis:
These can be dealt with outside of this PR:

> Longer term fixes:

> Check for other potential examples - remove range of drains (for testing purposes only); find terms that are the subject of a drains relationships that are not subclasses of vein.
>  Add high level disjoints to catch issues like this in future
>  TBD - remove domain of 'drains' in RO?

bvarner-ebi:
Thank you for reviewing, @dosumis.

I also removed the 'kidney calyx' 'part of' some 'renal pelvis' relationship. Do you agree with this?

Blausen_0592_KidneyAnatomy_01
https://en.wikipedia.org/wiki/Renal_calyx

src/ontology/uberon-edit.obo
```diff
@@ -190993,8 +190990,8 @@ name: rima vulvae
 def: "The fissure between the labia majora" [https://www.thefreedictionary.com/pudendal+cleft]
 synonym: "pudendal cleft" EXACT [FMA:19995]
 synonym: "pudendal cleft" EXACT DEPRECATED [doi:10.1002/ca.23705]
-synonym: "rima pudendi" EXACT DEPRECATED [doi:10.1002/ca.23705]
 synonym: "rima pudendi" EXACT LATIN [A09.2.01.006]
+synonym: "rima pudendi" EXACT DEPRECATED [doi:10.1002/ca.23705]
```
bvarner-ebi, line 190994: not sure how these edits occurred... I did not add and delete this

dosumis: I think this is just a serialisation bug => change of line order.  No change to meaning/content of file, so fine to ignore.

src/ontology/uberon-edit.obo
```diff
@@ -113928,7 +113926,6 @@ xref: SCTID:276832000
 xref: UMLS:C0022651 {source="ncithesaurus:Renal_Calyx"}
 is_a: UBERON:0000063 ! organ subunit
 relationship: contributes_to_morphology_of UBERON:0001224 ! renal pelvis
-relationship: part_of UBERON:0001224 ! renal pelvis
```
dosumis, line 113931: I think this should be in a separate PR, as it's a distinct issue and potentially controversial.  ASCT+B tables have this link as well as Uberon

image
See https://hubmapconsortium.github.io/ccf-asct-reporter/vis?selectedOrgans=kidney-v1.1&playground=false

dosumis (changes requested):

src/ontology/uberon-edit.obo
bvarner-ebi: Makes sense. Added back:
 'kidney calyx' part_of some 'renal pelvis'.

dosumis (approved):

PR #2346, opened 2022-03-11
Add human_reference_atlas subset

anitacaron:


dosumis (approved):
I see no problem with this being treated as a component - it specifies a set of axioms we want in the editors file and from there available for all pipelines.  

The component is generated from a Robot template, just not on this repo.  Future task: work out how we will update this in future.  Can/should we pull automatically, e.g. with WGET OR CURL ?

matentzn:
+1 @dosumis I think for full transparency, add a make 2 make goals to uberon.Makefile that:

- mirror the robot template from whatever source
- generate the component from the template

I also would like to not that this PR is doing more than just adding a subset: why are all these "present in taxon" declaration not going straight into Uberon? Seems odd to have them managed elsewhere (also forgot what the purpose of "present in taxon" was).

dosumis:
> I think for full transparency, add a make 2 make goals to uberon.Makefile that:
> 
> * mirror the robot template from whatever source
> * generate the component from the template

+1

> I also would like to not that this PR is doing more than just adding a subset: why are all these "present in taxon" declaration not going straight into Uberon? Seems odd to have them managed elsewhere (also forgot what the purpose of "present in taxon" was).

Background - these are all terms from the Human Reference Atlas (HubMap), so we want to tag all terms with a subset indicating their origin, and also indicate that they are present in humans.  present_in_taxon expansion makes an individual instance of the class with type assertion 'in taxon' some X.  This does work in consistency checking and may also be used in future for creating taxon subsets.

Think about it, the consistency checking, while the check is very weak, could potentially be an issue for a generated file.  Errors would need to be fixed in source.

matentzn:
I want to mention for completeness sake that you are adding an external dependency here that can break the build at release time, but I guess this is not much different from imports.

matentzn:
Actually probably not, but it definitely should. @ubyndr is working on something that would allow this - right now we cannot do this because the memory required for taxon constraint checking is too high for GitHub actions. 

Also, the component needs to be refreshed - QC will never do that all by itself. Only during the workflow that rebuilds the component will that be checked - and at the moment, that happens only at release time.

anitacaron:
I am not sure if it will be possible to mirror the templates and convert them to owl because I need to merge ~28 templates to have the final subset.

I can mirror the final owl file here and put it in the components directory.

dosumis:
Hi @matentzn  - we'd like to push this through ASAP, but agree we should move away from external file dependency. 

General plan for dealing with subsets:

Background: Historically it has been common for external groups to specify and maintain subset tags.  This was (is?) especially common for GO, where these tags are used in enrichment analyses.

Plan: 
1. Move subset tag generation to a robot template  on this repo driving with dynamic generation of component.  These are easier to review than whole OWL files and less likely to add problematic axioms. We currently generate this OWL file with an ROBOT template on another repo, so easy for us to switch.
2. Further down the line I think we should consider some generic support for adding subset tags, built into ODK  - allowing external users to specify a tag name and a list of terms.  This would be turned into a ROBOT template driven component.

**We'd like to get these tags out before the next release.  Would you be happy to merge as-is for now and make a new ticket to turn this into a dynamically generated component?**

Issue #2025, opened 2021-08-04
Dont use species specific anatomy terms in Uberon?


matentzn:
Someone used http://purl.obolibrary.org/obo/FBbt_01000119

![image](https://user-images.githubusercontent.com/7070631/128206622-93d5a164-132a-4e99-9456-9d3103e11086.png)

What is the policy here?

matentzn:
@gouttegd do you agree the best way here is to:

- [ ] Create a new Uberon class with the same scope as the FBbt class
- [ ] Map new Uberon class to FBbt class
- [ ] Have as policy not ever to have any species specific identifiers in Uberon?

paolaroncaglia:
@matentzn 

> * [ ]  Have as policy not ever to have any species specific identifiers in Uberon?

In principle I agree, with the exception of terms that are clearly taxon-specific but whose labels are ambiguous, usually because they come from species-specific ontologies. See e.g. UBERON:6003624, previously labelled 'adult brain' (xref FBbt:00003624 and child of UBERON:6003007 'insect adult head'), and renamed 'insect adult brain' by Chris to disambiguate (https://github.com/obophenotype/uberon/issues/1759). For its nature, Uberon contains many terms that may have that issue. Some are being reviewed as part of producing the human slim etc, but I suspect that there are still cases where it'd be helpful to make the label species-specific in addition to taxon constraints. Years ago we had a similar discussion in GO for dinoflagellate-specific terms that had ambiguous labels, and Chris suggested to make the label species-informative (https://www.ebi.ac.uk/ols/search?q=dinoflagellate&ontology=go).

paolaroncaglia:
@matentzn also recently in CL we agreed to let some terms be labelled 'human' (or similar) because they do represent human-specific types and it helps to have that info made explicit for those particular cells in that branch, see https://github.com/obophenotype/cell-ontology/issues/1279#issuecomment-966378826.

gouttegd:
@paolaroncaglia My understanding of @matentzn ’s third point was that it is about using a species-specific term in the definition of an Uberon term or in a relationship with a Uberon term. That is, a Uberon term should never be a `SubClassOf`, be a `part_of`, or more generally have any kind of relationship with a term from a species-specific ontology (as in this example, where UBERON:600094 is `part_of` FBbt:01000119).

I would agree with such a policy – if this was indeed what Nico had in mind. This would have no bearing on using taxon-specific _labels_.

gouttegd:
[...]

PR #2347, opened 2022-03-11
Obsolete children of `insect anlage in statu nascendi`.

### LINKED ISSUE #2154: [NTR] anterior ectoderm anlage
uberon currently has a floating term from FBbt: 
http://www.ontobee.org/ontology/FBbt?iri=http://purl.obolibrary.org/obo/FBbt_01000119

It is currently use in:
'clypeo-labral anlage in statu nascendi' part of FBbt 01000119

If this is right, this term needs to be naturalised into uberon 

@gouttegd hope it's alright if I assign you, happy for you to reassign to someone else from the fly world - i have zero idea on fly stuff so can't help much here >.<

gouttegd:
The two terms `clypeo-labral anlage in statu nascendi` and `visual anlage in statu nascendi` are way too specific to belong in Uberon. Having the parental term `insect anlage in statu nascendi` (mapped to the corresponding term in FBbt) should be enough--anyone looking for a more precise term should go look for the children of the FBbt term.

Obsoleting `clypeo-labral anlage in statu nascendi` allows to remove the FBbt term `anterior ectoderm anlage`, which was imported into Uberon since `clypeo-labral anlage in statu nascendi` has a part_of relationship to it. (As stated in #2025, we do not want any Uberon term to have any kind of relationship to a species-specific foreign ontology.)

Obsoleting these two terms has cascading consequences:

* `insect clypeo-labral anlage` is the starting point of a chain of `develops_from` relationships that goes up to `insect clypeus`;
* similarly, `insect visual anlage in statu nascendi` is the starting point of a chain that goes up to `insect Bolwig organ`.

It is doubtful that all these terms are really useful, but I don’t want to obsolete them all for now, so this PR simply breaks the chains at their beginnings by removing the first `develops_from` axioms. Deciding what to do with all the "insect anlage" and "insect primordium" terms will need to be done later (probably as part of a larger review of all the 6xxxxxx terms).

close #2154


src/ontology/uberon-edit.obo
```diff
@@ -219746,11 +219746,9 @@ is_a: UBERON:6007248 ! insect chorionic specialization
 
 [Term]
 id: UBERON:6000094
-name: clypeo-labral anlage in statu nascendi
-def: "." [FBC:VH]
-synonym: "Asn Cly" RELATED [FBC:DOS]
-is_a: UBERON:6005413 ! insect anlage in statu nascendi
-relationship: part_of FBbt:01000119
+name: obsolete clypeo-labral anlage in statu nascendi
+comment: No need for such a specific term in Uberon.
```
shawntanzk, line 219750: could you put a link to this PR in the comment please - think we are trying to do that with obsoletions now.

src/ontology/uberon-edit.obo
```diff
@@ -219746,11 +219746,9 @@ is_a: UBERON:6007248 ! insect chorionic specialization
 
 [Term]
 id: UBERON:6000094
-name: clypeo-labral anlage in statu nascendi
-def: "." [FBC:VH]
-synonym: "Asn Cly" RELATED [FBC:DOS]
-is_a: UBERON:6005413 ! insect anlage in statu nascendi
-relationship: part_of FBbt:01000119
+name: obsolete clypeo-labral anlage in statu nascendi
```
shawntanzk, line 219749: not 100% sure how this will affect mapping component? or if mapping is sufficient that you don't need to put a 'item replaced by' annotation (personally I would still add it, pointing to the FBbt term)

shawntanzk (changes requested):
Generally looks ok, have some suggestions partly based on last obsoletion training session in line.

src/ontology/uberon-edit.obo
gouttegd: I hope you mean “a link to the original ticket” (here #2154) and not a link to the _PR_… Otherwise that would mean that for any obsoletion one need to first edit the ontology file to obsolete the term (without adding a link to a PR that does not yet exist), create the PR, get the PR number, edit the ontology file again to add the link, and update the PR… I quite don’t see the benefit of such a cumbersome approach.

src/ontology/uberon-edit.obo
gouttegd: Even if the mapping between the (now obsolete) Uberon term and the corresponding FBbt term still exists (because Uberon/FBbt mappings are managed on the FBbt side, so the mapping will exist until I update FBbt accordingly, and until those mappings are refreshed in Uberon), obsolete terms are ignored when generating the bridge files.

src/ontology/uberon-edit.obo
shawntanzk: oh yeah if there is a ticket, that's perfect :) its mainly to keep discussion on why the obsoletion is done - I saw the write up on this PR so thought it was all here.

shawntanzk (approved):

Issue #2268, opened 2022-01-31
New policy proposal to deal with PRs: absence of objection is tacit approval


gouttegd:
_Note: The following proposal would apply to CL as well as Uberon._

We still regularly have pull requests that are several weeks or even several months old and that are waiting to be merged because nobody jumped in to review them.

The “shepherding“ system put in place last summer is not enough, at least in part, I believe, because the “domains of responsibility” of available reviewers (as decided in the Uberon call of July 26, 2021) are quite vague and therefore many PRs end up being assigned to the same two or three reviewers (typically Chris or David).

I would thus like to suggest a new approach, based on Grace Hopper’s advice that it is “easier to ask for forgiveness than to get permission”, in which the author of a PR would be tacitly allowed to merge it after a certain period of time (I tentatively propose one month, though this is debatable) if _nobody_ raised any concern on that PR since its submission.

This would _not_ be automatic, but left at the discretion of the PR author. If they are really not sure whether what they did is the best solution to the issue they were trying to tackle, they may still choose not to merge until they got an _explicit_ green light from a reviewer.

Worst-case scenario, this approach may lead to a PR with a questionable change being merged to master. But then, once the change is in master, if it is really problematic then it will (finally) be noticed and a corrective action can be taken.

I believe such an approach is more suitable to projects like Uberon or CL, where there’s no single official “project lead” to whom we can defer every single decision.

matentzn:
Great!

PR #2351, opened 2022-03-15
Automatically ping PRs after one month of inactivity.

gouttegd:
Update the 'stale-issues' workflow to implement the decision mentioned in #2268.

Any pull request that has not seen any activity in one month will be tagged with `unreviewed` and a comment will be added with the following message:

"This PR has not seen any activity in the past month; if nobody comments or reviews it in the next week, the PR editor will be allowed to proceed with merging without explicit approval, should they wish to do so."

Pull requests may be explicitly marked with a tag `approval required` to prevent them from receiving this comment. This way, a PR editor (or anybody else) can express that they do wish the PR to be explicitly approved.

closes: #2268

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -11,9 +11,13 @@ jobs:
       - uses: actions/stale@v4
         with:
           days-before-issue-stale: 180
-          days-before-pr-stale: -1
           days-before-issue-close: 365
           close-issue-label: 'autoclosed-unfixed'
           stale-issue-message: 'This issue has not seen any activity in the past 6 months; it will be closed automatically in one year from now if no action is taken.'
           close-issue-message: 'This issue has been closed automatically because it has not been updated in 18 months. Please re-open if you still need this to be addressed.'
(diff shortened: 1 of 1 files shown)


shawntanzk (approved):
love it! was wondering if we could have documentation for this on uberon. I notice uberon docs are a bit different from CL so not 100% sure how to add uberon specific stuff.

matentzn:
I love this!

gouttegd:
@matentzn 
> I notice uberon docs are a bit different from CL so not 100% sure how to add uberon specific stuff.

Of note, the plan is to do the same for CL afterwards. ;) But since I have more interest in CL than in Uberon (I work on single-_cell_ RNA-seq after all), I tend to use Uberon as my testing ground. :D

shawntanzk:
> the plan is to do the same for CL afterwards

happy to write up docs for CL when you make that PR on CL :) thanks @gouttegd

gouttegd:
I’d suggest to wait for a few weeks to see this workflow in action on Uberon and be sure we are all happy with this new way of dealing with PRs, and then I’ll port the change over to CL.

PR #2361, opened 2022-03-25
Fix label and synonyms for UBERON:8410070

paolaroncaglia:
Addresses #1904


paolaroncaglia:
@matentzn are you releasing Uberon right now or is it ok if I push this commit in, please? Ta.

PR #2364, opened 2022-03-25
Pre Release 2022 03 25 FIXES

matentzn:
## Under construction

Some errors slipped into Uberon due to missing QC, and had to be fixed first. The QC is part of @cmungall old battery of tests which I am introducing when I am done fixing.

## Blocked by

- [x] #2368
- [x] #2367 
- [ ] When refreshing imports, 1000s of unsatistifiable classes happen, need to investigate why


matentzn:
@sbello you are a star thank you! Can you suggest a better relationship to use then? I am not really versed on these anatomy relations..

sbello:
Given that the sinus = space I don't know that there is an adequate way to relate the ceiling to the space. Even assuming adjacent_to can be used between a material and immaterial entity it loses that this is the top portion and we will probably need to enter a term for the floor at some point. 
You can say that this is part_of the lymph node, that much is accurate.
Note there is inconsistency in Uberon as to whether X sinus refers to the space or the tissue surrounding the space. This is probably due to the way biologists name things but it does make finding relevant examples tricky.
Sorry not to give you a definite answer.

matentzn:
@shawntanzk @dosumis @cmungall Release is blocked until someone can respond to @sbello who prevented me from committing some hasty fix :) Please advice.

matentzn:
We will go the minimal route, closing this in favour of a clean PR: https://github.com/obophenotype/uberon/pull/2376/files

PR #2371, opened 2022-03-28
subcapsular sinus floor and ceiling fix

shawntanzk:
This is my suggested fix.
I've reverted it back to organ part and add an adjacent to axiom.
I don't think we need to add logical axioms about ceiling/floor, I've taken a look at a few terms like that and they don't seem to have it. 
Thoughts on this @paolaroncaglia @sbello


src/ontology/uberon-edit.obo
```diff
@@ -67,6 +67,7 @@ import: http://purl.obolibrary.org/obo/uberon/imports/ncbitaxon_import.owl
 import: http://purl.obolibrary.org/obo/uberon/imports/pato_import.owl
 import: http://purl.obolibrary.org/obo/uberon/imports/pr_import.owl
 import: http://purl.obolibrary.org/obo/uberon/imports/ro_import.owl
+import: http://purl.obolibrary.org/obo/uberon/imports/bspo_import.owl
```
shawntanzk, line 70: whoops, not sure if we still want this in, was trying to use upper from BSPO, but its a class and not an object property so no idea how to use it

paolaroncaglia:
@shawntanzk thanks, let me double-check the original publication to see how to we can best address this - it's [this figure](https://www.cell.com/immunity/fulltext/S1074-7613(19)30297-3?_returnURL=https%3A%2F%2Flinkinghub.elsevier.com%2Fretrieve%2Fpii%2FS1074761319302973%3Fshowall%3Dtrue) (really hard to see)

paolaroncaglia:
I'd say that the SBS ceiling is part_of the capsule, and the SCS floor is part of the cortex. Triple-checking now...

paolaroncaglia:
For a quick fix I suggest:

'subcapsular sinus ceiling' 
subclass of 'organ part' 
part of 'capsule of lymph node'
"The section of the lymph node capsule that faces the subcapsular sinus. In human, it's characterized by specific protein markers."

'subcapsular sinus floor' 
subclass of 'organ part' 
part of 'cortex of lymph node'
"The section of the lymph node cortex that faces the subcapsular sinus. In human, it's characterized by specific protein markers."

paolaroncaglia:
The quick fix above doesn't include a relationship to subcapsular sinus if adjacent_to won't work ('surrounds' does not convince me 100% and I can't find good usage examples). Also, we could do with exact synonyms to clarify the meaning of those classes... 'subcapsular sinus ceiling' really means 'floor of capsule of lymph node', but that might be confusing

shawntanzk:
> if adjacent_to won't work 

Don't see why this won't work. I know @sbello mentioned about a material entity being adjacent to an immaterial entity, but I'm looking at the range and domain of adjacent to, and both of them are independent continuant, which allows for both material and immaterial entity.

sbello:
I was uncertain about the use of adjacent to with immaterial and material entities I think it makes sense but sometimes RO has logic restrictions that I miss.

I agree that the terms could go in as part of relations as @paolaroncaglia suggested above. We could then take some time to figure out the full eq without holding up the use of the terms.

PR #2376, opened 2022-04-01
Release 2022 03 31 (easier review)

matentzn:


shawntanzk:
removed anatomical entity doesn seem a big deal, as they have is_a axioms to something that is basically an anatomical entity, im guessing that this is because of the changes in bridge equivalence to caro - don't think its a huge deal

shawntanzk:
might be of concern, left is this branch, right is last release - GCIs seem to be removed
Screenshot 2022-04-01 at 16 33 12

shawntanzk:
I'm looking at a few classes with 'structure with developmental contribution from neural crest' - and it seems that they can be inferred some other way, so it makes sense that those axioms are removed. The question I have is why is it happening this time?

matentzn:
This is super weird! So many GCIs removed does not smell good. Can we somehow figure out which file they were supposed to be in? originally? (The GCIs cannot be in Uberon..). Is it possible they came back through some import in the past?

shawntanzk:
> This is super weird! So many GCIs removed does not smell good. Can we somehow figure out which file they were supposed to be in? originally? (The GCIs cannot be in Uberon..). Is it possible they came back through some import in the past?

oh that explains it, sorry lol, was using the edit file, should probably have used the built uberon >.< just checked all good

shawntanzk:
left to do: see if epithelium stuff is also just removing things that can be inferred

shawntanzk:
It seems like this new release removed many axioms that can technically be inferred: 
Screenshot 2022-04-04 at 08 17 35
Screenshot 2022-04-04 at 08 18 42

I'm not sure why this is done as the inferences are pretty complex (like you need to know what you are looking for for that) Need input on how to proceed.

shawntanzk:
Potential reason for massive changes: https://github.com/obophenotype/cell-ontology/issues/1532

balhoff:
I agree, it sure took a while! But the imports have been a tangle. There are some axioms that only live in imports and get passed back and forth.

matentzn:
@shawntanzk can you do the final review here and approve!

matentzn (commented):
I approve this!

shawntanzk (approved):

matentzn:
For future reference, this is yet undadressed: https://github.com/obophenotype/uberon/issues/2382

PR #2389, opened 2022-04-08
[Do not merge] breaking zfa taxon constraint

shawntanzk:
Added a ZFA xref to nipple and feather


shawntanzk:
@matentzn - looks like the check works, it is breaking

shawntanzk:
ah dang, i saw failed and just thought that was what i expected, but yes, it is killed >.<

PR #2390, opened 2022-04-08
Add taxon-specific GCIs for existential relations.

### LINKED ISSUE #2348: replace metazoan-specific relationships on generic caro-equivalent structures with taxon GCIs
There are a broader set of issues I will link to later about whether or not it is a good idea to have

uberon:anatomical_structure = caro:anatomical_structure and part-of some metazoa

for now people are using the uberon classes as if they were more general. Issues arise from this e.g. with starts-and-ends to zygote for anatomical structure where anatomical structure is pan-cellular organisms.

Let's replace this with taxon GCIs while we figure out what the longer term solution is

Note that @fbastian relies on the stages, even for generic terms I believe. But the stages will 'come back' in the composite-metazoan release

balhoff:
Fixes #2348.


src/ontology/uberon-edit.obo
```diff
@@ -7376,7 +7376,7 @@ xref: XAO:0003037
 xref: ZFA:0001488
 is_a: UBERON:0010000 {source="CARO"} ! multicellular anatomical structure
 relationship: has_part UBERON:0000479 ! tissue
-relationship: part_of NCBITaxon:6072 ! Eumetazoa
+relationship: part_of NCBITaxon:6072 {gci_filler="NCBITaxon:6072", gci_relation="in_taxon"} ! Eumetazoa
```
cmungall, line 7379: this was an odd axiom to begin with.. but I guess the GCI is not wrong

cmungall (approved):

Issue #2354, opened 2022-03-17
Identified 20 nodes with invalid xrefs


cthoyt:
I'm putting together a tool for automatic analysis of the xrefs in a given ontology by using the Bioregistry. When I applied it to UBERON, it identified the following xrefs whose values were not CURIEs. Perhaps when this happens it means it's supposed to be a RO term? It seems to only be getting this issue for relations besides the two entries imported from CL

| node_id                                                                         | xref                                 |
|---------------------------------------------------------------------------------|--------------------------------------|
| http://purl.obolibrary.org/obo/CL_0008047                                       | H3.03.00.0.00012                     |
| http://purl.obolibrary.org/obo/CL_4023076                                       | nifext_55                            |
| http://purl.obolibrary.org/obo/nbo#by_means                                     | by_means                             |
| http://purl.obolibrary.org/obo/nbo#has_participant                              | has_participant                      |
| http://purl.obolibrary.org/obo/uberon/core#anteriorly_connected_to              | anteriorly_connected_to              |
| http://purl.obolibrary.org/obo/uberon/core#channel_for                          | channel_for                          |
| http://purl.obolibrary.org/obo/uberon/core#channels_from                        | channels_from                        |
| http://purl.obolibrary.org/obo/uberon/core#channels_into                        | channels_into                        |
| http://purl.obolibrary.org/obo/uberon/core#conduit_for                          | conduit_for                          |
| http://purl.obolibrary.org/obo/uberon/core#distally_connected_to                | distally_connected_to                |
| http://purl.obolibrary.org/obo/uberon/core#filtered_through                     | filtered_through                     |
| http://purl.obolibrary.org/obo/uberon/core#posteriorly_connected_to             | posteriorly_connected_to             |
| http://purl.obolibrary.org/obo/uberon/core#protects                             | protects                             |
| http://purl.obolibrary.org/obo/uberon/core#proximally_connected_to              | proximally_connected_to              |
| http://purl.obolibrary.org/obo/uberon/core#subdivision_of                       | subdivision_of                       |
| http://purl.obolibrary.org/obo/uberon/core#transitively_anteriorly_connected_to | transitively_anteriorly_connected_to |
| http://purl.obolibrary.org/obo/uberon/core#transitively_connected_to            | transitively_connected_to            |
| http://purl.obolibrary.org/obo/uberon/core#transitively_distally_connected_to   | transitively_distally_connected_to   |
[...]

PR #2409, opened 2022-04-11
Update NBO import

anitacaron:
The xref for the properties listed in #2354 was being imported from NBO.

I updated the import and now the xrefs are gone.


matentzn (approved):
So much trash being important. We need to move to base soon..

PR #2413, opened 2022-04-11
Updating pro import

### LINKED ISSUE #2373: OBO file uses undeclared synonym type def
`PRO-short-label` is used but not declared. This causes some OBO parsers to fail (correctly). Another I saw is `PRO-common-name`. These are properly declared in PRO.

shawntanzk:
Fixes #2373
@matentzn - doesnt seem to fix after updating pro -> not sure if it got updated right? (I update mirror and all) could you help me check if the diff has the changes?


matentzn (commented):
This does not seem to solve the issue.. Hmmmm.. How did no one ever notice that?

matentzn:
I provided an ODK fix..

We should prioritise moving Uberon to base module system..

matentzn:
@shawntanzk How did you verify this? I checked the source, and it looks correct (I also tagged you on the line so you can see it)

shawntanzk:
> @shawntanzk How did you verify this? I checked the source, and it looks correct (I also tagged you on the line so you can see it)

Oh I thought you wanted to remove it, I just checked in protege if it was still there. But I see that you changed it to a has_synonym_type AP

matentzn:
I dont know whether I ever explained the problem to you: Basically OBO format has this thing where it expects synonym types and subsets to be sub-annotation-properties of some specific parent. But the ROBOT module extraction process does not know about that - and `robot extract` drops this connection (the parentage). When this happens, OBO format does not know what to do anymore. Thats why we need to inject these sub-properties back in with SPARQL postprocessing. IIIIII HACK

shawntanzk:
btws is this ready to merge? also do we have to do a release for this?

matentzn:
Yeah, lets do another release on Friday. It will be easier now.

Issue #2367, opened 2022-03-26
Taxon constraint violation after import updates


matentzn:
<img width="747" alt="image" src="https://user-images.githubusercontent.com/7070631/160240263-1e3d06b3-ef5b-4cc9-aff8-4450ce6db40a.png">

Blocks release #2364

shawntanzk:
dont think I can be super helpful here, I have like ZERO idea bout fish let alone their eyes.
My guess is that jugular vein part_of some neck is over-restrictive. 
Did a quick search and it seems fish do indeed have jugular veins and im guessing fishes don't have necks

shawntanzk:
I would think removing the part_of neck would be the right thing to do, but I don't know how that would affect modelling, I'm guessing there's a reason that axiom is there? I guess inherent difficulty with multi-species anatomy ontology

shawntanzk:
> I would think removing the part_of neck would be the right thing to do

shall I do this in the branch @matentzn

matentzn:
Can you branch off from my branch and then make a pr to the branch? Same with the other ticket. You may have to undo changes I made previously

matentzn:
Hmm.. Still a problem:

## [ZFA_0005829](http://purl.obolibrary.org/obo/ZFA_0005829) SubClassOf [Nothing](http://www.w3.org/2002/07/owl#Nothing) ##

  - [ZFA_0005829](http://purl.obolibrary.org/obo/ZFA_0005829) EquivalentTo [uterine horn](http://purl.obolibrary.org/obo/UBERON_0002247) and ([part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [Danio](http://purl.obolibrary.org/obo/NCBITaxon_7954))
    - [Danio](http://purl.obolibrary.org/obo/NCBITaxon_7954) SubClassOf [Danioninae](http://purl.obolibrary.org/obo/NCBITaxon_2743711)
      - [Danioninae](http://purl.obolibrary.org/obo/NCBITaxon_2743711) SubClassOf [Danionidae](http://purl.obolibrary.org/obo/NCBITaxon_2743709)
        - [Danionidae](http://purl.obolibrary.org/obo/NCBITaxon_2743709) SubClassOf [Cyprinoidei](http://purl.obolibrary.org/obo/NCBITaxon_30727)
[...]

PR #2415, opened 2022-04-13
Add Mammalia Taxon GCI for jugular vein to be partof neck

shawntanzk:
See https://github.com/obophenotype/uberon/issues/2367#issuecomment-1096815312


matentzn:
Sorry not the right person to review this. :/ Can help you find reviewers if you need!

shawntanzk:
@matentzn - no worries, put you as reviewer as the original ticket was opened by you.
Either way I will not go ahead with this unless @cmungall gives an explicit positive review here. (I know long term we would want to make neck more generalizable)

cmungall (approved):
Let’s go with it but let’s not forget to implement the better solution later

Issue #2422, opened 2022-04-15
http://purl.obolibrary.org/obo/uberon/subsets/life-stages-composite.obo has disappeared


cmungall:

matentzn:
These are the subsets that have not been included in the last years:

https://github.com/obophenotype/uberon/tree/master/subsets

Any subset/release you want included needs to be added there. If you make a pr on that config file, I can take care of the rest for you.

cmungall:
what is the conf file?

matentzn:
https://github.com/obophenotype/uberon/blob/master/src/ontology/uberon-odk.yaml#L65

matentzn:
Recently added https://github.com/obophenotype/uberon/releases/tag/v2022-05-27 (#2425) is empty.

Needs to be investigates.

shawntanzk:
just fyi, might take awhile for me, im on leave for a bit, will only do when i have time

anitacaron:
@ubyndr will try to investigate

matentzn:
Same is true for subsets/life-stages-minimal.tsv. The problem is that GH (the github client does not allow uploading empty files).

fbastian:
It would be great if it could be fixed before the start of the ISMB conference (July 10), since this file is explicitely linked in an accepted short paper.

PR #2425, opened 2022-04-18
adding life-stages-minimal subset

cmungall:
For #2422


shawntanzk:
@matentzn think it should be a straightforward merge, but didn't want to do it without you given that you commented on the ticket. Thanks

matentzn:
@shawntanzk can you:

- [ ] Run `sh run.sh make update_repo` on this branch, commit changes
- [ ] Make sure there is a goal like this in uberon.Makefile:

```
subsets/life-stages-minimal.owl: subsets/merged-partonomy.owl
	$(eval TERM_ID := $(TERM_appendicular))
	$(SUBSETCMD)
```

I do not know what `TERM_appendicular` should be? Maybe of you scroll a bit up and down in the Makefile you can guess it.. A new variable like `$(TERM_lifestages)` probably, not sure what the id is but maybe there is a "life stage" class in Uberon?

shawntanzk:
diff.yml now gets added when I update repo, I assume you updated the ODK build for this? I have removed it from here as it won't work as is for UBERON just wanted to check. Do note that this also  downgrades from 1.3.1 to 1.3.0 - the changes seem negative :( Happy to revert till 1.3.1 fully out?

matentzn:
I updated it from my machine, can you pull?

Let's hope it just works..

matentzn:
I shared it in uberon channel again to have chris take a look. It seems ok to me.

matentzn (approved):

PR #2432, opened 2022-04-23
Add new MESH xrefs

bgyori:
Following up on #2431, I created a further set of mappings using Gilda, this time for UBERON entries that _didn't yet_ have a MeSH xref. I then used Biomappings (https://github.com/biopragmatics/biomappings) to review and curate 130 of these mappings as correct. This PR adds these 130 new MeSH xrefs. I also exported a table of these mappings to help review what these are: https://github.com/bgyori/uberon/blob/mesh_review/uberon_mesh_new.tsv.


dhimmel:
Thanks @bgyori... this is awesome. Dropping a permalink to [`uberon_mesh_new.tsv`](https://github.com/bgyori/uberon/blob/3f3fe53c6b8d3ed46e93d783962dda614ab6f7f4/uberon_mesh_new.tsv) to prevent rot (via [y hotkey](https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files)).

matentzn (approved):
I looked the mappings over a bit hastily but all that I saw looks great. Thank you so much @bgyori !

PR #2435, opened 2022-04-25
Uberon adding additional QC checks

matentzn:


matentzn:
@shawntanzk I had to delete two totally misplaced subset declarations, they will have never made a difference anywhere, but feel free to make a ticket to fix. I would just sweep them under the rug.

shawntanzk:
Love the adding of more QC checks - happy to merge if ready?

shawntanzk (approved):

PR #2443, opened 2022-04-29
2436 add automated diff to uberon

### LINKED ISSUE #2436: Add automated diff to uberon
now that we have figured a way to do this, we should implement this

ubyndr:
Fixes #2436 

Adding automated diff report for both unreasoned version and reasoned version. I have used current branch as left input and master branch as right input in the robot diff command. I have also used uberon-edit.obo for unreasoned version and uberon-simple.owl for reasoned version, as @shawntanzk suggested. 

Would be good if make steps and robot diff part is reviewed

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Diff report
         run: |
           cd src/ontology
-          docker run -v $PWD/../../:/work -w /work/src/ontology -e ROBOT_JAVA_ARGS='-Xmx9G' -e JAVA_OPTS='-Xmx9G' --rm obolibrary/odklite robot diff --left uberon-simple.owl --left-catalog catalog-v001.xml --right ../../master/src/ontology/uberon-simple.owl --right-catalog ../../master/src/ontology/catalog-v001.xml -f markdown -o simple-report.md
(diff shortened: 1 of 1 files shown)


.github/workflows/diff.yml
```diff
@@ -0,0 +1,84 @@
+name: 'Create ROBOT diffs on Pull requests'
+
+on:
+  # Triggers the workflow on pull request events for the master branch
+  pull_request:
+    branches: [ master ]
+    paths:
+      - 'src/ontology/uberon-edit.obo'
```
shawntanzk, line 8: I assume this means it only runs when that file is edited? Wonder if we should keep this or get it to run every time (given it runs on reasoned file, albeit -simple, we might want to run this every time so that when other files that change that affects it, this will show). Thoughts @matentzn?

matentzn: We could add all the files that can potentially affect the diff in here.. Remeber this runs for quite long - so would be better if we could somehow rain in the runtime when people just change the docs..

shawntanzk: ill run a test to make sure everything works well :)

shawntanzk: sweet! works nicely! thanks @ubyndr! This will be SO helpful! 
btws took `1h 13m 47s` to run :D

shawntanzk (approved):

matentzn (approved):

PR #2450, opened 2022-05-03
Major fixes to visual areas in the brain

### LINKED ISSUE #2412: Bug category: is linking functional brain description with anatomical brain regions an error?
In preparing material from the ferret brain to be added to Uberon (issue #2352), we came across a possible error in UBERON brain design, or at least a question on how to specify it. (@shawntanzk )

In some animals, functional areas are grouped by their association with an anatomical feature of the brain, such as a gyrus, sulcus, or lobe. We've run into a problem in that these relationships are not invariant across species. For example, in humans, primary visual cortex, secondary visual cortex, and visual association cortex are all classified as “part of” visual cortex (UBERON_0000411), where visual cortex is “part of” occipital cortex (UBERON_0016540). 

However, in ferret, there exists several functional areas that should be classified as “part of” visual cortex by their function, but are located in temporal or parietal regions and thus should not be considered “part of” occipital cortex.

As an example, ferret areas 20a (temporal visual area a), 20b (temporal visual area b) and the posterior suprasylvian area (PS) are identified as being “temporal” in their anatomical position and have visual function (Manger, Nakamura, Valentiniene & Innocenti, 2004; Hommona-Ludiye, Manger & Bourne, 2010).

Presumably, the relationship between "visual cortex" being "part of" "occipital lobe" comes from the human or monkey, where most (but not all) of the visual areas are in the occipital lobe.

As another example, the posterior parietal rostral cortical area (PPr) and posterior parietal caudal area (PPc) are visual in function and defined as being “parietal” anatomically (Manger, Masiello & Innocenti, 2002; Dell, Innocenti, Hilgetag, & Manger, 2019).

Here we are a little stuck. It would seem the right thing to do would be to somehow clarify that "visual cortex" is "part of" "occipital lobe" only in humans or primates. Ferrets don't have formal lobe definitions that we can uncover.

The problem: if someone searches for all areas that are “part of” occipital cortex, and if we say that these areas in ferret temporal cortex are “part of” “visual cortex”, then, because “visual cortex” is part of “occipital cortex”, all of these ferret temporal areas will be identified.

Perhaps the fix is to say “visual cortex” “overlaps” "occipital cortex” ?? That is true in all mammals. That would be a big in the current UBERON for the big sensory functional categories.

An alternative would be to modify it so that being "part of" certain anatomical regions needs some sort of "in taxa" modifier to be right. Is that a better way or is the better way to say “overlaps”?

Thanks
Steve

**Uberon term**

**Bug description**

shawntanzk:
1) Created term cortical visual area
2)  Broaden primary visual cortex by removing human specific stuff
3) Create a new term Broadmann area 17 by splitting out human specific stuff in visual cortex
4) added capable of part of some visual perception to all visual areas in brain

Fixes #2412

@dosumis, @stevevanhooser, @cmungall, @daniscafidi - could you take a look at this PR and make sure you agree with the changes? thanks :)


shawntanzk:
PS: if you wait an hourish, there should be an easier to review ROBOT diff on this that will have more human readable change log that includes reasoned changes.

shawntanzk:
error: `make: *** [Makefile:170: reports/uberon-edit.obo-obo-report.tsv] Error 1` - will figure whats breaking this

shawntanzk:
`make: *** [uberon.Makefile:754: reports/uberon-edit-xp-check] Error 1`

shawntanzk:
@stevevanhooser, @daniscafidi, & @cmungall - could I get a thumbs up on this if you are good with this solution? want to make sure the template you all curated doesn't slip through the cracks but I think it requires this to be agreed upon first.

stevevanhooser:
Hi @shawntanzk , is there a way to preview what was done? Sorry to be clueless... Thanks!

shawntanzk:
> is there a way to preview what was done? 

Yes! you can look at the automated diff comment (github-actions comments) above to see a human-readable (sort of) version of what we changed. The edit file diff is what I made changes to, and the classified ontology one is how it affects the reasoned ontology. If you want to view it graphically, I guess the only way is to check out the branch and open the edit file in something like protege.

stevevanhooser:
This looks good to me @shawntanzk , thanks!

dosumis:
Looks good.  I worry a bit that occiptal lobe is meant to apply in species where there is not a distinct occipital lobe (did I get that right?) but OK with it if @stevevanhooser considers a good/non-confusing compromise.

One other question, do we get any automated classification to 'cortical visual area'? Expected to see that as superclass to PVC and new areas coming from @stevevanhooser (although perhaps those new areas still to be added?)

shawntanzk:
>  new areas still to be added?

new areas are not added yet (see [googlesheets](https://docs.google.com/spreadsheets/d/1M7QLZFFs0w78SPWvUytGEABSAxLoazzuB34HmcTz9Hk/edit#gid=0)) - I'm waiting for this PR to go through to get to it, would rather see how the new terms fall after merging template in so i can make edits needed immediately.

stevevanhooser:
@dosumis I think the occipital lobe comment is well put..even in species where there is no agreed occipital lobe division, sometimes people still use the term. So I think it is the right comment.

Issue #2186, opened 2021-11-23
UBERON:0008339 'microvascular endothelium' isn't a subclass of UBERON:0004638 'blood vessel endothelium'


paolaroncaglia:
UBERON:0008339 'microvascular endothelium' 
and
UBERON:0004638 'blood vessel endothelium'
are siblings currently. The former is undefined both textually and logically, and xrefs BTO:0003122 'microvascular endothelium'. The latter is better defined and xrefs BTO:0000766 'blood vessel endothelium'. "microvasculature" means "tiny blood vessels", so it sounds like UBERON:0008339 'microvascular endothelium' should be a(n asserted) subclass of UBERON:0004638 'blood vessel endothelium'. I presume that any species with a microvasculature would have larger blood vessels too, so the relationship would always hold true. Any concerns please?
Thanks.
Paola
(Stemming from https://github.com/obophenotype/cell-ontology/pull/1318#pullrequestreview-813862576)

paolaroncaglia:
Hi @cmungall / @dosumis ,
Would you have any objection please if I asserted UBERON:0008339 'microvascular endothelium' as a subclass of UBERON:0004638 'blood vessel endothelium'? Rationale in previous comment. Thanks!

dosumis:
Short answer: No objection.

Longer answer:

'microvascular endothlium' is undefined both textually and logically and has an oddly specific classification given its general name:

![image](https://user-images.githubusercontent.com/112839/146418144-c679d0ad-78d6-4c0c-bc45-cfc34e67f428.png)

I think it makes sense to remove this specific classification and give it a logical definition that puts it under 'blood vessel endothelium.'  For this to work we need a term from microvasculature/microvessel.  Based on  [wikipedia](https://en.wikipedia.org/wiki/Microcirculation), the term microvessels covers arterioles, metarterioles, capillaries, and venules.  Not totally sure of the name.  Maybe  blood microvessel ? (we need to avoid confusion with Uberon:"microcirculatory vessel" which includes lymph)

You can then define 'microvascular endothelium' textually and logically with reference to this grouping:

EquivalentTo 'endothelium' and part_of some blood microvessel
=> inferred classification as

paolaroncaglia:
Note to self:
- [ ] change the equivalence for 'capillary' to point to blood microvessel instead of blood vessel, then close.

PR #2456, opened 2022-05-05
Add UBERON:8410081 'blood microvessel' and related edits

paolaroncaglia:
Addresses #2186


ubyndr:
I think you can ignore diff report and subset report for now, it's been 15 min and docker hasn't started within the action, after 10-12 min if it hasn't started, it will not start and stay idle. Currently working on that issue.

paolaroncaglia:
@dosumis also please note that the diffs and report checks may take forever, but the QC passed fine.

dosumis (approved):
Looks perfect.  Thanks.

PR #2463, opened 2022-05-06
Fovt depreciate issue2333

megbalk:


megbalk:
I've tried removing and adding the database_cross_reference, and still get an error. I'm not sure what to do.

paolaroncaglia:
Uberon editors, could one of you please help Meghan? She's trying to obsolete 3 terms but the diffs check choked on the first one and failed:

:: ERROR: definition-lacks-dbxref: UBERON:7500035
[43](https://github.com/obophenotype/uberon/runs/6329767649?check_suite_focus=true#step:10:43)
  :: ERROR: definition-lacks-dbxref: UBERON:7500039
[44](https://github.com/obophenotype/uberon/runs/6329767649?check_suite_focus=true#step:10:44)
  :: ERROR: logic-axiom-references-obsolete: UBERON:7500039 has relationship to https://orcid.org/0000-0003-2699-3066
[45](https://github.com/obophenotype/uberon/runs/6329767649?check_suite_focus=true#step:10:45)
  :: ERROR: logic-axiom-references-obsolete: UBERON:7500035 has relationship to https://orcid.org/0000-0003-2699-3066
[46](https://github.com/obophenotype/uberon/runs/6329767649?check_suite_focus=true#step:10:46)

I've loaded her branch in Protege and looked at UBERON:7500035, and it looks fine. Is this down to the issue of how to declare an editor's ORCID? Or has a new check been implemented that doesn't want _any_ annotation on obsolete terms? Thanks!

shawntanzk:
assigning @megbalk so PRs don't go unassigned.
see that @anitacaron has replied:

Screenshot 2022-05-10 at 08 39 16

Wonder if it's worth doing a mass change of all dce -> dc terms / term:editor -> dc:contributor?

anitacaron:
In this case, I think the problem is that, in OBO, the contributor is declared as a logical axiom, so the check is to not have any logical axioms in obsolete terms.

shawntanzk:
>  in OBO, the contributor is declared as a logical axiom

huh, did not know that, learn something new everyday :) anyway, just checked, currently the orcids are dbxref on defs though. I see that other obsolete terms have xrefs on them, just not orcids, I'm guessing the pipeline changes orcids to contributor or something somewhere along the line then? If that's the case should we be looking into getting it to not do that on obsolete terms? (or we could strip them out, though I think it might be useful to keep - thoughts?)

megbalk:
@shawntanzk so should I remove the ORCIDS from database_cross_reference? The issue there is that I also get an error saying there is no database_cross_reference for the definition.

shawntanzk:
@megbalk honestly no idea here, not sure what uberon does that is triggering this, maybe @matentzn can help (he might be on leave though, so might take awhile)

matentzn:
All fine again.

Issue #2424, opened 2022-04-18
uberon polysemy issues


vasvir:
Hi,

Trying to use UBERON (uberon-simple.obo from https://github.com/obophenotype/uberon/releases/download/v2022-04-05/uberon-simple.obo) for text mining when I noticed the following polysemy issues.

The following terms are assigned in more than one UBERON Ids using only the term name and the EXACT synonym tags.

**Uberon terms**

      | Gasserian ganglion |   2 | UBERON:0001675, UBERON:3011045 |
      | nV                 |   2 | UBERON:0002633, UBERON:2005093 |
      | temporal pole      |   2 | UBERON:0002576, UBERON:0006479 |
      | trochlea of femur  |   2 | UBERON:4200096, UBERON:7500077 |

**Bug description**

How should I proceed resolving these polysemy issues? Should I try to create individual Issue/PR for each term?

Apologies if this is not the correct place to raise such an issue. I looked for a mailing list or a forum but I didn't find any. Any pointers to right direction are much appreciated.

Thanks in advance.

   Vassilis

matentzn:
Thank you @vasvir!

How comfortable are you editing text files? 

If you make a pull request in this file:

https://github.com/obophenotype/uberon/blob/master/src/ontology/uberon-edit.obo

shawntanzk:
will assign ticket to you @vasvir (please let us know if you need help etc.) as I'm trying to make sure tickets are assigned in uberon :) thanks

vasvir:
Hi @matentzn and everybody else. I would like your advice if possible.

I had created (retrofitted actually) a perl script that checks an OBO ontology for polysemy issues. Here is how the output likes:

```
$ src/scripts/obo_check_polysemy.pl src/ontology/uberon-edit.obo
Set synonym_level: EXACT case sensitivity: 0 for file: src/ontology/uberon-edit.obo
2 ids for name: 'nv' ids: UBERON:0002633, UBERON:2005093, 
2 ids for name: 'phy' ids: UBERON:2000438, UBERON:8440028, 
2 ids for name: 'prelimbic area' ids: UBERON:0013560, UBERON:8440032, 
Problems found: 3

$ src/scripts//obo_check_polysemy.pl -h
Usage: /home/bill/workspace/processing_scripts/obo_check_polysemy.pl [-l|--synonym-level {EXACT|NARROW|RELATED|BROAD}] [-s|case-sensitive] ontology-file.obo

```
[...]

PR #2464, opened 2022-05-07
Polysemy: gasserian ganglion. Delete offending UBERON:3011045

vasvir:
This PR tries to resolve the 1 of 4 of issues mentioned in #2424

"gasserian ganglion" (UBERON:3011045) is an exact synonym to "trigeminal ganglion" (UBERON:0001675) due to WIKIPEDIA. This creates unecessary polysemy.

The proposed solution is to eliminate UBERON:3011045 altogether. There is a comment there that reads
> property_value: provenance_notes "This class was sourced from an external ontology (amphibian_anatomy). Its definitions, naming conventions and relationships may need to be checked for compatibility with uberon" xsd:string {source="http://purl.obolibrary.org/obo/aao.owl"}

I am not sure I follow proper procedure here so if there is any feedback it much appreciated.

As requested by @matentzn I am tagging him.


vasvir:
@matentzn explained the proper procedure to me.

Please ignore.

Closing...

PR #2465, opened 2022-05-10
Polysemy: gasserian ganglion. Obsolete offending UBERON:3011045

vasvir:
This PR tries to resolve the 1 of 4 of issues mentioned in obophenotype#2424

"gasserian ganglion" (UBERON:3011045) is an exact synonym to "trigeminal ganglion" (UBERON:0001675) due to Wikipedia. This creates unecessary polysemy.

The proposed solution is to eliminate UBERON:3011045 altogether. There is a comment there that reads
> property_value: provenance_notes "This class was sourced from an external ontology (amphibian_anatomy). Its definitions, naming conventions and relationships may need to be checked for compatibility with uberon" xsd:string {source="http://purl.obolibrary.org/obo/aao.owl"}


src/ontology/uberon-edit.obo
```diff
@@ -214663,10 +214663,13 @@ property_value: provenance_notes "This class was sourced from an external ontolo
 
 [Term]
 id: UBERON:3011045
-name: gasserian ganglion
+name: obsolete gasserian ganglion
 synonym: "gasserian ganglia" RELATED [UBERON:3011045]
 is_a: UBERON:0000045 ! ganglion
```
matentzn, line 214668: There cannot be is_a (or subClassOf), see https://oboacademy.github.io/obook/howto/obsolete-term/

src/ontology/uberon-edit.obo
```diff
@@ -214663,10 +214663,13 @@ property_value: provenance_notes "This class was sourced from an external ontolo
...
 property_value: provenance_notes "This class was sourced from an external ontology (amphibian_anatomy). Its definitions, naming conventions and relationships may need to be checked for compatibility with uberon" xsd:string {source="http://purl.obolibrary.org/obo/aao.owl"}
```
matentzn, line 214669: delete this `provenance_notes`

src/ontology/uberon-edit.obo
```diff
@@ -214663,10 +214663,13 @@ property_value: provenance_notes "This class was sourced from an external ontolo
 
 [Term]
 id: UBERON:3011045
-name: gasserian ganglion
+name: obsolete gasserian ganglion
 synonym: "gasserian ganglia" RELATED [UBERON:3011045]
```
matentzn, line 214667: move the synonym to the other term (the one that is the replacement for this one)

src/ontology/uberon-edit.obo
```diff
@@ -214663,10 +214663,13 @@ property_value: provenance_notes "This class was sourced from an external ontolo
...
 is_a: UBERON:0000045 ! ganglion
 property_value: provenance_notes "This class was sourced from an external ontology (amphibian_anatomy). Its definitions, naming conventions and relationships may need to be checked for compatibility with uberon" xsd:string {source="http://purl.obolibrary.org/obo/aao.owl"}
+comment: Obsoleted as gasserian ganglion is an exact synonym to trigeminal ganglion per Wikipedia
+is_obsolete: true
+consider: UBERON:0001675
```
matentzn, line 214672: Didnt you say they are exact replacements? You should use `replaced_by` then.

src/ontology/uberon-edit.obo
```diff
@@ -214663,10 +214663,13 @@ property_value: provenance_notes "This class was sourced from an external ontolo
 
 [Term]
 id: UBERON:3011045
-name: gasserian ganglion
+name: obsolete gasserian ganglion
 synonym: "gasserian ganglia" RELATED [UBERON:3011045]
 is_a: UBERON:0000045 ! ganglion
 property_value: provenance_notes "This class was sourced from an external ontology (amphibian_anatomy). Its definitions, naming conventions and relationships may need to be checked for compatibility with uberon" xsd:string {source="http://purl.obolibrary.org/obo/aao.owl"}
+comment: Obsoleted as gasserian ganglion is an exact synonym to trigeminal ganglion per Wikipedia
```
matentzn, line 214670: Wikipedia is not a great source for disambiguation terms - Such decisions should be made by an anatomist! Did you get any expert opinion on this?

matentzn (changes requested):

src/ontology/uberon-edit.obo
vasvir: Aah item 8 missed that. Will do thanks for catching this

src/ontology/uberon-edit.obo
vasvir: I didn't add the provenance line. It has comment greases the way for the obsoletion and I kept it. I also thought I should go for the minimum changes. I will do it though no problem.

src/ontology/uberon-edit.obo
vasvir: ok if you say so. Definitely out of my depth. Will do

src/ontology/uberon-edit.obo
vasvir: it is synoyn: EXACT. I will use replaced_by instead of consider

src/ontology/uberon-edit.obo
vasvir: I consulted a biologist (not anatomist) and google. My understanding is what I propose with this PR. How else could this be resolved? The other term (UBERON:0001675) looks older, more developed and correct and without scary provenance notes.

The third alternative is to not do anything but this leaves the polysemy problem untouched.

I will check again with the biologist, there is nothing else I can do. Sorry...

src/ontology/uberon-edit.obo
vasvir: Ah. It is already there as

synonym: "Gasserian ganglia" RELATED PLURAL [Wikipedia:Trigeminal_ganglion]

I have checked for it but I must have missed the case insensitive flag.

So nothing to do here.

matentzn: No worries, just delete it from here then!

src/ontology/uberon-edit.obo
vasvir: ok we double checked with the biologist. Here is another link that confirms it: The semilunar sensory ganglion (also known as the trigeminal ganglion or Gasserian ganglion) is a...

So I am pushing my changed PR.

Thanks a lot for reviewing and the hand holding in general.

src/ontology/uberon-edit.obo
matentzn: Looks good :)

src/ontology/uberon-edit.obo
matentzn: Looks good to me!

matentzn:
Nicely done @vasvir 

@shawntanzk will when he gets back get some of our anatomy experts to weigh in on the obsoletion here. Regardless, this is a good first PR, hopefully more to come :)

matentzn (approved):

matentzn:
The biology: is this indeed a correct merge (are the two terms the same or not)

shawntanzk (approved):
Sensible obsoletion, thanks for pointing this out :)

Issue #2428, opened 2022-04-19
Developmental bridge files causing unsats


shawntanzk:
Unsats caused by:
[occurrent](http://purl.obolibrary.org/obo/BFO_0000003) DisjointWith [part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [continuant](http://purl.obolibrary.org/obo/BFO_0000002)

Follows the following pattern: 
[FBdv_00005336](http://purl.obolibrary.org/obo/FBdv_00005336) EquivalentTo [larval stage](http://purl.obolibrary.org/obo/UBERON_0000069) and [instar larval stage](http://purl.obolibrary.org/obo/UBERON_0004730) and ([part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [Drosophila melanogaster](http://purl.obolibrary.org/obo/NCBITaxon_7227))
[HsapDv_0000006](http://purl.obolibrary.org/obo/HsapDv_0000006) EquivalentTo [blastula stage](http://purl.obolibrary.org/obo/UBERON_0000108) and ([part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [Homo sapiens](http://purl.obolibrary.org/obo/NCBITaxon_9606))
[ZFS_0000044](http://purl.obolibrary.org/obo/ZFS_0000044) EquivalentTo [post-juvenile adult stage](http://purl.obolibrary.org/obo/UBERON_0000113) and ([part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [Danio](http://purl.obolibrary.org/obo/NCBITaxon_7954))
[XAO_1000001](http://purl.obolibrary.org/obo/XAO_1000001) EquivalentTo [zygote stage](http://purl.obolibrary.org/obo/UBERON_0000106) and ([part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [Xenopus ](http://purl.obolibrary.org/obo/NCBITaxon_8353))
[FMA_24120](http://purl.obolibrary.org/obo/FMA_24120) EquivalentTo [life cycle stage](http://purl.obolibrary.org/obo/UBERON_0000105) and ([part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [Homo sapiens](http://purl.obolibrary.org/obo/NCBITaxon_9606))
[...]

PR #2466, opened 2022-05-11
bridges: Change relation to life cycle terms to BFO:0000066.

gouttegd:
Update the bridge generation scripts so that, in species-specific bridges, terms that are cross-referenced with Uberon’s `life cycle` (UBERON:0000104 or its subclasses) are defined in the generated bridges as equivalent to the Uberon term `occuring in` (BFO:0000066) the specific taxon, rather than the Uberon term `being part of` (BF:0000050) the specific taxon.

This is dependent on having a list of all terms that are cross-referenced to the Uberon's `life cycle` hierarchy. Generating such a list is outside the scope of this patch.

Assuming the list of cross-referenced terms is available in a file named `lsxrefs.txt`, generate the bridges with

```sh
./run.sh make bridge/bridges BRIDGE_LSXREFS='-l ../lsxrefs.txt'
```

Here is an example of the effects of this patch on the `uberon-bridge-to-fbdv.obo` bridge file (FBdv:00000000 being cross-referenced to Uberon’s UBERON:0000104):

```diff
 [Term]
 id: FBdv:00000000 ! 
 property_value: IAO:0000589 "life cycle (FBdv)" xsd:string
 intersection_of: UBERON:0000104 ! life cycle
-intersection_of: part_of NCBITaxon:7227
+intersection_of: BFO:0000066 NCBITaxon:7227
 
 [Term]
 id: FBdv:00007012 ! 
```

### DIFF
--- a/src/ontology/make-bridge-ontologies-from-xrefs.pl
+++ b/src/ontology/make-bridge-ontologies-from-xrefs.pl
@@ -84,7 +84,7 @@ while (<>) {
                 elsif ($t eq 'gd') {
                     print $fh "intersection_of: $id ! $n\n";
                     if (exists($lsxrefs{$x})) {
-                        $rel = "BFO:0000066";
+                        $rel = "occurs_in";
                     }
                     print $fh "intersection_of: $rel $filler\n";
                 }
@@ -127,6 +127,9 @@ foreach my $s (keys %fhmap) {
     print $fh "[Typedef]\n";
     print $fh "id: develops_from\n";
     print $fh "xref: RO:0002225\n\n";
+    print $fh "[Typedef]\n";
+    print $fh "id: occurs_in\n";
+    print $fh "xref: BFO:0000066\n\n";
 
     open(F,"ftr.obo");
     while(<F>) {
--- a/src/ontology/make-bridge-ontologies-from-xrefs.pl
+++ b/src/ontology/make-bridge-ontologies-from-xrefs.pl
@@ -9,6 +9,7 @@ my $n;
 my $n_xrefs = 0;
 my $stype;
 my %nh = ();
+my %lsxrefs = ();
 
 my @fns = ();
 my $base = 'uberon';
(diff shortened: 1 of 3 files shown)


src/ontology/make-bridge-ontologies-from-xrefs.pl
```diff
@@ -74,6 +83,9 @@
                 }
                 elsif ($t eq 'gd') {
                     print $fh "intersection_of: $id ! $n\n";
+                    if (exists($lsxrefs{$x})) {
+                        $rel = "BFO:0000066";
```
matentzn, line 87: Just for consistency, can we declare occurs_in same way as we do part of at the end of the file (https://github.com/obophenotype/uberon/blob/master/src/scripts/make-bridge-ontologies-from-xrefs.pl#L113)

src/ontology/make-bridge-ontologies-from-xrefs.pl
```diff
@@ -18,6 +19,14 @@
     if ($opt eq '-b' || $opt eq '--base') {
         $base = shift @ARGV;
     }
+    elsif ($opt eq '-l' || $opt eq '--life-cycle-xrefs') {
```
matentzn, line 22: I cant read this :P this is opening a file, right? Not a bunch of CL arguments?

matentzn (commented):
Excellent THANK YOU!

src/ontology/make-bridge-ontologies-from-xrefs.pl
gouttegd: Sure.

anitacaron:
Fixes #2428

src/ontology/make-bridge-ontologies-from-xrefs.pl
matentzn: thx

matentzn (approved):

PR #2462, opened 2022-05-06
Revert "depreciating FOVT terms"

megbalk:
Reverts obophenotype/uberon#2457

PR #2469, opened 2022-05-12
Fix main branch (uberon-edit)

matentzn:


shawntanzk:
@matentzn - is it a concern that the error message in the diff-report build is still the same (see https://github.com/obophenotype/uberon/pull/2462#issuecomment-1124696361) - tried this before and got the same thing here and was worried that it would cause issues down the road in building release etc.

matentzn:
Shouldnt be a problem from now on, since the fix is now on master?

PR #2470, opened 2022-05-13
Polysemy: temporal pole. Delete offending synonym from UBERON:0006479

vasvir:
Polysemy: temporal pole. Delete offending synonym from UBERON:0006479 (Brodmann (1909) area 38)

This PR tries to resolve the 2 of 4 of issues mentioned in #2424

"temporal pole" in UBERON:0006479 is listed as an exact synonym of "Brodmann (1909) area 38". However, "temporal pole" is also the canonical for UBERON:0002576. This creates unnecessary polysemy.

Looks like UBERON:0006479 is a more specific term (area 38). Therefore the proposed solution deletes "temporal pole" from the synonyms of "Brodmann (1909) area 38".


shawntanzk:
Yeah the broadmann area stuff is a bit problematic in a multispecies ontology like uberon (many terms are exact synonyms in broadmann area that shouldn't be as broadmann area is specific to humans) - it's a problem I'm facing at the moment too. I agree with your changes.

shawntanzk (approved):

shawntanzk:
An alternative, if we want to be conservative and keep this, is to change it to related synonym.

vasvir:
Disclaimer: I am approaching this from a mechanistic/technical side. I have consulted a biologist but she also is not an expert in brain anatomy. Personally I am mostly weighting a term based on its synonyms...

Disclaimer2: Furthermore I am not familiar with the project and my 'conservartivness' intuition is just starting to grow... So any guidance is much appreciated.

If you feel that I should go with related just let me know. I don't have strong opinions as I am not an expert.

shawntanzk:
@vasvir - lets change it to related synonym instead. I've seen this with a few other similar names for brain regions before. Thanks :)

PR #2474, opened 2022-05-17
Uberon release 17 May 2022

matentzn:


shawntanzk (approved):
don't see any issues is hte -base file at all. 
strange that the diff.md you added in this PR only shows changes in genid stuff (I assume this is not an issue) though I see other differences (eg mesh changes and term additions) - also not sure if the addition of the diff file is intentional

PR #2475, opened 2022-05-18
Update vasculature_class template

anitacaron:


dosumis (approved):
Looks good, but please check that this version is stable enough before merge.

Issue #2482, opened 2022-05-19
coracoid process of calcaneus


megbalk:
**Preferred term label:**
coracoid process of calcaneus

**Synonyms**
processus coracoideus of calcaneus

**Definition (free text, please give PubMed ID)**
At the dorsal part of the base of the calcaneal tuberoisty is a pointed process called the coracoid process.

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
part of some calcaneus tuberosity; zone of bone organ

**Your nano-attribution (ORCID)** 
https://orcid.org/0000-0003-2699-3066

megbalk:
modeling this after the coracoid process of scapula (UBERON:0006633)

PR #2483, opened 2022-05-19
adding coracoid process of calcaneus

megbalk:
from issue #2482

### DIFF
--- a/docs/odk-workflows/RepositoryFileStructure.md
+++ b/docs/odk-workflows/RepositoryFileStructure.md
@@ -54,3 +54,4 @@ These are the components in UBERON
 | -------- | --- |
 | mappings.owl | None |
 | in-subset.owl | None |
+| hra_subset.owl | None |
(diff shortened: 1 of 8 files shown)


megbalk:
@nicolevasilevsky I'm not sure why this is failing...

shawntanzk:
```
FLAG: relation used but not defined
'http://purl.org/dc/elements/1.1/date' -- property_value: http://purl.org/dc/elements/1.1/date 2022-05-19T21:16:42Z xsd:dateTime

FLAG: relation used but not defined
'http://purl.org/dc/elements/1.1/contributor' -- property_value: http://purl.org/dc/elements/1.1/contributor http://orcid.org/0000-0003-2699-3066
```

These should be dc:terms instead of dc:elements

megbalk:
I think I fixed it...

megbalk:
@nicolevasilevsky can you please review this?

nicolevasilevsky (approved):
I just fixed the conflict, can you double check that the diff looks correct

PR #2492, opened 2022-05-24
Auto refresh fbbt mappings

gouttegd:
Update the custom Makefile to automatically download the FBbt SSSOM file from the FBbt repository, and therefore automatically update the FBbt mappings.

Downloading the SSSOM file is made conditional on the `IMP` variable, so that the file is only refreshed when running with `IMP=true`. Updating the mappings is thus considered similar to refreshing the imports, which I believe is sensible.


matentzn (approved):
Yes, this is the way to go I think! Looks great!

Issue #2493, opened 2022-05-25
Update 'genital system': populate with subclasses, add text definition and synonym


bvarner-ebi:
**Uberon term**
UBERON:3011048 'genital system'

**Synonym to be added**
- [x] broad: genitalia
- [x] broad: reproductive system

**Notes**
- [x] Populate with appropriate subclasses.
- [x] Add text definition. Definition from [XAO term](http://purl.obolibrary.org/obo/XAO_0000142) can be appropriated.
- [x] After PR is merged, update [OrganList](https://docs.google.com/spreadsheets/d/1QzuB07M61TqCElRzydWVs4ap5RIIlmg7xIpSLxC6PFU/edit?skip_itp2_check=true#gid=0) to map major.Organ 'genitals' to Uberon 'genital system'
- [x] Ask @ubyndr to rerun mapping coverage analysis (related [Organ and Ontology Discussion](https://capdevelopment.atlassian.net/wiki/spaces/CAP/pages/162889729/2022-05-24+Organ+ontology+discussion)).

Motivation: update term with subclasses to improve coverage with the Cell Annotation Platform (CAP) organ terms, which are being leveraged to build the autosuggest search term feature for the platform.

cmungall:
You perhaps weren't to know this but I think the correct course of action here was to merge genital system. See #2514

bvarner-ebi:
@cmungall, do you mean merge 'genital system' with 'reproductive system'?

FYI @dosumis

balhoff:
Is 'genital system' just amphibian-specific terminology for reproductive system? https://www.ontobee.org/ontology/XAO?iri=http://purl.obolibrary.org/obo/XAO_0000142

dosumis:
I think the key question is whether [reproductive system](https://github.com/obophenotype/uberon/issues/2493) is broader than genitalia (we have terms for internal and external genitalia, but not the grouping term - genital system is the closest we have to that).

![image](https://user-images.githubusercontent.com/112839/172811481-6c1b09ab-9ee2-42d6-8143-48d3e7e8f515.png)

We have terms under reproductive system that I don't think anyone would consider genitalia, e.g. 

cotylephore
[http://purl.obolibrary.org/obo/UBERON_2002114 ](http://purl.obolibrary.org/obo/UBERON_2002114) 
Multi-tissue structure that is a transient outgrowth of tissue on skin brooding fish that acts as an attachment site for embryos. [ TAO : wd ]

[...]

PR #2494, opened 2022-05-25
Address #2493 Update 'genital system'

bvarner-ebi:
Address #2493 Update 'genital system'


shawntanzk:
@bvarner-ebi - if you dont need the diff, just ignore it. If you want it try re-running later (or do it manually yourself on your local machine if you know how to). Basically it's just resource issues I think.

ubyndr (approved):

src/ontology/uberon-edit.obo
```diff
@@ -10024,6 +10024,7 @@ xref: Wikipedia:Gonad
 xref: XAO:0003146
 xref: ZFA:0000413
 is_a: UBERON:0003133 ! reproductive organ
+is_a: UBERON:3011048 ! genital system
```
dosumis, line 10027: I think part_of would be better.

dosumis: Also for most of all other children.  Same scope for refactoring to clean up too given hidden GCIs.

dosumis (changes requested):
See comments

Issue #2165, opened 2021-11-10
Remove all axioms that reference GO:excretion


cmungall:
GO is obsoleting excretion

https://github.com/geneontology/go-ontology/issues/22368

Related issues:

 * Alliance wants a general "excretory system" that spans all organisms: #1450

shawntanzk:
@anitacaron - could you help bring this up in tech call (I've placed it on the board, just wont be around for the next meetings and all)

gouttegd:
`GO:excretion` is referenced by three terms in Uberon:

* `excreta`, defined as `'organism substance' and (output_of some excretion)`
* `excreted gas`, defined as `'bodily gas' and (output_of some excretion)`
* `renal system`, defined as `organ system and (capable_of some excretion)`

The last one is the most problematic in my opinion as `excretion` in GO has a definition that goes beyond what is normally performed by the renal system (for example, it encompasses defecation). Suggestion here is to update the definition of `renal system` to `organ system and (capable_of some 'renal tubular secretion')`, where `renal tubular secretion` is a subclass of `excretion`.

As for `excreta` and `excreted gas`, not sure we can as easily replace `excretion` in their definitions by a more precise subclass. Both terms are (purposefully, I assume) broadly defined to encompass several types of excretions. Given that GO has decided not to obsolete `excretion` in the end, I would be inclined to keep those references as they are.

gouttegd:
Of note, if we implemented my suggestion above (redefining `renal system` to `organ system and (capable_of some 'renal tubular secretion')`), then we probably also should create a higher-level `excretory system` to encompass all excretory stuff (not limited to kidneys and kidney-like), as requested in #1450.

There are at least one instance where `renal system` is currently used as if it referred to a generic “excretory system”: `excretory gland` is currently `part_of some 'renal system'`, even though the text definition of `excretory gland` does not tie it to the renal system at all (on the contrary, it gives “lacrimal gland” as an example of an excretory gland).

Issue #1450, opened 2018-08-07
Alliance expression ribbon needs new term for "excretory system"


doughowe:
<img width="349" alt="screen shot 2018-08-07 at 9 44 51 am" src="https://user-images.githubusercontent.com/7142077/43790127-8def3066-9a26-11e8-932b-4fd956eeb9de.png">

Currently, uberon has the term 'renal system' (UBERON:0001008) as a child of "anatomical structure" (UBERON:0000467).  That term is defined "The renal system in an anatomical system that maintains fluid balance and contributes to electrolyte balance, acid/base balance, and disposal of nitrogenous waste products." (NOTE: there is an extra period at the end of the definition).

Can this term be renamed to "excretory system"?  "renal system" is already a synonym even thought the term itself is "renal system".

If the term is renamed, the definition changes to:
The excretory system in an anatomical system that maintains fluid balance and contributes to electrolyte balance, acid/base balance, and disposal of nitrogenous waste products.

If a "renal system" term is desired, then a new term "excretory system" would be placed between "anatomical system" and "renal system"

cmungall:
Uberon attempts to follow the GO classification that came out of the kidney working group, but it seems to have been vertebrate + arthropod centric, see my comments here:

Linked GO ticket: https://github.com/geneontology/go-ontology/issues/16143

sbello:
The VTO called this urinary system  http://purl.obolibrary.org/obo/VT_1000777 Would this work? Do C. elegans produce urine? Cloaca (i.e. http://purl.obolibrary.org/obo/VT_0015055) lives under the surface structure branch and not under urinary or digestive system branches.

cerivs:
Another possible term name is biofiltration system.

PR #2497, opened 2022-05-25
Add excretory system

gouttegd:
This PR redefines `renal system` as `capable_of some GO:'renal tubular secretion'` (instead of `capable_of some GO:secretion`), as discussed in #2165.

The `secretion` capability is moved to a new term `secretory system`, as requested in #1450. `renal system` is made a subclass of that new `excretory system`.


src/ontology/uberon-edit.obo
```diff
@@ -223782,6 +223781,16 @@ relationship: part_of UBERON:0000992 ! ovary
 property_value: dcterms-date 2022-04-19T11:08:10Z xsd:dateTime
 property_value: seeAlso https://github.com/obophenotype/uberon/issues/476 xsd:anyURI
 
+[Term]
+id: UBERON:8450002
+name: excretory system
+def: "An anatomical system that eliminates waste products that arise as a result of metabolic activity." [GO:0007588]
+intersection_of: UBERON:0000467 ! anatomical system
+intersection_of: capable_of GO:0046903 ! secretion
+relationship: existence_ends_during UBERON:0000066 ! fully formed stage
```
bvarner-ebi, line 223790: @gouttegd, I'm unsure about this axiom. Why does existence of the 'excretory system' end at a fully formed stage? It looks like this was pointed out before in other instances, but never addressed: https://github.com/geneontology/go-ontology/issues/14501

gouttegd: Good question. I am not sure what those `existence_ends_*` axioms are supposed to model…

Most other anatomical systems have a similar axiom (including the old `renal system` that is being redefined in this PR). And on top of that, `anatomical structure` has a `existence_ends_during_or_before some 'death stage'`.

So as a result, we have anatomical systems whose existence is either said to
* “end during fully formed stage” (if there’s an explicit axiom on the anatomical system itself)
* “end during or before death stage” (without an explicit axiom, using the one inherited from `anatomical structure`).

What is the difference between the two supposed to mean? Frankly I don’t know. For now, I applied the “existence ends during fully formed stage” axiom to `excretory system` solely on the ground that the axiom was already applied to `renal system`. I don’t mind removing it and maybe leaving that to a later issue where we would review all uses of `existence_ends_*` axioms.

bvarner-ebi: > I don’t mind removing it and maybe leaving that to a later issue where we would review all uses of existence_ends_* axioms.

Sounds reasonable.

bvarner-ebi (approved):

PR #2501, opened 2022-05-26
Reformat the ID ranges file to proper Manchester syntax.

gouttegd:
The `uberon-idranges.owl` file was incorrectly formatted, preventing its use by Protégé. Reformat it so that ID range policies can be automatically applied when the next version of Protégé will be available.

No changes to the actual content of the file.

### DIFF
--- a/src/ontology/uberon-idranges.owl
+++ b/src/ontology/uberon-idranges.owl
@@ -1,141 +1,209 @@
-Prefix: idrange:=<http://purl.obolibrary.org/obo/uberon/idrange/>
-Prefix: allocatedto:=<http://purl.obolibrary.org/obo/IAO_0000597>
-Prefix: iddigits:=<http://purl.obolibrary.org/obo/IAO_0000596>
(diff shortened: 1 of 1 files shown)


shawntanzk (approved):
Looks good, thanks for changing! I'm not 100% sure on the details (eg is equivalent what protege reads etc) but I assume you know :)

gouttegd:
In case you’d want to check by yourself that the change is good: try converting the `uberon-idranges.owl` file with ROBOT:

```
robot convert -i uberon-idranges.owl -f ofn -o tmp.ofn
```

You’ll get an error with the current version of the file. Try again after applying this PR: all good. ;)

The incorrect format was never noticed, probably because for now, this file is actually never read by anything (apart by human beings, when they need to update their ID range settings in Protégé). But this will change when the next version of Protégé is out, so now we‘ll be ready for it. :)

gouttegd:
P.S.: Assuming that I know what I‘m doing is a dangerous thing… 😆

Issue #2514, opened 2022-06-02
gonad is inappropriately placed in the renal system


cmungall:
From @balhoff in GO:

![image](https://user-images.githubusercontent.com/50745/171673900-f135c23a-0926-42a4-bfff-9f0e2bc5d59e.png)

the fact this wasn't caught in the uberon release system is a problem in itself

Another tip: be very very conservative with uberon IDs in the 2x, 3x and 4x range. This term “genital system” is from AAO. It should probably be merged. In the interim, be ultra-careful when linking classes to these terms

```yaml
id: UBERON:3011048
name: genital system
def: "Anatomical system that has as its parts the organs concerned with reproduction, especially the external genital organs." [https://www.merriam-webster.com/dictionary/genitalia, XAO:0000142]
synonym: "genitalia" BROAD []
synonym: "reproductive system" BROAD []
is_a: UBERON:0000467 ! anatomical system
relationship: part_of UBERON:0004122 ! genitourinary system
property_value: provenance_notes "This class was sourced from an external ontology (amphibian_anatomy). Its definitions, naming conventions and relationships may need to be checked for compatibility with uberon" xsd:string {source="http://purl.obolibrary.org/obo/aao.owl"}
```

dosumis:
Why do we even have this term? Super confusing.  Anyone using it in annotation will expect the child terms Bradley added.

balhoff:
Fixing this issue will contribute to turning the Ubergraph source coherency check green: https://github.com/INCATools/ubergraph#coherency-checks

nlharris:
Can this be closed?

dosumis:
@bvarner-ebi can you confirm fixed?

bvarner-ebi:
`'gonad' SubClassOf 'part of' some 'genital system'` was removed.
`'genital system'` was obsoleted.
@balhoff previously confirmed Ubergraph coherency checks passed.

Closing this ticket.

PR #2521, opened 2022-06-09
Associated with 2493 and 2514

bvarner-ebi:
See #2493 and #2514
immediate fix to address the ubergraph update issue


balhoff:
@shawntanzk I think left and right is reversed for the unreasoned diff action.

balhoff (approved):

PR #2522, opened 2022-06-10
swap left and right in unreasoned diff

shawntanzk:
swap left and right in unreasoned diff
addresses https://github.com/obophenotype/uberon/pull/2521#issuecomment-1151343168


matentzn (approved):
Looks good!

ubyndr (approved):
My bad, apologies.

PR #2523, opened 2022-06-10
Addresses #2493 merge: genital system into reproductive system

bvarner-ebi:
Addresses #2493 merge: genital system into reproductive system


bvarner-ebi:
Unsure why the QC check failed. 

Last few rows of QC report...
```
Replaced: 26588
[50](https://github.com/obophenotype/uberon/runs/6836964400?check_suite_focus=true#step:4:51)
Cloning into 'developmental-stage-ontologies'...
[51](https://github.com/obophenotype/uberon/runs/6836964400?check_suite_focus=true#step:4:52)
make: *** [uberon.Makefile:752: reports/uberon-edit-xp-check] Error 1
[52](https://github.com/obophenotype/uberon/runs/6836964400?check_suite_focus=true#step:4:53)
Error: Process completed with exit code 2.
```

shawntanzk:
```
FLAG: relation used but not defined
'http://purl.org/dc/elements/1.1/date' -- property_value: http://purl.org/dc/elements/1.1/date 2022-06-10T18:22:34Z xsd:dateTime
```
you used dce instead of dc

bvarner-ebi:
> ```
> FLAG: relation used but not defined
> 'http://purl.org/dc/elements/1.1/date' -- property_value: http://purl.org/dc/elements/1.1/date 2022-06-10T18:22:34Z xsd:dateTime
> ```

> you used dce instead of dc

Strange, I don't think I intentionally added / changed this is the first commit.  Regardless, I updated this.

bvarner-ebi:
@cmungall, @dosumis, if this PR looks OK, can you approve so I can move forward with the merge?

src/ontology/uberon-edit.obo
```diff
@@ -214638,13 +214640,14 @@ replaced_by: UBERON:0001675
 
 [Term]
 id: UBERON:3011048
-name: genital system
-def: "Anatomical system that has as its parts the organs concerned with reproduction, especially the external genital organs." [https://www.merriam-webster.com/dictionary/genitalia, XAO:0000142]
+name: obsolete genital system
+def: "OBSOLETE Anatomical system that has as its parts the organs concerned with reproduction, especially the external genital organs." [https://www.merriam-webster.com/dictionary/genitalia, XAO:0000142]
```
nicolevasilevsky, line 214644: in Mondo, we prepend the definition with OBSOLETE. text... (we add a period after the word obsolete). I spot checked a few examples in the file and it doesn't look like Uberon has prepended definitions with OBSOLETE. But I think this is fine.

src/ontology/uberon-edit.obo
```diff
@@ -214638,13 +214640,14 @@ replaced_by: UBERON:0001675
...
+comment: https://github.com/obophenotype/uberon/issues/2493
```
nicolevasilevsky, line 214645: in Mondo, we use seeAlso for the URL. 

Is there any documentation anywhere about the obsoletion workflow for Uberon?

nicolevasilevsky (approved):

nicolevasilevsky:
this looks fine to me. Could we create some documentation about the obsoletion workflow for Uberon? This looks a bit different than how we do it in Mondo but I'm not sure what is preferred for Uberon.

We could base the documentation on our Mondo documentation: https://mondo.readthedocs.io/en/latest/editors-guide/merging-and-obsoleting/

src/ontology/uberon-edit.obo
cmungall: Let's make the workflows 100% identical. Are there differences other than prepending the definition string? Let's just make a universal decision across our ontologies yes/no to prepending definition strings.

(I don't think the comment comment is related to the obsoletion workflow per se, but in general Uberon should use largely the same APs as Mondo, and the same way of relating a term to an issue)

balhoff: In GO we use [IAO:0000233](http://purl.obolibrary.org/obo/IAO_0000233)

nicolevasilevsky: should we use [IAO:0000233](http://purl.obolibrary.org/obo/IAO_0000233) term tracker item in Mondo too? I can write up some documentation on obsoletion workflows. Should I add it to OBO Academy? We want that to be the definitive source, right?

bvarner-ebi:
Thanks very much for reviewing, @nicolevasilevsky! 
It has been a while since I made these edits, but I think I followed the steps outlined here: https://oboacademy.github.io/obook/howto/obsolete-term/#obsoletion-process-manual

The screenshot in that documentation shows "OBSOLETE" (with no period) prepending the definition.

Screenshot 2022-08-10 at 09 37 10

Of note, the steps in this documentation are all labelled step 1 ("1.").

cmungall:
let's merge this and take the obsoletion workflow discussion elsewhere, thanks!

Issue #2526, opened 2022-06-14
Switching mba/dmba to uberon bridges 


shawntanzk:
Currently on building release there are 3 unsats from xrefs: 
```
UNSAT: DMBA:15612 DMBA:15612
UNSAT: MBA:30 MBA:30
UNSAT: MBA:689 MBA:689
NUMBER_OF_UNSATISFIABLE_CLASSES: 3

```
These are from our efforts to add new terms as we are re-doing mba and dmba bridge to uberon to be more accurate: https://github.com/obophenotype/uberon/pull/2484 

I'm not 100% sure what is causing the unsats but I guess this is as good as anytime to switch over the bridge files 

We have generated new bridges that are manually curated here: 
uberon-bridge-to-dmba.owl: https://github.com/obophenotype/ABA_Uberon/blob/new_bridge/src/ontology/new-bridges/new-uberon-bridge-to-dmba.owl
uberon-bridge-to-mba.owl: https://github.com/obophenotype/ABA_Uberon/blob/new_bridge/src/ontology/new-bridges/new-uberon-bridge-to-mba.owl

I would like to stop these two files from generating through xrefs and instead download from that repo and use that - I'm however not super sure how to handle this 

Was wondering if @matentzn and/or @gouttegd could help me out with this :) 

From there I'm hoping with the new fixes, unsats will go away, if not we will investigate what is going on.

Thanks!

shawntanzk:
ps if we want a release out before we can get to this, I'm also happy to remove those xrefs - given the end goal will be to use the new bridge files anyway

gouttegd:
Probably won't have a look at that until next week, so if a release needs to be done maybe you could indeed just remove the offending xrefs for now.

Longer term, I'd really like to avoid accumulating different ad-hoc solutions to deal with/generate bridge files. There's already the FBbt bridge files that are treated differently from all the others (since I used them to start testing SSSOM). I'd like to have a generalized, ODK-managed solution for all bridges -- I was planning to work on that as soon as we figure out 1) what property to use for cross-species mappings and 2) how do we "translate" a mapping predicate in a SSSOM mapping set to a OWL axiom that can be put in a bridge file.

shawntanzk:
I like the idea of a generalised ODK-managed solution for bridges with good documentation of how to handle it etc. Happy to remove the 3 offending xrefs for now for release purposes.

gouttegd:
[...]

PR #2529, opened 2022-06-14
Overwriting mba/dmba bridge to custom component

shawntanzk:
Related to https://github.com/obophenotype/uberon/issues/2526
Currently importing as just a component which is 100% wrong as it leaves dangling classes etc. Should probably handle it more like a bridge file, but im not sure how to do that


shawntanzk:
Changed it so that it basically just downloads into the bridge - hence overwriting the old mba/dmba bridge files that were generated through xrefs

shawntanzk:
Fixes https://github.com/obophenotype/uberon/issues/2526

gouttegd (changes requested):
Few things:

* Unless I‘ve missed something, the new rules to download the custom MBA/DMBA bridges are completely disconnected from any other workflows and therefore need to be called explicitly. Normally the bridge files are generated by invoking the `$(BRIDGEDIR)/bridges` target, it would be nice if downloading the MBA/DMBA bridges could be made part of that target.
* We normally provide the bridges in both OWL and OBO format. In case there are consumers out there who were relying on the OBO-format MBA/DMBA bridges, I think we should continue providing them (by adding a OWL-to-OBO conversion step after downloading the bridges in OWL format) — either that or add a breaking change warning in the release notes.

Most importantly: This is fine as a temporary measure _or_ for a one-off case, but I really don’t want to see that approach being extended organically to other bridges. If other automatically generated bridges are going to need to be replaced by custom-generated, externally maintained bridges, I strongly believe we need to stop and think about a well-defined solution, ideally at the ODK level.

shawntanzk:
> Normally the bridge files are generated by invoking the $(BRIDGEDIR)/bridges target

Do you mean `$(TMPDIR)/bridges`? I can't find `$(BRIDGEDIR)/bridges` - this looks like it builds bridges then push it through for QC 
but yes, 100% agree it would be nice to have some automated functionality

>  adding a OWL-to-OBO conversion step after downloading the bridges in OWL format

Happy to do this :) thanks!

shawntanzk:
> If other automatically generated bridges are going to need to be replaced by custom-generated, externally maintained bridges, I strongly believe we need to stop and think about a well-defined solution, ideally at the ODK level.

btws, I'm hoping this is just a neuro thing, cause we always think we are special, but I definitely cannot 100% for sure say that this wont be an issue otherwise. An alternative would be that I can eventually move the template into uberon instead and get it maintained here so its at least a bit cleaner, but I don't think I can escape from using a template

gouttegd:
> Do you mean $(TMPDIR)/bridges? I can't find $(BRIDGEDIR)/bridges

Ah yes, I have missed the refactoring that Nico did on that part of the pipeline in #2473, my bad. Should have refreshed my Uberon copy on my laptop before going on vacation… :P 

> I don't think I can escape from using a template

To be clear, I have no objection to that. If someone has complex needs that cannot be satisfied with simple bridges automatically generated from xrefs (or later from SSSOM mapping sets), then it’s perfectly fine to use another method, and it should be supported.

My point is that the ODK should offer basic support for that, so that those who needs “complex“ bridges do not have to come up with their own way of integrating them into a repo.

I‘m hoping that the MBA/DMBA case can serve as a “test” to explore exactly what we need and what we can do, and then we can move the code into the ODK.

src/ontology/uberon.Makefile
```diff
@@ -632,6 +632,21 @@ $(TMPDIR)/uberon-edit-plus-tax-equivs.owl: $(OWLSRC) $(TMPDIR)/external-disjoint
...
+
+UBERON_BRIDGE_MBA = "https://raw.githubusercontent.com/obophenotype/ABA_Uberon/new_bridge/src/ontology/new-bridges/new-uberon-bridge-to-mba.owl"
+$(BRIDGEDIR)/uberon-bridge-to-mba.owl: $(SRC)
+	$(ROBOT) annotate -I $(UBERON_BRIDGE_MBA) --ontology-iri $(ONTBASE)/$@ -o $@
+$(BRIDGEDIR)/uberon-bridge-to-mba.obo: $(BRIDGEDIR)/uberon-bridge-to-mba.owl
+	$(ROBOT) convert --input $(BRIDGEDIR)/uberon-bridge-to-mba.owl --output $(BRIDGEDIR)/uberon-bridge-to-mba.obo
+
+UBERON_BRIDGE_DMBA = "https://raw.githubusercontent.com/obophenotype/ABA_Uberon/new_bridge/src/ontology/new-bridges/new-uberon-bridge-to-dmba.owl"
+$(BRIDGEDIR)/uberon-bridge-to-dmba.owl: $(SRC)
+	$(ROBOT) annotate -I $(UBERON_BRIDGE_DMBA) --ontology-iri $(ONTBASE)/$@ -o $@
```
matentzn, line 646: Please add `if [ $(BRI) = true ]; then` condition as we do for the main `bridges` goal

src/ontology/uberon.Makefile
```diff
@@ -632,6 +632,21 @@ $(TMPDIR)/uberon-edit-plus-tax-equivs.owl: $(OWLSRC) $(TMPDIR)/external-disjoint
 	$(OWLTOOLS_CAT_DYNAMIC) $< $(TMPDIR)/external-disjoints.owl `ls $(BRIDGEDIR)/uberon-bridge-to-*.owl | grep -v emap.owl` --merge-support-ontologies -o -f ofn $@
 .PRECIOUS: $(TMPDIR)/uberon-edit-plus-tax-equivs.owl
 
+# uberon bridges to mba and dmba are now manually curated and generated in https://github.com/obophenotype/ABA_Uberon/tree/new_bridge
+# Note: the bridges are generated in new_bridge branch - this might change in the future, if it breaks here, please check ABA_Uberon repo to make sure that the new bridges are appropriately linked
+
+UBERON_BRIDGE_MBA = "https://raw.githubusercontent.com/obophenotype/ABA_Uberon/new_bridge/src/ontology/new-bridges/new-uberon-bridge-to-mba.owl"
```
matentzn, line 638: Risky to use a link to a branch rather than `main`..

src/ontology/uberon.Makefile
matentzn: (same for all the bridge goals).

matentzn: note the `; fi` in the end of each row!

src/ontology/uberon.Makefile
```diff
@@ -632,6 +632,21 @@ $(TMPDIR)/uberon-edit-plus-tax-equivs.owl: $(OWLSRC) $(TMPDIR)/external-disjoint
...
+$(BRIDGEDIR)/uberon-bridge-to-mba.owl: $(SRC)
+	$(ROBOT) annotate -I $(UBERON_BRIDGE_MBA) --ontology-iri $(ONTBASE)/$@ -o $@
+$(BRIDGEDIR)/uberon-bridge-to-mba.obo: $(BRIDGEDIR)/uberon-bridge-to-mba.owl
+	$(ROBOT) convert --input $(BRIDGEDIR)/uberon-bridge-to-mba.owl --output $(BRIDGEDIR)/uberon-bridge-to-mba.obo
+
+UBERON_BRIDGE_DMBA = "https://raw.githubusercontent.com/obophenotype/ABA_Uberon/new_bridge/src/ontology/new-bridges/new-uberon-bridge-to-dmba.owl"
+$(BRIDGEDIR)/uberon-bridge-to-dmba.owl: $(SRC)
+	$(ROBOT) annotate -I $(UBERON_BRIDGE_DMBA) --ontology-iri $(ONTBASE)/$@ -o $@
+$(BRIDGEDIR)/uberon-bridge-to-dmba.obo: $(BRIDGEDIR)/uberon-bridge-to-dmba.owl
+	$(ROBOT) convert --input $(BRIDGEDIR)/uberon-bridge-to-dmba.owl --output $(BRIDGEDIR)/uberon-bridge-to-dmba.obo
```
matentzn, line 648: General tip, insread of spelling out the --output parameter, use the much shorter `-o $@`. No need now, just for you to know.

src/ontology/uberon-edit.obo
```diff
@@ -206,7 +205,6 @@ treat-xrefs-as-reverse-genus-differentia: HAO part_of NCBITaxon:7399
 treat-xrefs-as-reverse-genus-differentia: HBA part_of NCBITaxon:9606
 treat-xrefs-as-reverse-genus-differentia: HsapDv part_of NCBITaxon:9606
 treat-xrefs-as-reverse-genus-differentia: MA part_of NCBITaxon:10090
-treat-xrefs-as-reverse-genus-differentia: MBA part_of NCBITaxon:10090
```
matentzn, line 209: I understand you think you know what this does, but I just want to note I dont.

matentzn (changes requested):

src/ontology/uberon.Makefile
shawntanzk: yes, I really disliked this when writing, but we don't have a choice atm, main is used as a comparison branch etc. 
For now if you don't mind, I'd rather leave it this way, but I think I'd want ot move the whole template building to bridge (so the tsv file and all) to uberon.

shawntanzk:
btws if this is a blocker to release, happy to remove these xrefs to get it through, and we can work on this on a slower timescale
UNSAT: DMBA:15612 DMBA:15612
UNSAT: MBA:30 MBA:30
UNSAT: MBA:689 MBA:689

src/ontology/uberon-edit.obo
gouttegd: This instructs the `make-bridge-ontologies-from-xrefs.pl` script to extract cross-references to MBA-prefixed terms and to generate Uberon/CL bridges to MBA (`uberon-bridge-to-mba.obo`, `cl-bridge-to-mba.obo`, etc.). Removing the line disables the generation of such bridges, which is what is wanted here since we are replacing them with custom, externally maintained bridges.

src/ontology/uberon.Makefile
gouttegd: @matentzn Disagree: Short options are nice on the command line, but in scripts and Makefiles, I think long, explicit options should be preferred.

matentzn: I agree with that in general, I am lobbying for that. but not use $@ introduces unnecessary redundancy in Makefiles. use then `--output $@` at least?

src/ontology/uberon.Makefile
```diff
@@ -1238,7 +1254,7 @@ $(BRIDGEDIR)/%.owl: $(BRIDGEDIR)/%.obo
 make-bridge-ontologies-from-xrefs.pl:
 	cp $(SCRIPTSDIR)/make-bridge-ontologies-from-xrefs.pl $@
 
-$(TMPDIR)/bridges: $(TMPDIR)/seed.obo $(TMPDIR)/cl-with-xrefs.obo $(TMPDIR)/cl-zfa-xrefs.obo $(BRIDGEDIR)/uberon-bridge-to-nifstd.owl make-bridge-ontologies-from-xrefs.pl $(REPORTDIR)/life-cycle-xrefs.txt
+$(TMPDIR)/bridges: $(TMPDIR)/seed.obo $(TMPDIR)/cl-with-xrefs.obo $(TMPDIR)/cl-zfa-xrefs.obo $(BRIDGEDIR)/uberon-bridge-to-nifstd.owl make-bridge-ontologies-from-xrefs.pl $(REPORTDIR)/life-cycle-xrefs.txt $(BRIDGEDIR)/uberon-bridge-to-mba.owl $(BRIDGEDIR)/uberon-bridge-to-dmba.owl $(BRIDGEDIR)/uberon-bridge-to-mba.obo $(BRIDGEDIR)/uberon-bridge-to-dmba.obo
```
gouttegd, line 1257: Good, this should be enough to have the custom MBA/DMBA bridges fetched at the same time as we build the other bridges.

Minor nitpick: Instead of listing the MBA/DMBA bridges here, I would rather have another Make variable (say, `CUSTOM_BRIDGES` or something similar) where we could list all the bridges that are generated by custom rules, and then we reference that variable in the dependencies of `$(TMPDIR)/bridges`.

gouttegd (approved):

src/ontology/uberon.Makefile
gouttegd: Sorry I misunderstood your comment, I thought it was solely about `-o` versus `--option`. Yes, I agree about using the target variable (`$@`) wherever possible.

src/ontology/uberon.Makefile
shawntanzk: things I don't think about and @hkir-dev ends up having to clean all my ugly code lol >.<

src/ontology/uberon.Makefile
matentzn: Ok I would create an issue on uberon and assign it to yourself to remember this issue in the future, to prevent it going stale.

shawntanzk:
> Ok I would create an issue on uberon and assign it to yourself to remember this issue in the future, to prevent it going stale.

Already did here: https://github.com/obophenotype/uberon/issues/2537

Issue #2349, opened 2022-03-14
Clarify CARO-Uberon connections


cmungall:
We have two possible approaches:

1. CARO generalizes
2. CARO and Uberon are equivalent

clearly 1 was intended originally, but if we are serious about this, we need to make sure

- uberon generic terms are unique (prefixed with "metazoan")
- make sure bridge files are easy to compose with main uberon release

I am not sure of the value of 1 anymore. I think it would be much simpler if uberon were analogous to CL, where generic structures are not artificially restricted. PO, FAO, and even GO CC would subclass the relevant Uberon structures (a small handful).

See

- #2348

shawntanzk:
added to agenda, hope it's alright if I assign this ticket to you for now @cmungall (trying to make sure all relatively recent tickets have someone assigned) - happy to make changes once there's an action plan. Thanks

balhoff:
For (2) It might be useful to explicitly mark the small set of Uberon pan-life terms. Then add 'present in taxon' statements to them forcing their applicability in various taxa.

balhoff:
Here's a draft set for a generic/pan-life subset:

- [anatomical entity](http://purl.obolibrary.org/obo/UBERON_0001062)
  - [anatomical cluster](http://purl.obolibrary.org/obo/UBERON_0000477)
  - [immaterial anatomical entity](http://purl.obolibrary.org/obo/UBERON_0000466)
    - [anatomical point](http://purl.obolibrary.org/obo/UBERON_0006983)
    - [anatomical space](http://purl.obolibrary.org/obo/UBERON_0000464)
    - [non-material anatomical boundary](http://purl.obolibrary.org/obo/UBERON_0000015)
      - [anatomical line](http://purl.obolibrary.org/obo/UBERON_0006800)
  - [material anatomical entity](http://purl.obolibrary.org/obo/UBERON_0000465)
    - [anatomical collection](http://purl.obolibrary.org/obo/UBERON_0034925)
      - [anatomical row](http://purl.obolibrary.org/obo/UBERON_0034926)
    - [anatomical structure](http://purl.obolibrary.org/obo/UBERON_0000061)
      - [acellular anatomical structure](http://purl.obolibrary.org/obo/UBERON_0000476)
        - [acellular membrane](http://purl.obolibrary.org/obo/UBERON_0005764)
      - [anatomical border](http://purl.obolibrary.org/obo/UBERON_0036304)
[...]

PR #2530, opened 2022-06-14
Make zygote stage part of life cycle only in Metazoa.

balhoff:
Related to #2349. This fixes an issue with taxon constraints in GO for yeast annotations: https://github.com/geneontology/go-ontology/issues/23012


balhoff:
Need to discuss before merging this; now I'm wondering if instead this other axiom should become a Metazoa-specific GCI:

```
'anatomical structure' SubClassOf 'existence ends during or before' some 'death stage'
```

I previously did that for:

```
'anatomical structure' and ('in taxon' some Metazoa) SubClassOf 'existence starts during or after' some 'zygote stage'
```

We need to decide exactly which Uberon terms are pan-life and which are metazoan. Is 'life cycle' general? Is 'anatomical structure' general?

cmungall:
@balhoff I think we should go ahead and merge this. This PR is consistent with your sketch of what are the general terms in #2349 

I'm actually not sure I am following your comment above. zygote is certainly more specific than the beginning of life. but to make start and end mirror each other maybe there should be a generic "coming into existence stage" with zygote stage a subclass?

In any case let's merge this and continue discussing details on the main issue

cmungall (approved):

PR #2538, opened 2022-06-17
Adding Paula's ID range

bvarner-ebi:
As @paulacalipho will be contributing to Uberon going forward, she will need an ID range.


shawntanzk (approved):
May she add 9999 terms :D

matentzn:
@bvarner-ebi since it makes sense twinned with @shawntanzk please only add him as a reviewer.. It creates a lot of noise in our inboxes if we get tagged for reviews like this. Only ever tag as few people as possible!

PR #2553, opened 2022-06-27
Remove bogus alt_ids.

gouttegd:
A handful of terms have bogus alternative IDs that are either of the form `UBERON:XXXX_YYYYYYY` (where `XXXX_YYYYYYY` is itself a CURIE-like string, as in `UBERON:FBbt_00005386`) or of the form `UBERONTEMP:aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee`, where `aaa...` looks like an autogenerated UUID.

This PR removes those bogus alternative identifiers.


matentzn (approved):
I thought I had reviewed all of these but seems I only did that for replaced by..

PR #2558, opened 2022-06-30
Release 2022-06-30

anitacaron:


shawntanzk:
are we not updating the bridges as part of the release anymore? I know there arent any changes except for date so its not a problem, but just wondering

matentzn:
@anitacaron bridges are definetly built.. what commands did you run?

matentzn:
```
uberon:
	$(MAKE) prepare_release IMP=false PAT=false BRI=true CLEANFILES=tmp/merged-uberon-edit.obo
	$(MAKE) copy_additional_files # Probably not needed anymore now that we put everything on GitHub
	$(MAKE) release-diff
```

shawntanzk:
If it helps, docs: http://obophenotype.github.io/uberon/uberon-release/

anitacaron:
Running again here, btw

uberon.obo
```diff
@@ -113,6 +113,7 @@ synonymtypedef: scientific_name ""
 synonymtypedef: SENSU "taxonomic disambiguation"
 synonymtypedef: synonym ""
 synonymtypedef: SYSTEMATIC "a systematic synonym used as the base name for design patterns"
+synonymtypedef: vertebrate ""
```
anitacaron, line 116: I also noticed this new synonym type. Is that right?

matentzn: seems dump. Can you figure out where it comes from form the edit file?

matentzn:
MAybe you need to add `-B` to the make uberon?

uberon.obo
shawntanzk: yeah I saw it too, but it seems to come from a BSPO term (posterior_to) so I figured it was not on uberons side.

uberon.obo
```diff
@@ -283690,15 +283798,18 @@ inverse_of: positively_regulated_by ! positively regulated by
...
 is_a: part_of ! part_of
 
 [Typedef]
 id: posterior_to
 name: posterior to
 name: posterior_to
+namespace: spatial
 def: "x posterior to y iff x is further along the antero-posterior axis than y, towards the body/tail. An antero-posterior axis is an axis that extends through an organism from head end to opposite end of body or tail." [BSPO:cjm]
 def: "x posterior_to y iff x is further along the antero-posterior axis than y, towards the body/tail. An antero-posterior axis is an axis that bisects an organism from head end to opposite end of body or tail." [http://orcid.org/0000-0002-6601-2165]
+synonym: "caudal to" NARROW vertebrate []
```
shawntanzk, line 283812: @anitacaron @matentzn here's the term, when searching in uberon-edit file, I only see the term as a BSPO term

anitacaron: Yeah, I see it in BSPO when the ontology changed to ODK in [this PR](https://github.com/obophenotype/biological-spatial-ontology/pull/30) but haven't updated the import in uberon since 2020.

shawntanzk (approved):

uberon.obo
shawntanzk: Not sure if we need to deal with it, I've approved the PR but maybe best to also get a thumbs up from @matentzn regarding this

matentzn: Its silly but not harmful. thumbs up.

PR #2567, opened 2022-07-07
update mba/dmba make goal link

shawntanzk:
ok, less scary a link now, we have decided to keep the bridge making in ABA_uberon, but have now moved the new bridges to master branch and kept our old stuff in a branch instead. I've also update my need changes to bridge file here


gouttegd (approved):

shawntanzk:
just checked, looks alright, new bridge terms were added, not sure what the lines deleted really are, but spot-checked the bridge itself and some deleted lines and they are still in there and all looked fine

Issue #2569, opened 2022-07-12
Relationship of gallbladder terms is inconsistent


allenbaron:
## Uberon terms
1. gallbladder (UBERON:0002423) 
2. gallbladder primordium (UBERON:0006242)

## Suggested revision
Make `gallbladder` and `gallbladder primordium` subclasses of `biliary system` (UBERON:0002294) instead of `hepatobiliary system` (UBERON:0002423) and `biliary tree` (UBERON:0001173), respectively.

### Current relationships
![image](https://user-images.githubusercontent.com/21225409/178555135-19dff97c-d002-41d2-9e6c-7f35ba2101c7.png)

## Reasons
1. `biliary system` (UBERON:0002294) includes the `gallbladder` in its definition:
    > Organ system subdivision that consists of the organs and ducts that are involved in the production and transportation of bile. In most species this is the **_gallbladder_** and the bile ducts (biliary tree).

2. `gallbladder` and `gallbladder primordium` should have the same subclass of relationships given the definition of `gallbladder primordium`:
    >A sac-like cavity ventral to the liver, lying in the sub-mesodermal space and closed by a thin sheet of ectodermal cells, **_from which the future gall bladder develops_**.

3. The `biliary tree` as defined and as mentioned in the definition of `biliary system` appears to exclude the `gallbladder`, suggesting it should also exclude the `gallbladder primordium`.
    - `biliary system` definition:
        >Organ system subdivision that consists of the organs and ducts that are involved in the production and transportation of bile. In most species this is the _gallbladder **and** the bile ducts (biliary tree)_.

shawntanzk:
Currently 'gall bladder' has the axioms: 
`'contributes to morphology of' some 'biliary tree'` & `'develops from' some 'gall bladder primordium'`
But because these are not part_of relations, they don't show on OLS 

biliary system has the equiv class axiom 
```
'organ system subdivision'
 and ('has part' some 'biliary tree')
 and ('capable of' some 'bile acid biosynthetic process')
 and ('capable of' some 'bile acid and bile salt transport')
```
biliary tree has the equiv class axiom
```
organ
 and ('part of' some 'biliary system')
 and ('composed primarily of' some 'bile duct')
```

[...]

PR #2570, opened 2022-07-13
Edit gallbladder and gallbladder primordium relation to biliary system

shawntanzk:
Fixes https://github.com/obophenotype/uberon/issues/2569


shawntanzk:
> In that case, would it makes sense to remove the 'part of' some 'biliary tree' axiom from 'gall bladder primordium'?

already done in this PR :)

bvarner-ebi:
> > In that case, would it makes sense to remove the 'part of' some 'biliary tree' axiom from 'gall bladder primordium'?
> 
> already done in this PR :) - should probably reflect that in title lol

Ah yes, I see it now... initially thought it was removed from 'gallbladder', not primordium.

bvarner-ebi (approved):

Issue #2449, opened 2022-05-03
Broadmann area should have taxon restrictions


shawntanzk:
Related to https://github.com/obophenotype/uberon/issues/2412

Broadmann area in uberon currently does not have any restrictions on taxon
This has resulted in terms like primary visual cortex being a subclassof Broadmann area which makes it over specific to humans/primates.
Adding taxon restriction to Broadmann area however might break some stuff, hence we should add the restriction (probably primate to be safe), and fix anything that is broken due to the change

shawntanzk:
Only attempt after https://github.com/obophenotype/uberon/pull/2450 is merged in (fixes there has major impact on this)

patrick-lloyd-ray:
There was some discussion on this in our BDS meeting so I just wanted to track down some info and document it somewhere. Brodmann himself (in his 1909, ISBN:978-0387-26917-7), writes (p. 100):

> As a preliminary I have determined the field distribution in the cortex of representatives of seven orders, namely
> 1. primates by man, several cercopithicids and marmosets, 
> 2. prosimians (*96) by the lemur and mouse lemur,
> 3. chiropterans by the flying fox (*97)
> 4. carnivores by the kinkajou (and partially the dog),
> 5. rodents by the rabbit (*98), the ground squirrel and partially the mouse, 
> 6. insectivores by the hedgehog,
> 7. marsupials by the kangaroo and possum.
> I can present finished brain maps of man, guenon, marmoset, flying fox, kinkajou, rabbit, ground squirrel and hedgehog. In addition I have taken limited samples from many other animals and animal groups (orangutan, capuchin monkey, indris, mouse lemur, pipistrelle, cat, weasel, opossum, echidna and others) but have not yet completed a definitive cortical localisational study of the whole hemisphere.

That doesn't mean that everyone still uses Brodmann for these species nor that it's recommended to use them nor that UBERON would want to track all of these, but there are some researchers that still use Brodmann areas for the above species and there is some historical precedent for this. Also, this list is presumably non-exhaustive.

shawntanzk:
Closing given the above comment

PR #2571, opened 2022-07-14
[DO NOT MERGE] Add homo sapien taxon restriction to Brodmann area

shawntanzk:
Let QC run through - might break somewhere

Fixes https://github.com/obophenotype/uberon/issues/2449


shawntanzk:
was 100% expecting something to break, but everything is built and it works! @dosumis & @cmungall asking you to review as I want to make sure this has no implications on the broader context that I should be aware of

shawntanzk:
hold off until https://github.com/obophenotype/uberon/issues/2449#issuecomment-1209986467 is discussed

shawntanzk:
Give the broader use of Broadmann area, will close this

Issue #2573, opened 2022-07-18
hormion


megbalk:
**Preferred term label:**
hormion

**Synonyms**
vomerine notch

**Definition (free text, please give PubMed ID)**
aboral border of the vomer bone along the median plane, where the vomer bone and sphenoid bones meet.

database_xref: ISBN: 9780873659505

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
inheres in some (anterior part of some vomer)
zone of bone organ
mesoderm-derived structure

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0003-2699-3066
@helenamachado: https://orcid.org/000-0001-8415-4848

megbalk:
@RDruzinsky how about:
Craniometric point along the median plane at the ventral (inferior in humans) end of the suture
of the vomer and sphenoid bones.

we want to specify median plane as to not have it confused with the other two more aboral points on either side.

(see this [image](http://locatelambda.org/wp-content/uploads/2018/02/hormion.jpg))

RDruzinsky:
Sounds good to me.

megbalk:
there isn't an "anterior" term in UBERON, so I changed "inheres in some (anterior part of some vomer)" to "'part of' some 'vomer'"

shawntanzk:
jsut a heads up (not fully following this discussion, just vaguely saw things that might use anterior_to or ventral_to in complex axioms) be careful with nested axioms in uberon yeah - if I rmbr right, OBO format cant hand nested expressions

RDruzinsky:
Aboral is not the proper term for this definition. Also, this is a point in
space where two bones meet, so I don't think that inheres in some vomer is
correct, either.

RDruzinsky:
I like this better.

PR #2576, opened 2022-07-20
adding hormion

megbalk:
issue #2573


src/ontology/uberon-edit.obo
```diff
@@ -222484,6 +222484,17 @@ relationship: dc-contributor http://orcid.org/0000-0003-2699-3066
...
 
+[Term]
+id: UBERON:7500106
+name: hormion
+def: "Craniometric point along the median plane at the ventral (inferior in humans) end of the suture of the vomer and sphenoid bones." [ISBN:9780873659505]
+synonym: "vomerine notch" EXACT []
+is_a: UBERON:0004120 ! mesoderm-derived structure
+is_a: UBERON:0005913 ! zone of bone organ
+relationship: part_of UBERON:0002396 ! vomer
+property_value: http://purl.org/dc/elements/1.1/contributor http://orcid.org/0000-0003-2699-3066
```
shawntanzk, line 222495: should be dc terms, not dc elements :)

shawntanzk: protege > preferences > new entities metadata > creator property> "http://purl.org/dc/terms/contributor" :) lmk if any of that is unclear

megbalk:
@shawntanzk I do not get that same error when I run the reasoner (ELK).

Jugal point (UBERON:0013423) is classified the same way.

shawntanzk:
The unsat is not from uberon-edit itself but a product that is used for the report tha the QC uses to check something. Seems to merge uberon-edit with bfo, ro, and bfo bridge. Not sure about the other term, will check again next week, but probably still need to fix the unsat

shawntanzk:
@megbalk 

I think this is your offending axiom: 
[UBERON_7500106](http://purl.obolibrary.org/obo/UBERON_7500106) SubClassOf [zone of bone organ](http://purl.obolibrary.org/obo/UBERON_0005913)

From my understand of your PR, hormion is a point which is immaterial. However zone of bone organ is a material thing - it refers to an actual bone structure itself. 

I assume this axiom would cause issue too: 

[UBERON_7500106](http://purl.obolibrary.org/obo/UBERON_7500106) SubClassOf [mesoderm-derived structure](http://purl.obolibrary.org/obo/UBERON_0004120)

Perhaps consider using 'part of' instead of subclass

megbalk:
fixed!

megbalk:
@shawntanzk do my changes look ok?

shawntanzk:
QC checks have passed, I'll leave @bvarner-ebi to review content :) thanks

src/ontology/uberon-edit.obo
```diff
@@ -222485,6 +222485,16 @@ relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
 relationship: part_of UBERON:7500094 ! tubercle of calcaneus
 property_value: dcterms-date 2022-05-19T21:16:42Z xsd:dateTime
 
+[Term]
+id: UBERON:7500106
+name: hormion
+def: "Craniometric point along the midline of the sphenovomerine suture on the ventral (inferior in humans) side." [https://locatelambda.org/manual, ISBN:9780873659505]
```
bvarner-ebi, line 222491: Note: [URL in dbxref](https://locatelambda.org/manual) has the following:
defintion: "The most posterior midline point on the vomer."
comment: "Hormion is taken posterior to the vomer at the midline, not into the anterior deflection if it is present."
illustration: http://locatelambda.org/wp-content/uploads/2018/02/hormion.jpg

and from https://medical-dictionary.thefreedictionary.com/hormion:
"A craniometric point at the junction of the posterior border of the vomer with the sphenoid bone." 

Cannot access ISBN.

@megbalk,
Would something like this work?
"A craniometric point at the junction of the posterior midline border of the vomer with the sphenoid bone."
dbxrefs above.

src/ontology/uberon-edit.obo
```diff
@@ -222485,6 +222485,16 @@ relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
...
+synonym: "vomerine notch" EXACT []
```
bvarner-ebi, line 222492: Do you have a dbxref to support this being an exact synonym?
I see it mentioned [here](https://palaeo-electronica.org/content/2019/2417-narial-morphology), and I'm not sure if it's the same. If it is indeed the same, it says it is only found in mammals, so a taxon restriction may be appropriate and they provide another synonym: incisura vomeris (Latin).

bvarner-ebi (commented):
See inline comments.

src/ontology/uberon-edit.obo
megbalk: you're right, I'll remove it.

megbalk:
@shawntanzk @bvarner-ebi could one of you review this, please?

shawntanzk:
@bvarner-ebi - hope its ok if I leave you to do the final review for this. Thanks

bvarner-ebi (approved):
Approved with the following caveats:
- Have not validated the ISBN dbxref.
- Consider most straightforward text definition provided by https://locatelambda.org/manual: "The most posterior midline point on the vomer."

megbalk:
here is a link to the dbxref: https://www.biblio.com/9780873659505. Let me know if I need to change how I'm reporting as this has been an issue before

I don't mind the definition provided by https://locatelambda.org/manual, but I do want to note that it is on the suture.

Issue #2229, opened 2022-01-05
mediodorsal thalamus parts, missing term 'MDM (medial part) of Mediodorsal thalamic nucleus' ?


ANiknejad:
Hi Uberon Team,

Happy New Year to All of You!

Here a recent paper with mediodorsal thalamus (MD) subdivisions (medial, central, and lateral):
https://pubmed.ncbi.nlm.nih.gov/34184265/

So far in Uberon we have:

MDC (central part) -> UBERON:0035113 central part of mediodorsal nucleus of the thalamus
MDL (lateral part) -> UBERON:0035114 lateral part of mediodorsal nucleus of the thalamus
but none MDM medial part of mediodorsal nucleus of the thalamus

However it exists 

UBERON:0002647 magnocellular part of medial dorsal nucleus
with synonym: "MDM" BROAD ABBREVIATION [BIRNLEX:1262, NIFSTD:NeuroNames_abbrevSource]

https://scicrunch.org/scicrunch/interlex/view/ilx_0106473

I am wondering if UBERON:0002647 can indeed be considered as the class describing MDM (medial part) of Mediodorsal thalamic nucleus ? or if there is a misunderstanding about the 'm', for 'medial' or 'magnocellular'?

If UBERON:0002647 does not describe the medial part of the mediodorsal nucleus of the thalamus, we then need a new class as below:

**Preferred term label: medial part of mediodorsal nucleus of the thalamus**

**synonym: "mediodorsal nucleus of the thalamus, medial part"**

**Definition to be retrieved from PMID: 34184265

**relationship: part_of UBERON:0002739 ! medial dorsal nucleus of thalamus**

https://orcid.org/0000-0003-3308-6245

Thank you in advance for your feedback,
Cheers,

Anne

shawntanzk:
@nicolevasilevsky given this is brain related, happy to look into it, you can remove yourself from assignee if you want :) thanks

nicolevasilevsky:
great, thanks @shawntanzk!

shawntanzk:
[...]

PR #2587, opened 2022-08-04
Add synonyms for magnocellular part of medial dorsal nucleus

shawntanzk:
Fixes https://github.com/obophenotype/uberon/issues/2229


shawntanzk:
@patrick-lloyd-ray ignore the failed actions - its just automated diff and we don't really need it for such a simple PR anyway so I'm not going to try to re-run -> its just memory issues probably

patrick-lloyd-ray (approved):

PR #2592, opened 2022-08-09
add magnocellular reticular nucleus

shawntanzk:


src/ontology/uberon-edit.obo
```diff
@@ -224404,6 +224404,16 @@ relationship: dc-contributor https://orcid.org/0000-0002-5480-427X
 relationship: in_taxon NCBITaxon:9669 ! Mustela putorius furo
 relationship: part_of UBERON:0002104 ! visual system
 
+[Term]
+id: UBERON:8440073
+name: magnocellular reticular nucleus
+def: "A nucleus of the brain that is the most ventral portion of the gigantocellular nucleus in the medullary reticular formation, between the lateral paragigantocellular nucleus and the midline or nucleus raphe magnus. Functionally it belongs to the reticular formation (functional) of the subcortical motor system." [NeuroNames:1801]
+synonym: "MARN" EXACT ABBREVIATION [MBA:307]
```
cmungall, line 224411: Add a class xref to MBA:307

(it's an anti-pattern to have a term as provenance when that term is not equivalent or an exact match)

cmungall (changes requested):

src/ontology/uberon-edit.obo
shawntanzk: whoops xref to MBA:307 should defs be there, forgot it. Thanks for the catch

patrick-lloyd-ray (approved):

Issue #286, opened 2013-07-02
Nerve roots vs nerve tracts


cmungall:
FMA has the following hierarchy under 'root of cranial nerve'
- FMA:5981 ! Nerve root
  - FMA:52609 ! Root of cranial nerve [SYNONYM: "Cranial nerve root" (exact)] [SYNONYM: "Cranial neural root" (exact)]
    - FMA:52610 ! Root of trigeminal nerve
    - FMA:53418 ! Root of facial nerve
    - FMA:71116 ! Trochlear nerve tract
    - FMA:72457 ! Central part of oculomotor nerve
    - FMA:72491 ! Abducens nerve tract **\* 
    - FMA:72632 ! Hypoglossal nerve tract
    - FMA:84573 ! Root of optic tract

Note that nerve tracts are treated the same as roots. This is confirmed by looking at the synonyms:
- FMA:52609 ! Root of cranial nerve [SYNONYM: "Cranial nerve root" (exact)] [SYNONYM: "Cranial neural root" (exact)]
- FMA:52610 ! Root of trigeminal nerve [SYNONYM: "Trigeminal neural root" (exact)]
- FMA:53418 ! Root of facial nerve [SYNONYM: "Central part of facial nerve" (exact)] [SYNONYM: "Facial nerve fibers" (exact)] [SYNONYM: "Facial neural root" (exact)]
- FMA:71116 ! Trochlear nerve tract [SYNONYM: "Central part of trochlear nerve" (exact)] [SYNONYM: "Trochlear nerve fibers" (exact)] [SYNONYM: "Trochlear nerve root" (exact)]
- FMA:72457 ! Central part of oculomotor nerve [SYNONYM: "Oculomotor nerve fibers" (exact)] [SYNONYM: "Root of oculomotor nerve" (exact)]
- FMA:72491 ! Abducens nerve tract **\*  [SYNONYM: "Abducens nerve fibers" (exact)] [SYNONYM: "Central part of abducens nerve" (exact)] [SYNONYM: "Root of abducens nerve" (exact)]
- FMA:72632 ! Hypoglossal nerve tract [SYNONYM: "Central part of hypoglossal nerve" (exact)] [SYNONYM: "Hypoglossal nerve fibers" (exact)] [SYNONYM: "Root of hypoglossal nerve" (exact)]
- FMA:84573 ! Root of optic tract [SYNONYM: "Optic tract root" (exact)]

In contrast, NIF GA creates separate classes for tracts/fibers and roots.

[...]

PR #2594, opened 2022-08-10
Add interstitial nucleus of the vestibular nerve

shawntanzk:
@patrick-lloyd-ray I don't quite understand this term. Neuronames suggest it to be part of a nerve, but ABA suggests that it isn't. I think that's why this is failing - it is nucleus of brain (ABA suggests this) which has to be gray matter, but also part_of a nerve, which should be white matter. Need help here


patrick-lloyd-ray:
Yeah, this is a weird one. 

Here's an image from ABA of the interstitial nucleus of the vestibulocochlear nerve (highlighted in purple, labeled I8) from human. The structure labeled r8 is the root of vestibulocochlear nerve. The interstitial nucleus of the vestibular nerve would be just those parts of I8 that are in the vestibular nerve portion of the vestibulocochlear nerve, which is just as you say in the definition. 

Screen Shot 2022-08-10 at 12 42 44 PM

So, I think the definition is fine as you have it. I think the problem is somewhere in the axioms upstream.

The superclass UBERON:0002308 (nucleus of brain), is "A neural nucleus that is part of the brain." which is equivalent to "neural nucleus" and part_of some "brain". It also has the subclass axiom `part_of some brain gray matter`.

UBERON:0000125 (neural nucleus) is "A spatially aggregated collection of nerve cell bodies in the CNS, consisting of one or more subpopulations that share cell type, chemical phenotype, and connections, and including nearby cells that share the same cell type, chemical phenotype, and connections." It has a subclass axiom of `composed_primarily_of some gray matter`, which seems fine to me. 

UBERON:0003528 (brain gray matter) is defined as "A gray matter that is part of a brain [Automatically generated definition]." UBERON:0003544 (brain white matter) is defined as "The regions of the brain that are largely or entirely composed of myelinated nerve cell axons and contain few or no neural cell bodies or dendrites."

I think that the problem is here: the cells that comprise the interstitial nucleus of the vestibular nerve are some of the few with soma in white matter -- i.e., they are "gray matter" surrounded by "white matter." 

The gray matter annotation on "nucleus of brain" seems to be the offending axiom here, but I don't want to recommend changing it because I suspect that it will have serious consequences for other classes. Maybe I'm misreading the problem, though, so let me know if this is consistent with what you are seeing.

shawntanzk:
> The gray matter annotation on "nucleus of brain" seems to be the offending axiom here, but I don't want to recommend changing it because I suspect that it will have serious consequences for other classes. Maybe I'm misreading the problem, though, so let me know if this is consistent with what you are seeing.


You are spot on on the difficulties I'm having here, and I agree with you in not wanting to change something as high up in the hierarchy as "nucleus of brain". We can fix it to UBERON:0000125 (neural nucleus) - but I think if we assume partonomy as a relationship in the hierarchy of ABA, it will still be a problem. That being said, we have talked about how using partonomy for all relationships has limitations. I think for now I will change the parent to neural nucleus. Thanks heaps! :)

shawntanzk:
@patrick-lloyd-ray seems like it still faces the same problem - it is still unsat at the materialised stage, probably due to the white matter grey matter thing. I think this case now calls for relooking at some axioms higher up, would be interested to get @dosumis or @cmungall view on this :)

shawntanzk:
@dosumis & @cmungall - it seems to me that the following axiom is wrong as there are defs nerves in the CNS too:
[nerve](http://purl.obolibrary.org/obo/UBERON_0001021) SubClassOf [part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [peripheral nervous system](http://purl.obolibrary.org/obo/UBERON_0000010) 
As this is quite high up I'd rather have a thumbs up from you before removing it.

balhoff:
> @dosumis & @cmungall - it seems to me that the following axiom is wrong as there are defs nerves in the CNS too:
> [nerve](http://purl.obolibrary.org/obo/UBERON_0001021) SubClassOf [part_of](http://purl.obolibrary.org/obo/BFO_0000050) some [peripheral nervous system](http://purl.obolibrary.org/obo/UBERON_0000010)
> As this is quite high up I'd rather have a thumbs up from you before removing it.

@shawntanzk careful, you're going to strike a nerve with that one!

dosumis:
Is this a nerve-root rather than part of a nerve?

see https://github.com/obophenotype/uberon/issues/286

shawntanzk:
Thanks @dosumis - did not know that whole history. Have changed it to vestibulocochlear nerve root instead, should fix it :) Will let QC run through and hopefully all will be alright

dosumis (approved):
Seems to have fixed it and to be a reasonable compromise.

Issue #2586, opened 2022-08-01
basion


helenamachado:
**Preferred term label:**
basion

**Synonyms**

**Definition (free text, please give PubMed ID)**

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**

**Your nano-attribution (ORCID)**

https://orcid.org/0000-0001-8415-4848

PR #2596, opened 2022-08-10
adding basion

megbalk:
issue #2586


src/ontology/uberon-edit.obo
```diff
@@ -222488,6 +222488,14 @@ relationship: dc-contributor http://orcid.org/0000-0003-2699-3066
 relationship: part_of UBERON:7500094 ! tubercle of calcaneus
 property_value: dcterms-date 2022-05-19T21:16:42Z xsd:dateTime
 
+[Term]
+id: UBERON:7500109
+name: basion
+def: "An anatomical point on the ventral border of foramen magnum in the median plane." [https://www.sciencedirect.com/topics/veterinary-science-and-veterinary-medicine/basion]
```
bvarner-ebi, line 222494: The dbxref seems to a list of search results that is subject to change. 
"These pages are auto-generated by ScienceDirect using heuristic and machine-learning approaches to extract relevant information from our extensive collection of content. We compile this information on a topic-by-topic basis providing the reader both depth and breadth on a specific area of interest."

Can you find a stable reference to support this definition?

src/ontology/uberon-edit.obo
```diff
@@ -222488,6 +222488,14 @@ relationship: dc-contributor http://orcid.org/0000-0003-2699-3066
...
+is_a: UBERON:0003687 ! foramen magnum
```
bvarner-ebi, line 222495: Should this be 'part of' instead of 'is a'?

bvarner-ebi (changes requested):
See inline comments.

megbalk:
thank you for catching those, they should be fixed.

src/ontology/uberon-edit.obo
```diff
@@ -222488,6 +222488,14 @@ relationship: dc-contributor http://orcid.org/0000-0003-2699-3066
 relationship: part_of UBERON:7500094 ! tubercle of calcaneus
 property_value: dcterms-date 2022-05-19T21:16:42Z xsd:dateTime
 
+[Term]
+id: UBERON:7500109
+name: basion
+def: "An anatomical point on the ventral border of foramen magnum in the median plane." [ISBN:9780873659505]
```
bvarner-ebi, line 222494: I wasn't able to find an accessible version of the dbxref to cross check, but this definition seems like it would encompass two points on the ventral surface of the foramen magnum- both anterior and posterior, since the median plane would cross through both. Based on some quick searches, I think the basion refers only to the point on the anterior margin. https://radiopaedia.org/articles/basion

bvarner-ebi (changes requested):
See inline comment regarding definition.

src/ontology/uberon-edit.obo
megbalk: that's correct, I'll update it to be more specific

bvarner-ebi (approved):

bvarner-ebi:
@megbalk, changes approved, but I think @shawntanzk is in the middle of making a new Uberon release, so please wait to merge until he confirms it is OK.

bvarner-ebi:
Hi, @shawntanzk, are the Uberon release issues resolved so @megbalk can merge this?

shawntanzk:
> Hi, @shawntanzk, are the Uberon release issues resolved so @megbalk can merge this?

honestly you can merge it any time lol, it doesn't overly affect our release process, it just might not be in that release, but yes you can go ahead and merge this :)

bvarner-ebi:
> honestly you can merge it any time lol, it doesn't overly affect our release process, it just might not be in that release, but yes you can go ahead and merge this :)

Thanks, @shawntanzk. At the time of [the original comment,](https://github.com/obophenotype/uberon/pull/2596#issuecomment-1219619159) I think there was an issue and merges were on hold.

@megbalk, as noted above, feel free to merge this.

Issue #2598, opened 2022-08-11
synsphenion and intersphenoid suture


megbalk:
**Preferred term label:**
1. intersphenoid suture
2. synsphenion

**Synonyms**
1. exact: intersphenoid synchondrosis

**Definition (free text, please give PubMed ID)**
1. A suture joining the basisphenoid and presphenoid of a sphenoid bone. This fuses at birth in humans.
2. Craniometric point along the midline of the intersphenoid suture.

ref: ISBN: 9780873659505

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
1. facial suture and connects some presphenoid bone and connects some basisphenoid bone.
2. anatomical point and part of some intersphenoid suture.

**Your nano-attribution (ORCID)**

PR #2599, opened 2022-08-11
synsphenion and intersphenoid suture

megbalk:
issue #2598


src/ontology/uberon-edit.obo
```diff
@@ -191,6 +191,7 @@ property_value: doap-SVNRepository https://obo.svn.sourceforge.net/svnroot/obo/u
...
...))\nDeclaration(Class(<http://purl.obolibrary.org/obo/UBERON_7500110>))\nDeclaration(Class(<http://purl.obolibrary.org/obo/UBERON_7500111>))\nDeclaration(ObjectProperty(<http://purl.obolibrary.org/obo/BFO_0000050>))\nDeclaration(ObjectProperty(<http://purl.obolibrary.org/obo/RO_0002176>))\n\n############################\n#   Classes\n############################\n\n# Class: <http://purl.obolibrary.org/obo/UBERON_7500110> (<http://purl.obolibrary.org/obo/UBERON_7500110>)\n\nSubClassOf(<http://purl.obolibrary.org/obo/UBERON_7500110> ObjectIntersectionOf(<http://purl.obolibrary.org/obo/UBERON_0006983> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/BFO_0000050> <http://purl.obolibrary.org/obo/UBERON_7500111>)))\n\n# Class: <http://purl.obolibrary.org/obo/UBERON_7500111> (<http://purl.obolibrary.org/obo/UBERON_7500111>)\n\nSubClassOf(<http://purl.obolibrary.org/obo/UBERON_7500111> ObjectIntersectionOf(<http://purl.obolibrary.org/obo/UBERON_0009199> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/RO_0002176> <http://purl.obolibrary.org/obo/UBERON_0006428>) ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/RO_0002176> <http://purl.obolibrary.org/obo/UBERON_0008828>)))\n\n\n)
```
bvarner-ebi, line 194: Seeing duplicates in the Anonymous Ancestor:
Screenshot 2022-08-12 at 15 51 43

src/ontology/uberon-edit.obo
```diff
@@ -222500,6 +222501,19 @@ intersection_of: connects UBERON:0002396 ! vomer
 relationship: dc-contributor https://orcid.org/0000-0001-8415-4848
 relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
 
+[Term]
+id: UBERON:7500110
+name: synsphenion
+def: "A craniometric point along the midline of the intersphenoid suture." [ISBN:9780873659505]
```
bvarner-ebi, line 222507: Is it any point along the midline or a specifically defined point?

src/ontology/uberon-edit.obo
```diff
@@ -222500,6 +222501,19 @@ intersection_of: connects UBERON:0002396 ! vomer
...
+relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
+
+[Term]
+id: UBERON:7500111
+name: intersphenoid suture
+def: "A suture joining the basisphenoid bone and the presphenoid bone of a sphenoid bone. This fuses at birth in humans." [ISBN:9780873659505]
+synonym: "intersphenoid synchondrosis" EXACT []
```
bvarner-ebi, line 222514: Do you have a dbxref to support this this? Is it the same as the definition?

bvarner-ebi: "This fuses at birth in humans."

From https://www.ajronline.org/doi/abs/10.2214/ajr.104.1.184 (N.B. from 1968):
"The intersphenoid synchondrosis remains patent in normal children until the age of approximately 3 years."

bvarner-ebi (commented):
See inline comments.

src/ontology/uberon-edit.obo
megbalk: thank you, I will add the xrefs and will be sure to do that in the future too.

src/ontology/uberon-edit.obo
megbalk: I think this is the difference in how CARO and UBERON define anatomical entity and anatomical point.

megbalk:
@bvarner-ebi it seems "facial suture" is under "anatomical entity" (CARO) in UBERON. I am not sure why this is, but I can't seem to not have it pop up. "anatomical point" is also an UBERON anatomical entity.

megbalk:
Also, it seems CARO anatomical entity is a subclass of biological entity, then independent continuant. Is there a reason to have them in UBERON?

src/ontology/uberon-edit.obo
balhoff: @megbalk no need for the intersection here. Just use two subclass axioms: 
- `'anatomical point'`
- `'part of' some 'inter sphenoid suture'`

balhoff (changes requested):
See inline comment.

megbalk:
fixed the EQ for intersphenoid suture (axiom was originally under "SubClass Of" and not "Equivalent To". Also separated the "SubClass Of" for synsphenion

megbalk:
Note that axioms under "SubClass Of" cannot be nested, e.g., "Class and (property some class) and (property some class)"

bvarner-ebi (approved):

Issue #2605, opened 2022-08-15
updating definition of 'subcommissural organ' to include hypendymal cell


shawntanzk:
see https://github.com/obophenotype/cell-ontology/pull/1676

PR #2606, opened 2022-08-15
update subcommissural organ def to include hypendymal cell

shawntanzk:
Fixes https://github.com/obophenotype/uberon/issues/2605
Noticed that ref is comments weren't annotated as dbxref - did not remove intext citations as no clear decision was reached on that previously and I'd rather not remove them for now as that removes information.


src/ontology/uberon-edit.obo
```diff
@@ -42870,8 +42870,8 @@ property_value: taxon_notes "present in all fishes and tetrapods" xsd:string {so
 [Term]
 id: UBERON:0002139
 name: subcommissural organ
-def: "The subcommissural organ is a circumventricular organ consisting of ependymal cells which secrete SCO-spondin[WP,partially vetted]." [PMID:9579598, Wikipedia:Subcommissural_organ]
-comment: Occurs throughout the vertebartes [PMID:9579598, DOI:10.1002/ar.1091260210]. First appears in hagfish(Olsson). In many species, including the human, it reaches its full development during embryonic life[PMID:9579598]. During the course of phyletic evolution, one notes various changes. Indeed, in lower vertebrates, the SCO presents a greater degree of enzyme activity than the ependyma while in birds the ependyma demonstrates a more intense activity than the SCO[PMID:479574] Sometimes preceded by flexural organ[DOI:10.1007/BF00303086]
+def: "The subcommissural organ is a circumventricular organ consisting of ependymal and hypendymal cells which secrete SCO-spondin[WP,partially vetted]." [PMID:9579598, Wikipedia:Subcommissural_organ]
```
bvarner-ebi, line 42873: Consider clarifying that hypendymal cells are rare in non-mammals / add to gloss.
Do axioms need to be updated?

shawntanzk: We can add 'has part' some 'hypendymal cell' but I'm hesitant due to the rare in non-mammals thing.

bvarner-ebi: I share that reservation. This is why I suggest making reference to that in the text def.

bvarner-ebi (approved):

Issue #2610, opened 2022-08-16
incoherency caused changes to genitourinary system


shawntanzk:
This commit is probably causing some incoherency issues in Ubergraph:
https://github.com/obophenotype/uberon/commit/40e47c686350786a411775cd2c1340652b6566e1#diff-7dde9e81f71ce11962436529956798fa723e49d1bc8d188cc399d8009c2f72b4
see:
https://github.com/INCATools/ubergraph/issues/82
@bvarner-ebi - I see that this was originally your PR, could you help look into this please? 
@nicolevasilevsky & @cmungall - putting you in assignee list as you were involved with the PR too
Thanks

shawntanzk:
btws the addition of: 
`[reproductive system](http://purl.obolibrary.org/obo/UBERON_0000990) SubClassOf [part of](http://purl.obolibrary.org/obo/BFO_0000050) some [genitourinary system](http://purl.obolibrary.org/obo/UBERON_0004122)`
seems to be the one causing the incoherence - I don't see anything inherently wrong with but perhaps relaxing: 
`[genitourinary system](http://purl.obolibrary.org/obo/UBERON_0004122) EquivalentTo [anatomical system](http://purl.obolibrary.org/obo/UBERON_0000467) and ([has part](http://purl.obolibrary.org/obo/BFO_0000051) some [reproductive system](http://purl.obolibrary.org/obo/UBERON_0000990)) and ([has part](http://purl.obolibrary.org/obo/BFO_0000051) some [renal system](http://purl.obolibrary.org/obo/UBERON_0001008))`
Might be the best modelling solution?

bvarner-ebi:
> not all reproductive systems are part of the genitourinary system 

@shawntanzk, can you point me to an example of this?

shawntanzk:
I think worms reproductive systems that arent part of urinary system 
Reproductive: 
![Screenshot 2022-08-16 at 13 29 27](https://user-images.githubusercontent.com/76212760/184879845-f34dff13-1615-4755-9eb7-64e9bbe8ff19.png)
Urinary: 
![Screenshot 2022-08-16 at 13 32 23](https://user-images.githubusercontent.com/76212760/184880376-08d259ca-9e70-4cb8-af8f-2ba9a54a402b.png)

shawntanzk:
> Would it be related to asexual reproductive systems?

yeah I think so

bvarner-ebi:
> I think worms reproductive systems that arent part of urinary system

[...]

PR #2612, opened 2022-08-16
Remove part of genitourinary system from reproductive system

shawntanzk:
Fixes https://github.com/obophenotype/uberon/issues/2610


src/ontology/uberon-edit.obo
```diff
@@ -9985,7 +9985,6 @@ disjoint_from: UBERON:8450002 ! excretory system
 relationship: existence_ends_during UBERON:0000066 ! fully formed stage
 relationship: fma_set_term FMA:75572
 relationship: has_part UBERON:0000991 ! gonad
-relationship: part_of UBERON:0004122 ! genitourinary system
```
bvarner-ebi, line 9988: Is this being removed to capture asexual reproductive systems?

shawntanzk: see https://github.com/obophenotype/uberon/issues/2610#issuecomment-1216574331

bvarner-ebi (approved):

Issue #2618, opened 2022-08-18
remove date from entries


megbalk:
**Uberon term**
UBERON:7500

**Bug description**
had date as part of preferences when making a new term, will remove them.

PR #2619, opened 2022-08-18
removing date from my entries

megbalk:
issue #2618


shawntanzk:
Is there a reason for removing them? Don't think they are a problem? But also not a problem to remove them so I'll approve, but really no need to remove them tbh.

shawntanzk (approved):

PR #2620, opened 2022-08-18
fix typo Ubeorn

nlharris:
also capitalized Editors Documentation to match Title Case of other references (even though my preference would be to make them all Sentence case)


shawntanzk (approved):
Thanks Nomi :)

PR #2624, opened 2022-08-19
2022-08-19 release

shawntanzk:


shawntanzk:
This removal of disjoint is still here - not sure where its from and if its right
Screenshot 2022-08-19 at 13 18 14

anitacaron:
The `renal system` term was redefined as subclass of `excretory system` so the disjointness changed from `renal system` to `excretory system` - https://github.com/obophenotype/uberon/pull/2497

matentzn (approved):

Issue #2495, opened 2022-05-25
The great base debugging


anitacaron:
EDIT: these unsats are not the problem anymore, look at my comment below.

## [neuroblast](http://purl.obolibrary.org/obo/CL_0000031) SubClassOf [Nothing](http://www.w3.org/2002/07/owl#Nothing) ##

  - [neuroblast](http://purl.obolibrary.org/obo/CL_0000031) SubClassOf [neuronal stem cell](http://purl.obolibrary.org/obo/CL_0000047)
    - [neuronal stem cell](http://purl.obolibrary.org/obo/CL_0000047) SubClassOf [part of](http://purl.obolibrary.org/obo/BFO_0000050) some [central nervous system](http://purl.obolibrary.org/obo/UBERON_0001017)
      - [central nervous system](http://purl.obolibrary.org/obo/UBERON_0001017) SubClassOf [part of](http://purl.obolibrary.org/obo/BFO_0000050) some [nervous system](http://purl.obolibrary.org/obo/UBERON_0001016)
        - [nervous system](http://purl.obolibrary.org/obo/UBERON_0001016) SubClassOf [existence ends during](http://purl.obolibrary.org/obo/RO_0002492) some [fully formed stage](http://purl.obolibrary.org/obo/UBERON_0000066)
          - [fully formed stage](http://purl.obolibrary.org/obo/UBERON_0000066) SubClassOf [preceded by](http://purl.obolibrary.org/obo/BFO_0000062) some [organogenesis stage](http://purl.obolibrary.org/obo/UBERON_0000111)
            - [organogenesis stage](http://purl.obolibrary.org/obo/UBERON_0000111) SubClassOf [preceded by](http://purl.obolibrary.org/obo/BFO_0000062) some [neurula stage](http://purl.obolibrary.org/obo/UBERON_0000110)
              - [neurula stage](http://purl.obolibrary.org/obo/UBERON_0000110) SubClassOf [simultaneous with](http://purl.obolibrary.org/obo/RO_0002082) some [neural tube formation](http://purl.obolibrary.org/obo/GO_0001841)
                - [neural tube formation](http://purl.obolibrary.org/obo/GO_0001841) SubClassOf [part of](http://purl.obolibrary.org/obo/BFO_0000050) some [neural tube development](http://purl.obolibrary.org/obo/GO_0021915)
[...]

PR #2626, opened 2022-08-19
Change ganglion's parent

anitacaron:
Related to https://github.com/obophenotype/uberon/issues/2495#issuecomment-1187364719


shawntanzk:
> or we could just generalise to the parent class (I'm sure very many ganglia also have whole cells).

Had the same thought - I think it can't be multi cell part structure as it would have complete cells as a part. Maybe broader to anatomical cluster?

dosumis:
(Sorry for slow response.  I actually wrote a response on this some time ago, but now realised I didn't post it - trying to reconstruct my argument here).

General issue: granularity constraints on super general/abstract classes used across species are a bad idea.  They will almost inevitable lead to problems like this, where we spend time arguing over things that few, if any biologists care about (see [The Pici principle](https://www.russet.org.uk/blog/2011/10/the-pici-principle-what-you-should-not-say/)).  Even granularity constraints on much more specific classes can cause problems across species, e.g.  2 structures  (muscles, sensory structures, glands) may be homologous across insects but consist of  a single cell in some species, and a small cluster of cells in others.  

Specific issue: If you look at both the official Uberon  test def and the external definitions  that uberon aggregates, they all emphasise that consisting of (mostly/primarily) an aggregation of neuronal cell bodies (soma) is important.  But none of them say that whole cells are not present.  The first of the external definitions could cover a whole arthropod brain, which has lots of whole cells, including neurons.  OTOH I know of a few uses in insects which refer entirely to an aggregation of neuronal cell bodies (see ['dorsal organ ganglion'](https://www.ebi.ac.uk/ols/ontologies/fbbt/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FFBbt_00007227)).

Conclusion: push the term up under 'anatomical structure'.  Consider the same approach in general when we encounter these problems in Uberon.  The only argument against is that browsing a flattened structure is harder, but I think that argument is bogus too:  If we can't decide where these classes should live, why should we expect users to know which branch to find them in while browsing.

dosumis (changes requested):
Please move up under 'anatomical structure'.  That way we are unlikely to be back here discussing this issue again.  See comment for details.

dosumis (approved):

Issue #1873, opened 2021-05-27
UBERON:0002030 nipple should be part_of UBERON:0001911 mammary gland


paola-scibite:
**Uberon term**
UBERON:0002030 nipple

**Suggested revision and reasons**
UBERON:0002030 nipple isn't linked to mammary gland (other than indirectly via 
develops_from some mammary ridge
and
mammary duct channels_into some nipple).
Please consider adding 
nipple part_of mammary gland
because in species that have nipples, these are always part of a mammary gland. 

Thanks,
Paola

Update: on further analysis, I see that 'nipple' has 2 general class axioms:
nipple and ('part of' some 'Mus musculus') SubClassOf 'part of' some 'mammary gland'
nipple and ('part of' some 'Homo sapiens') SubClassOf 'part of' some breast
I still think that 'nipple' should always be part of 'mammary gland', but the existence of the axioms above suggests that there may have been previous discussion on this, so I'll leave things as they are and close.

RDruzinsky:
I would argue that mammary glands and nipples (or teats) are 'part of' the
breast. Nipples or teats are not parts of the mammary glands.

Robert E. Druzinsky, Ph.D.
Clinical Associate Professor
Dept. of Oral Biology
College of Dentistry
University of Illinois at Chicago
801 S. Paulina
Chicago, IL 60612
***@***.***

Office: 312-996-0406
Lab: 312-996-0629
Website: www.peerj.com/RobertDruzinsky

paolaroncaglia:
@RDruzinsky @cmungall 
Thanks for your feedback,

> > I would argue that mammary glands and nipples (or teats) are 'part of' the
> > breast.
> 
> remember uberon is multi-species

Exactly. Only the '_thoracic_ mammary gland' is part_of 'breast'. Nipples in mammals are generally located on the whole trunk, and cows' teats would be part_of 'abdominal mammary gland'...

> > Nipples or teats are not parts of the mammary glands.

But 'areola' and 'mammary fat pad' are - here's the current view:

![Screenshot 2022-03-25 at 10 13 13](https://user-images.githubusercontent.com/4982516/160091249-34b02d7c-0c44-432e-a564-868ea7e4f27b.png)

If areolas and fat tissue are considered part_of mammary glands, rather than adjacent to or similar, then nipples and teats should be too. I suspect that the difference in the existing GCIs  
nipple and ('part of' some 'Mus musculus') SubClassOf 'part of' some 'mammary gland'
[...]

PR #2628, opened 2022-08-23
Make nipple part of mammary gland

shawntanzk:
Fixes https://github.com/obophenotype/uberon/issues/1873
I  _**think**_ this is the conclusion of the above
also tagging @raymond91125
Asking @bvarner-ebi to review too as I know you have some reservations about this

### DIFF
--- a/src/ontology/bridge/uberon-bridge-to-emapa.obo
+++ b/src/ontology/bridge/uberon-bridge-to-emapa.obo
@@ -1,5 +1,5 @@
 ontology: uberon/bridge/uberon-bridge-to-emapa
-data-version: 2022-08-19
+data-version: 2022-09-06
 property_value: seeAlso "http://purl.obolibrary.org/obo/uberon/references/reference_0000026" xsd:anyURI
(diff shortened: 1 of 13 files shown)


src/ontology/uberon-edit.obo
```diff
@@ -39467,6 +39467,7 @@ relationship: part_of UBERON:0002110 ! gallbladder
 id: UBERON:0002030
 name: nipple
 def: "Projection of skin containing the outlets for 15-20 lactiferous ducts arranged cylindrically around the tip." [FMA:67771, Wikipedia:Nipple]
+comment: Nipple being part_of mammary gland is due to a practical modelling in uberon that treats mammary gland as including adnexa. For discussion of this, please refre to https://github.com/obophenotype/uberon/issues/1873.
```
bvarner-ebi, line 39470: "refre" typo

bvarner-ebi (changes requested):
As mentioned in the linked issue, I do not support this modelling. A nipple is not necessarily part of a mammary gland, even if mammary gland is defined to include adnexa.

PR #2633, opened 2022-08-26
add rhinal incisure

shawntanzk:


src/ontology/uberon-edit.obo
```diff
@@ -224523,6 +224523,19 @@ relationship: dc-contributor http://orcid.org/0000-0001-7258-9596
 relationship: has_part UBERON:0001384 ! primary motor cortex
 relationship: has_part UBERON:0016634 ! premotor cortex
 
+[Term]
+id: UBERON:8440077
+name: rhinal incisure
+def: "A fissure in the cerebral cortex that separates the tenia tecta rostrally from the orbital gyri and caudally from the infralimbic cortex" [NeuroNames:1975]
+xref: MBA:1078
+is_a: UBERON:0000464 ! anatomical space
```
bvarner-ebi, line 224531: Consider replacing with`'surrounded by' some 'anatomical structure'`.

'anatomical space' text definition: "Non-material anatomical entity of three dimensions, that is generated by morphogenetic or other physiologic processes; is surrounded by one or more anatomical structures; contains one or more organism substances or anatomical structures."

It is not clear to me if the 'rhinal incisure' "contains one or more organism substances or anatomical structures". 
Also, other 'fissure' and 'sulcus' terms in Uberon are not classified as anatomical spaces.

src/ontology/uberon-edit.obo
```diff
@@ -224523,6 +224523,19 @@ relationship: dc-contributor http://orcid.org/0000-0001-7258-9596
 relationship: has_part UBERON:0001384 ! primary motor cortex
 relationship: has_part UBERON:0016634 ! premotor cortex
 
+[Term]
+id: UBERON:8440077
+name: rhinal incisure
+def: "A fissure in the cerebral cortex that separates the tenia tecta rostrally from the orbital gyri and caudally from the infralimbic cortex" [NeuroNames:1975]
+xref: MBA:1078
```
bvarner-ebi, line 224530: Not able to cross reference this by ID "1078".

bvarner-ebi (approved):
See inline comments.

src/ontology/uberon-edit.obo
shawntanzk: > Consider replacing with'surrounded by' some 'anatomical structure'.

anatomical space is the superclass (which is a requirement), so really shouldn't replace it with that. Anatomical space also has `'surrounded by' some 'anatomical structure'` so it will infer that axiom. 

> It is not clear to me if the 'rhinal incisure' "contains one or more organism substances or anatomical structures".

This is a strange one - given that anatomical space is a immaterial entity, it really shouldn't contain a structure or substance. 

> Also, other 'fissure' and 'sulcus' terms in Uberon are not classified as anatomical spaces.

Another odd thing, there defs are fissures that are classified as anatomical space - see http://purl.obolibrary.org/obo/UBERON_0014466 and its child nodes. But I also see things like Optic fissure (http://purl.obolibrary.org/obo/UBERON_0005412) which is classified under surface groove which is material. I think this comes down to if we consider the fissure (or sulcus or incisure) the space of the fold, or the fold itself. I would argue that it should be the space as the fold itself is occupied by a structure. It is a really good general question though and something we should get a more unified modelling on. Especially important with immaterial and material disjointness

Issue #2617, opened 2022-08-18
cheek tooth row


megbalk:
**Preferred term label:**
cheek tooth row

**Synonyms**
cheektooth row

**Definition (free text, please give PubMed ID)**
Anatomical cluster consisting of adjacent cheek teeth constituting a row.

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
'anatomical row' and ('has member' some 'cheek tooth')

**Your nano-attribution (ORCID)**

megbalk:
waiting for cheek tooth to merge

PR #2639, opened 2022-09-02
creating cheek tooth row

megbalk:
issue #2617


bvarner-ebi (approved):
Consider aligning text definition with logical definition:
An anatomical row consisting of adjacent cheek teeth.
OR
A row of cheek teeth.

megbalk:
@bvarner-ebi I modeled it after "tooth row". Is it preferred to align with previous examples or the logical definition?

bvarner-ebi:
> @bvarner-ebi I modeled it after "tooth row". Is it preferred to align with previous examples or the logical definition?

Ideally both, assuming the previous examples are modelled correctly. I would provide the same feedback for the text definition of 'tooth row'. I was trained to match the text definition with the logical definition. 
If you would like to get more feedback, you could add to the next Uberon call agenda. The text definition does not directly impact reasoning, so it should not cause any conflicts as it is.

megbalk:
@bvarner-ebi @shawntanzk is this ok to merge? It seems not all the checks passed, but I forget which matter.

shawntanzk:
```
 :: ERROR: definition-lacks-dbxref: UBERON:7500114
```
Prob just add your ORCID to it and it will be fine

PR #2640, opened 2022-09-05
Cross Term Label-Synonym Violation QC

vasvir:
This PR detects synonym-synonym and label-synonym across different Terms (UBERON Ids).

Another implementation (but with the same logic and functionality) of this check was
used to detect and fix (merged) the followings issues:

* https://github.com/obophenotype/uberon/pull/2465
* https://github.com/obophenotype/uberon/pull/2470
* https://github.com/obophenotype/uberon/pull/2506

The backstory is in https://github.com/obophenotype/uberon/issues/2424

The existing
https://robot.obolibrary.org/report_queries/duplicate_exact_synonym detects
only synonym-synonym violations. It was in reported in
https://github.com/ontodev/robot/issues/607 but closed at the time.

The existing
https://robot.obolibrary.org/report_queries/duplicate_label_synonym check
for all label-synonym duplication both in the same Term and across
different terms. The duplicity in the same term is a warning but the
duplicity across different terms should be an error.

In this PR the test is case insensitive because a downstream applications
have a strong possibility to be case insensitive in order to maximize
coverage.

Currently this test flags 3 errors and as such cannot be merged. I will try
to address the issues revealed by this test and are shown below in future PR hoping
that at some point this PR will be merged.

entity,property,value
http://purl.obolibrary.org/obo/UBERON_0002633,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,nV
http://purl.obolibrary.org/obo/UBERON_2005093,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,NV
http://purl.obolibrary.org/obo/UBERON_2000438,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,phy
http://purl.obolibrary.org/obo/UBERON_8440028,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,PHY
http://purl.obolibrary.org/obo/UBERON_0013560,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,Prelimbic area
http://purl.obolibrary.org/obo/UBERON_8440032,http://www.w3.org/2000/01/rdf-schema#label,prelimbic area

Thanks @matentzn and @anitacaron for the guidance and the hand-holding.

PR #2641, opened 2022-09-05
Polysemy: NV. Demote synonymity to RELATED in UBERON:2005093

vasvir:
This PR tries to resolve the 4 of 4 of issues mentioned in #2424

The issue also blocks QC PR #2640

* nV is listed as an EXACT synonym of UBERON:0002633 motor nucleus of trigeminal nerve
* NV is listed as an EXACT synonym of UBERON:2005093 nasal vein

The proposed solution demotes NV to RELATED for nasal vein.  From a cursory
web search in google and pubmed looks like that "nv" is more often
associated with "motor nucleus of trigeminal nerve" than the "nasal vein".

In my opinion both synonyms should be deleted since they are not used widely
enough and they represent a significant risk of misinterpretation by any
automated system. Having said the path of minimum change was selected for this PR.


matentzn:
Ideally, there would be a way to capture acronyms as a special synonym type to avoid ambiguity like this.. For now I am ok with this change, as it is tiny, but allows a nice QC check to be switched on. Technically, I would perceive this case here as an exception to the check though!

shawntanzk:
see https://github.com/obophenotype/uberon/pull/2640#issuecomment-1252539710

shawntanzk (approved):
looks good - probably would still let qc run through, but can't see why it would fail with a change like that. Thanks heaps :)

vasvir:
Well all my PR are failing at post_comment...

shawntanzk:
> Well all my PR are failing at post_comment...

Yeah sorry about that, it requires github secret token something something and cause you arent a member, it has issues I think. I'll merge it in :) thanks

vasvir:
It is somewhat disheartening for new folks... But then again I can understand these things may resist in various spectacular ways when you try to fix them and at the end sometimes it is really difficult.

shawntanzk:
> It is somewhat disheartening for new folks... 

Yeah I totally understand, and I'm sorry about it, but we really honestly appreciate your help with this :) 
The diff thing is annoying, cause it's actually really pushing github actions, but it actually does help (at least me) in the review process. The action failing at comment part when it's not one of us creating the PR - I still can't quite figure out how to get around it, but if you have any suggestions, super welcome :) 
PS if you have any suggestions for anything at all, please do let us know, we are 100% committed to democratising editing the ontology and all, but legit, thanks so much for your contributions!

PR #2642, opened 2022-09-05
Polysemy: prelimbic area Obsolete offending UBERON:8440032

vasvir:
This PR tries to resolve the 5 of 6 of the issues mentioned in #2424 (There are 2 new ones from the last time check was run)

The problem that is addressed here is that:

* "Prelimbic area" is listed as an EXACT synonym of UBERON:0013560 with name: "Brodmann (1909) area 32"
* "prelimbic area" is listed as the label of UBERON:8440032

This creates unnecessary polysemy and it is one of the cases that demonstrates the usefulness of QC #2640

The proposed solution is to obsolete UBERON:8440032 altogether since it is newer and not so developed
as the original one.

I preserved all entries of the obsoleted term but I didn't move the synonyms
as they generally do not feel right (referencing single PMID).

Tagged @elekka since she helped with the biology and @matentzn as requested.


shawntanzk:
> The proposed solution is to obsolete UBERON:8440032 altogether since it is newer and not so developed
as the original one.

We should not do this, instead prelimbic area should be a related synonym. Broadmann area can technically be specific to humans. Rodents for example should not have Broadmann area, but instead prelimbic is correct

cmungall:
@shawntanzk - under your scheme would the xref for MBA:972 be removed from the BA32 class?

vasvir:
Hopefully I got it right this time around.

PR #2643, opened 2022-09-05
Polysemy: PHY. Demote synonymity to RELATED in UBERON:8440028

vasvir:
This PR tries to resolve the 6 of 6 of the issues mentioned in #2424 (There are 2 new ones from the last time check was run)

The issue also blocks QC PR #2640

The problem that is addressed here is that:

* "phy" is listed as an EXACT synonym of UBERON:2000438 with name: "parhypural"
* "PHY" is listed as an EXACT synonym of UBERON:8440028 with name: "Perihypoglossal nuclei"

The proposed solution demotes PHY to RELATED for "Perihypoglossal nuclei".

In my opinion both synonyms should be deleted since they represent a significant risk of misinterpretation by any
automated system. Having said the path of minimum change was selected for this PR.

Hopefully this PR is the last in the polysemy PR series.

Tagged @elekka since she helped with the biology and @matentzn as requested.


shawntanzk:
@dosumis - I'm sure we have discussed this before. 

Personally I don't think it should be removed - automated systems should take into consideration context instead. It is almost inevitable that there will be duplicated synonyms that are rightly exact.

dosumis:
In general, I think we need comprehensive typing of synonyms like this as type:abbreviation - especially where we have a clear reference for use of the abbreviation to refer exactly to the class. NLP can use this as a flag to ignore, or at least assign a lower probability, but this will remain as an index term for driving search interfaces. 

I also think use of ontology structure by NLP could help to disambiguate cases of apparent polysemy.  A hit is more likely to be correct if other hits in the same text have high semantic similarity to it.

cmungall:
I agree with @shawntanzk and @dosumis - and this should all be in the guide here: https://github.com/obophenotype/uberon/wiki/Using-uberon-for-text-mining

Perhaps it gets a bit silly when we get down to single-letter abbreviations but even here, if things are appropriately tagged it should cause no problems:

```
[Term]
id: UBERON:0001947 ! red nucleus
name: red nucleus
synonym: "nucleus rotundus subthalamo-peduncularis" RELATED LATIN [NeuroNames:505]
synonym: "nucleus ruber" EXACT LATIN [TAO:0000552]
synonym: "nucleus ruber" RELATED LATIN [Wikipedia:Red_nucleus]
synonym: "nucleus ruber tegmenti" RELATED LATIN [MP:MP]
synonym: "nucleus ruber tegmenti (Stilling)" RELATED LATIN [NeuroNames:505]
synonym: "R" BROAD ABBREVIATION [BIRNLEX:1478, NIFSTD:NeuroNames_abbrevSource]
synonym: "red nucleus (Burdach)" RELATED [NeuroNames:505]

[Term]
id: UBERON:0002687 ! area X of ventral lateral nucleus
name: area X of ventral lateral nucleus
synonym: "anteromedial part of ventral lateral posterior nucleus (jones)" EXACT []
synonym: "area X" EXACT [FMA:62198]
synonym: "area X of Olszewski" EXACT [BIRNLEX:1379]
synonym: "nucleus lateralis intermedius mediodorsalis situs ventralis medialis" EXACT [BIRNLEX:1379]
synonym: "nucleus ventralis oralis, pars posterior (dewulf)" EXACT [BIRNLEX:1379]
synonym: "nucleus ventro-oralis internus (Hassler)" RELATED LATIN [NeuroNames:342]
synonym: "nucleus ventrooralis internus (hassler)" EXACT [BIRNLEX:1379]
synonym: "nucleus ventrooralis internus, superior part" EXACT [BIRNLEX:1379]
synonym: "X" BROAD ABBREVIATION [BIRNLEX:1379, NIFSTD:NeuroNames_abbrevSource]

[Term]
id: UBERON:0014759 ! entopallium
name: entopallium
synonym: "core region of ectostriatum" RELATED [PMID:14681937]
synonym: "E" RELATED ABBREVIATION [PMID:16041718]
```

cmungall (changes requested):
add ABBREVIATION

shawntanzk:
Hey I think this eventually needs input from @cmungall, but let me try to answer some of the above :) 

> Not sure if this what you suggested.

I think it should be exact synonym with abbreviation instead of related with abbreviation - I'll make a suggestion in the PR :) 

> returns zero result in both cases.

The abbreviation comes from the allen brain atlas, please see [this link](http://atlas.brain-map.org/atlas?atlas=1&plate=100960516#atlas=1&plate=100960132&resolution=16.75&x=4272&y=4008.875&zoom=-3&structure=154) for it. Given that the allen brain atlas is pretty popular a reference atlas (it's basically only this or paxinos that is used by wet lab scientists in my experience), there is a case for the synonyms here considered as being popular. 

>The uberon build pipeline ensures that no two classes share the same string as either a label or exact synonym. This helps detect common categories of errors.

I think we might have turned this check off for the reason above. If I'm not wrong we still do have a label duplicity check but not a synonym check. There has been discussions on if we should allow duplicate exact synonyms, and I think a good suggestion is to have this check ignore abbreviations and turn it back on.

src/ontology/uberon-edit.obo
```diff
@@ -224043,7 +224043,7 @@ def: "A group of neurons in the floor of the fourth ventricle, in close proximit
 synonym: "nuclei perihypoglossales" EXACT [WikipediaVersioned:Perihypoglossal_nuclei&oldid=821664008]
 synonym: "perihypoglossal complex" EXACT [WikipediaVersioned:Perihypoglossal_nuclei&oldid=821664008]
 synonym: "perihypoglossal nuclear complex" EXACT [WikipediaVersioned:Perihypoglossal_nuclei&oldid=821664008]
-synonym: "PHY" EXACT [MBA:154]
+synonym: "PHY" RELATED ABBREVIATION [MBA:154]
```
shawntanzk, line 224046:
```suggestion
synonym: "PHY" EXACT ABBREVIATION [MBA:154]
```

vasvir: That helps a lot . Thanks.

Shouldn't the other (UBERON:2000438 "parhypural") treated the same way?

vasvir:
Can't commit your suggestion from github interface so I added it to the PR

shawntanzk (approved):

PR #2625, opened 2022-08-19
Removing multiple xref mappings

shawntanzk:
1) remove HBA:9371 - UBERON:0003936

PR #2645, opened 2022-09-06
Update bridge files

anitacaron:
Apply [changes](https://github.com/obophenotype/uberon/pull/2625) done in the mappings of EMAPA, FMA, HBA, MA and EV.


matentzn (approved):
I assume the mappings lost are all intended, this PR looks otherwise unproblematic

Issue #2646, opened 2022-09-06
UBERON:7500033 & 7500034


megbalk:
**Uberon term**
distal-most point of medial condyle of femur
distal-most point of trochlea of humerus

**Reason for deprecation**
not needed, too specific of a term

**Suggested replacement or term to consider**
none

PR #2647, opened 2022-09-06
obsoleting 7500033 and 7500034

megbalk:
issue #2646


megbalk:
@shawntanzk should I assign a different reviewer?

shawntanzk:
oh im sorry - totally slipped my radar, thanks for pinging
```
 :: ERROR: logic-axiom-references-obsolete: UBERON:7500033 has relationship to https://orcid.org/0000-0003-2699-3066
  :: ERROR: logic-axiom-references-obsolete: UBERON:7500034 has relationship to https://orcid.org/0000-0003-2699-3066
```

shawntanzk:
argh it should be an annotation not a logical axiom - not sure why this is happening, might be obo format stuff? @matentzn is that the case?

matentzn:
Is it a logical axiom? This could simply be a a problem with the QC check.

- Quick fix: remove the dc contributor annotations from the obsolete classes
- Correct fix: Fix the failing check in the makefile to allow contributor relationships (I am assuming that `contributor` has `metadata: True` in the OBO file.

megbalk:
I can remove contributor if that helps? Luckily because I have an id range, one could figure out who made the terms in the first place?

shawntanzk:
Yeah for now I would remove the contributor - ill write a ticket to fix the QC check

megbalk:
@shawntanzk fixed, let me know if it looks ok.

shawntanzk:
```
 :: ERROR: logic-axiom-references-obsolete: UBERON:7500034 has relationship to https://orcid.org/0000-0003-2699-3066
```

megbalk:
I wasn't working on 7500034...why is that popping up as an error now?

src/ontology/uberon-edit.obo
```diff
@@ -221904,19 +221904,20 @@ relationship: part_of UBERON:6001767 ! insect antennal disc
...
+is_obsolete: true
+replaced_by: UBERON:0009984
 
 [Term]
 id: UBERON:7500034
-name: distal-most point of trochlea of humerus
-def: "The distal-most point of the trochlea of a humerus." [https://orcid.org/0000-0003-2699-3066]
-intersection_of: UBERON:0006983 ! anatomical point
-intersection_of: distalmost_part_of UBERON:0000144 ! trochlea of humerus
+name: obsolete distal-most point of trochlea of humerus
```
shawntanzk, line 221915: looks like you did obsolete UBERON:7500034

megbalk: oops! ok I'll fix it.

shawntanzk (approved):

cmungall:
checks pass, 1 approval, go ahead and merge!

I think in general we should be more precise about reasons - when is too specific too specific? But as the obsoleter is the original requester let's just merge!

Issue #2651, opened 2022-09-14
opisthokranion


megbalk:
**Preferred term label:**

opisthocranion

**Synonyms**

opisthokranion

**Definition (free text, please give PubMed ID)**

A craniometric point on the nuchodorsal border of the foramen magnum in the median plane.

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**

Subclass of
anatomical point
part of some foramen magnum

rdfs:comment
In pigs this point is projected into space as a point between a line between the aboral points of the right and left superior nuchal lines.

**Your nano-attribution (ORCID)**

PR #2652, opened 2022-09-14
opisthocranion

megbalk:
issue #2651


src/ontology/uberon-edit.obo
```diff
@@ -222532,6 +222532,16 @@ synonym: "sagittal ridge" EXACT []
 is_a: UBERON:4200133 ! crest
 relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
 
+[Term]
+id: UBERON:7500117
+name: opisthocranion
+def: "A craniometric point on the nuchodorsal border of a foramen magnum in the medial plane." [ISBN:9780873659505, Wiktionary:opisthocranion]
```
bvarner-ebi, line 222538: Hi, @megbalk, the wiktionary definition I see is:

"([anatomy](https://en.wiktionary.org/wiki/anatomy)) The [posteriormost](https://en.wiktionary.org/wiki/posteriormost) point in the [midsagittal](https://en.wiktionary.org/wiki/midsagittal) plane of the [occiput](https://en.wiktionary.org/wiki/occiput)"
This definition aligns with the images (of human skulls) I see with a cursory online search.

I am not familiar with the term "nuchodorsal" and could not find a definition online. Even so, this point does not appear to be part of the foramen magnum, at least not in humans.

bvarner-ebi (commented):
See inline comment.

src/ontology/uberon-edit.obo
megbalk: @bvarner-ebi thank you for checking. The definition comes from von den Driesch 1979. I'll find other definitions, then, to make sure that it is more broad and applicable.

megbalk: The definition [here](https://locatelambda.org/manual/) has the opisthion and the opisthocranion, which point is on the skin, not a bone (but paleontologists define it on the skull). 

I will change this to be the opposite side from the glabella on the skull.

I must've gotten this confused with the opisthion, and so would also like to make that term.

bvarner-ebi: Hi, @megbalk,
To consider: add mention that the point exists in the midline since that appears in all the definitions referenced that I can see. 
Please also add dbxref (probably just the URL) that supports this text definition.

src/ontology/uberon-edit.obo
```diff
@@ -222532,6 +222532,25 @@ synonym: "sagittal ridge" EXACT []
 is_a: UBERON:4200133 ! crest
 relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
 
+[Term]
+id: UBERON:7500117
+name: opisthocranion
+def: "A craniometric point which is the furthest point from the glabella." [https://locatelambda.org/manual/, ISBN:9780873659505, Wiktionary:opisthocranion]
+comment: In pigs this point is projected into space as a point between a line connecting the aboral points of the right and left superior nuchal lines. {xref="ISBN:9780873659505"}
```
bvarner-ebi, line 222539: Cannot access the ISBN reference to confirm.
Also, I'm trying to reconcile the comment with the text definition. Is the opisthocranion located on the surface of the skull in all species other than pigs?

src/ontology/uberon-edit.obo
```diff
@@ -222532,6 +222532,25 @@ synonym: "sagittal ridge" EXACT []
...
+synonym: "opisthokranion" EXACT []
```
bvarner-ebi, line 222540: Is there a dbxref to support this?

bvarner-ebi (commented):
See inline comments.

src/ontology/uberon-edit.obo
```diff
@@ -222532,6 +222532,25 @@ synonym: "sagittal ridge" EXACT []
...
+is_a: UBERON:0006983 ! anatomical point
+relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
+relationship: part_of UBERON:0001676 ! occipital bone
```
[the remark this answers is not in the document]
megbalk: It is in the line of the occipital bone, but not on it. Do you recommend omitting the logical definition?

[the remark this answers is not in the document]
bvarner-ebi: After looking at other terms for anatomical points and lines in Uberon, I think it is fair to keep this assuming an  opisthocranion does not exist in the absence of some 'occipital bone'.

bvarner-ebi (approved):
See inline comments.

src/ontology/uberon-edit.obo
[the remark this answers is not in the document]
megbalk: I removed the logical definition

PR #2654, opened 2022-09-15
fixes#2405-semitendinosus_tendon

paulacalipho:
fixes#2405-semitendinosus_tendon


bvarner-ebi (approved):
Approved, with notes below...

Consider adding an equivalent axiom like 
`tendon and 'attached to' some semitendinosus` (and remove redundant SubClass Of axioms)  if appropriate.

Also, I see this is failing QC checks. Unsure why. Maybe because dc:contributor and dbxref for definition are the same? If you can't resolve, kindly reach out to @shawntanzk or @matentzn.

bvarner-ebi:
I also see "MEMORY" referenced in the QC report, so it may just be a case of needing to rerun.

`OWLTOOLS_MEMORY=20G owltools --use-catalog --no-debug basic.owl --list-cycles -f > reports/basic-allcycles`

shawntanzk:
yes, probably a memory issue, I've triggered a re-run - hopefully it will run through this time

paulacalipho:
Thanks @bvarner-ebi : I did as you suggested: added "tendon and 'attached to' some semitendinosus" as "Equivalent to" and removed them from "SubClass of"

PR #2660, opened 2022-09-20
Exclude dc-contibutor from OBO checks

### LINKED ISSUE #2655: QC fails at dc-contributor annotation on obsolete terms
Currently QC seems to be unhappy about dc-contributor annotation on obsolete terms
It seems to think that dc-contributor is a logical axiom - possibly cause of `relationship:` in obo format? 
See https://github.com/obophenotype/uberon/pull/2647

- [ ] Fix the failing check in the makefile to allow contributor relationships (I am assuming that contributor has metadata: True in the OBO file.

anitacaron:
Fix #2655

### DIFF
--- a/src/scripts/check-obo-for-standard-release.pl
+++ b/src/scripts/check-obo-for-standard-release.pl
@@ -165,10 +165,12 @@ while (<>) {
         if ($2 =~ /^$id/) {
         	flag("circular $1 ref", $id);
         }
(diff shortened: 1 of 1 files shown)


matentzn (approved):
I assume you tested it. Looks good to me!

gouttegd (approved):

Issue #2644, opened 2022-09-05
Question on anatomical modelling of 'dorsal root ganglion' and 'spinal nerve root'


bvarner-ebi:
**Context:**

Currently [UBERON:0009623 ‘spinal nerve root’](http://purl.obolibrary.org/obo/UBERON_0009623) and [UBERON:0000044 ‘dorsal root ganglion’](http://purl.obolibrary.org/obo/UBERON_0000044) (DRG) have the axiom `'located in' some 'vertebral canal'.` 

[UBERON:0006692 ‘vertebral canal’ ](http://purl.obolibrary.org/obo/UBERON_0006692)is defined as

```
'anatomical conduit'
 and ('lumen of' some 'vertebral column')
 and ('conduit for' some 'spinal cord')
```
 and

"A conduit enclosed within the vertebral foramen of the vertebrae through which the spinal cord passes."
 
A Netter image of an axial view including these structures:
 
![image](https://user-images.githubusercontent.com/95643797/188483467-0922f8f6-0d15-44e4-ac76-2102a56c7b8a.jpeg)

The nerve roots and DRG appear to extend bilaterally, beyond the boundary of the vertebral foramen, passing laterally through the intervertebral foramina. This seems to be in conflict with the axiom that is associated with both the 'spinal nerve root' and DRG classes: 

`'located in' some 'vertebral canal'`

**Question:**

Is it anatomically correct to have the axiom `'located in' some 'vertebral canal'` for classes [UBERON:0009623 ‘spinal nerve root’](http://purl.obolibrary.org/obo/UBERON_0009623) and [UBERON:0000044 ‘dorsal root ganglion’](http://purl.obolibrary.org/obo/UBERON_0000044)?

bvarner-ebi:
@emquardokus to provide more input with spinal expert. Thanks, @emquardokus!

PR #2665, opened 2022-09-27
Addresses #2644 Question on anatomical modelling of 'dorsal root ganglion' and 'spinal nerve root' 

bvarner-ebi:
Addresses #2644 Question on anatomical modelling of 'dorsal root ganglion' and 'spinal nerve root'


bvarner-ebi:
@shawntanzk, it doesn't look like QC checks have ran / are running. Have they been disabled?

shawntanzk:
> @shawntanzk, it doesn't look like QC checks have ran / are running. Have they been disabled?

It's running - sometimes you have to refresh your page

shawntanzk:
could you put a comment or something on what we discussed, or if you prefer, write it down in a ticket and use seealso and point it to the ticket. Thanks :)

shawntanzk:
oh wait i see theres already a ticket, add discussion there and add seealso pointing there then. Thanks!

bvarner-ebi:
> oh wait i see theres already a ticket, add discussion there and add seealso pointing there then. Thanks!

Is it standard to add a seealso annotation when fixing incorrect modelling?
I understand doing that for controversial edits, but this edit seems to fix a seemingly obvious mistake.

shawntanzk:
its not so much the obvious mistake, its the lack of putting any replacement located_in that would be good to have an explanation for

cmungall (approved):

PR #2668, opened 2022-09-29
2022-09-30 Release

shawntanzk:
Whole lot of date removals are from https://github.com/obophenotype/uberon/pull/2619 - author removed them

Other than that all looks alright to me

Also tagging @anitacaron & @matentzn so you know that I'm doing this release :)


bvarner-ebi (approved):
Every line change was not reviewed, but at a glance I have found several expected changes, and nothing sticks out as alarming. 
However, there are many changes in the uberon-diff.md file that I'm not sure how to reconcile. 

Approving with the assumption that the uberon-diff.md changes are expected.

src/ontology/reports/uberon-edit.obo-obo-report.tsv
```diff
@@ -4849,6 +4849,7 @@ WARN	missing_definition	UBERON:6007046	IAO:0000115
 WARN	missing_definition	UBERON:6007245	IAO:0000115	
 WARN	missing_definition	UBERON:6007288	IAO:0000115	
 WARN	missing_definition	UBERON:6007289	IAO:0000115	
+WARN	missing_definition	UBERON:8440076	IAO:0000115
```
matentzn, line 4852: Lets try to avoid moving forward new terms being added without definitions..

matentzn (approved):
🎉  To be honest I find it scary the release just worked, but it looks great!

src/ontology/reports/uberon-edit.obo-obo-report.tsv
shawntanzk: hmm thats strange, let me check on that

shawntanzk: argh thats my bad, I put the def in the wrong field, I'll fix it

matentzn: No need to rerun a release because of this but great :D

PR #2669, opened 2022-09-30
updating MBA bridge

shawntanzk:
Probably should have merged branch in my bridge repo before release - but in my mind I needed to do a uberon release first (I don't - its just for a report). Anyway updating bridge here :)

### DIFF
--- a/src/ontology/bridge/uberon-bridge-to-mba.obo
+++ b/src/ontology/bridge/uberon-bridge-to-mba.obo
@@ -338,6 +338,13 @@ property_value: IAO:0000589 "crus 1" xsd:string
 property_value: status "Verified" xsd:string {source="https://ror.org/03cpe7c52"}
 
 [Term]
+id: MBA:1057
+intersection_of: UBERON:8440075
+intersection_of: BFO:0000050 NCBITaxon:10090
+property_value: IAO:0000589 "gustatory cortex" xsd:string
+property_value: status "Curated" xsd:string {source="https://orcid.org/0000-0001-7258-9596"}
+
+[Term]
 id: MBA:1059
 relationship: BFO:0000050 NCBITaxon:10090
 relationship: BFO:0000050 UBERON:0015800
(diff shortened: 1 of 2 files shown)


matentzn:
I cant really review this - lack of expertise. Either trust your judgement of ask David!

shawntanzk:
oh @matentzn - didnt mean to ask to review content, actually I don't see nay technical issue that can really go wrong, but yeah it was more towards making sure I didnt mess up anything. Content wise is stuff that has been suggested by a SME - and we will get a re-review. Thanks!

shawntanzk:
I'll just merge actually - given that its to get stuff ready for re-review by SME in allen :)

src/ontology/bridge/uberon-bridge-to-mba.obo
```diff
@@ -934,8 +941,12 @@ property_value: status "Verified" xsd:string {source="https://ror.org/03cpe7c52"
 [Term]
 id: MBA:1079
 intersection_of: UBERON:0002985
+intersection_of: UBERON:8440077
+intersection_of: BFO:0000050 NCBITaxon:10090
 intersection_of: BFO:0000050 NCBITaxon:10090
```
gouttegd, line 946: Why the duplicated `intersection_of` tag? I don’t think it does any harm as the redundant tag seems ignored by the OBO parser, but still, probably better to avoid that?

src/ontology/bridge/uberon-bridge-to-mba.obo
```diff
@@ -1391,45 +1402,45 @@ property_value: status "Verified" xsd:string {source="https://ror.org/03cpe7c52"
 
 [Term]
 id: MBA:12993
-relationship: BFO:0000050 NCBITaxon:10090
-relationship: BFO:0000050 UBERON:0008930
+intersection_of: BFO:0000050 NCBITaxon:10090 {all_some="true"}
```
gouttegd, line 1405: I don’t think the `all_some` qualifier is needed here. It’s not really explicitly said in the OBO Flat File format spec, but `all_some` seems to be the default (that’s how the OWL API treats it anyway). Again, it doesn’t do any harm, but since it’s not needed, probably better avoided?

gouttegd (commented):
Can’t say anything about the mappings themselves, obviously, but technically, I have some minor concerns about the OBO version, which seems some duplicated tags and unneeded qualifiers.

How is the OBO version generated?

gouttegd:
Oh, nevermind, I am too late! :sweat_smile:

src/ontology/bridge/uberon-bridge-to-mba.obo
shawntanzk: yeah notice that, but i figured it doesn't harm - mostly focused on the owl format - the obo is just i conversion using ROBOT

src/ontology/bridge/uberon-bridge-to-mba.obo
shawntanzk: this one is strange - wonder if its me duplicating some equiv stuff with different annotations on it that the obo parser is confused? - A lot of the mappings are manual which = duplications >.< well anyway its ok for now - once I get the final look through by SME, and given the thumbs up, I'll clean stuff up :) Thanks @gouttegd

PR #2678, opened 2022-10-20
Update logical axioms for otic glangion

### LINKED ISSUE #2589: Replace the use of innervates for ganglia with `extends_fibers_into`
* I think we could replace the use of innervates for ganglia with 'extends_fibers_into', preserving use of innervates for neuron projection bundles (tracts and nerves) - although I guess that might lose the sense that those fibers synapse to targets in the object of the relationship.

_Originally posted by @dosumis in https://github.com/obophenotype/uberon/issues/2495#issuecomment-1187364719_

anitacaron:
Summarising the discussion in the related issue, the changes are:
- remove the equivalence class
- add 'otic glanglion' is_a `cranial ganglion`
- add `extends_fibers_into` some 'parotid gland'

Fixes #2589


shawntanzk (approved):
Agree with this change - not sure if we should have a GCI for humans that links otic ganglion to glossopharyngeal nerve? but do not want to hold back this change with that.

bvarner-ebi (approved):

PR #2506, opened 2022-05-30
Polysemy: trochlea of femur. Obsolete offending UBERON:7500077

vasvir:
This PR tries to resolve the 3 of 4 of the issues mentioned in #2424

The problem that is addressed here is that

"trochlea of femur" (UBERON:7500077) is an exact synonym to "intercondylar fossa" (UBERON:4200096). This creates unecessary polysemy.

The proposed solution is to obsolete UBERON:7500077 altogether.

I also moved from the obsoleted term the following:
* synonym: "patellar surface of femur" EXACT []
* is_a: UBERON:4200230 ! surface of bone
* relationship: part_of UBERON:0004406 ! distal epiphysis of femur

Tagged @chandron, @elekka since he helped with the biology.

Tagged @matentzn as requested.

PR #2681, opened 2022-10-21
Check for both Label and Synonym polysemy across Terms

vasvir:
Check for both Label and Synonym polysemy across Terms

This PR started as #2640 but then I tried to outsmart github and I managed to close it.

This PR detects synonym-synonym and label-synonym across different Terms (UBERON Ids).
Another implementation (but with the same logic and functionality) of this check was
used to detect and fix (merged) the followings issues:

* https://github.com/obophenotype/uberon/pull/2465
* https://github.com/obophenotype/uberon/pull/2470
* https://github.com/obophenotype/uberon/pull/2506

The backstory is in https://github.com/obophenotype/uberon/issues/2424

The existing check
https://robot.obolibrary.org/report_queries/duplicate_exact_synonym detects
only synonym-synonym violations. It was reported in
https://github.com/ontodev/robot/issues/607 but closed at the time.

The existing check
https://robot.obolibrary.org/report_queries/duplicate_label_synonym
detects all label-synonym duplication both in the same Term and across
different terms. The duplicity in the same term is a warning but the
duplicity across different terms should be an error.

In this PR the test is case insensitive because downstream applications
have a strong possibility to be case insensitive in order to maximize
coverage.

Finally I have added an exemption for ABBREVIATION per @shawntanzk suggestion for
cases like #2643, #2641.

The errors should look like this. Note that for the example below a fix is already
merged and thus it is not a problem anymore.

entity,property,value
http://purl.obolibrary.org/obo/UBERON_0002633,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,nV
http://purl.obolibrary.org/obo/UBERON_2005093,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,NV

Thanks @matentzn and @anitacaron for the guidance and the hand-holding.

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -39,7 +39,7 @@ REPORT_FAIL_ON =            ERROR
 REPORT_LABEL =              
 REPORT_PROFILE_OPTS =       --profile $(ROBOT_PROFILE)
 OBO_FORMAT_OPTIONS =        
-SPARQL_VALIDATION_CHECKS =  equivalent-classes owldef-self-reference illegal-annotation-property taxon-range orcid-contributor obsolete-replaced_by xrefs-mesh-pattern 
+SPARQL_VALIDATION_CHECKS =  equivalent-classes owldef-self-reference illegal-annotation-property taxon-range orcid-contributor obsolete-replaced_by xrefs-mesh-pattern label-synonym-polysemy 
 SPARQL_EXPORTS =            basic-report 
 ODK_VERSION_MAKEFILE =      v1.3.1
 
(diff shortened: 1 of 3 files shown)


shawntanzk:
@anitacaron could you help me take a final look at this please? it looks good to me, but would like your more skilled eyes on this :) if all good happy for this to be merged

Thanks so much for your contribution @vasvir - really really appreciate it!

anitacaron (approved):
I tested all cases, and it seems it's working.

I've adjusted the query a bit. I moved the dot from where it doesn't need to where it needs to and removed `?iname` from the `order by` because it's not part of the selection.

vasvir:
Great. Thanks for the review and fixes!

Issue #2677, opened 2022-10-20
Update `extends fibers into` definition


anitacaron:
(...)
(b) I think the definition of 'extends fibers into' should avoid cases like this:
"there are also neurones that pass through the otic ganglion that do not synapse in the ganglion"

suggested def:

X extends_fibers into Y iff there exists some neuron(n) n has_soma_location(RO:0002100) x and n 'has synaptic IO in region'(RO:0013001) Y

I think it would be safe to add this property chain as I believe has_soma_location has domain neuron.

has_soma_location o 'has synaptic IO in region' -> 'extends_fibers_into'

_Originally posted by @dosumis in https://github.com/obophenotype/uberon/issues/2589#issuecomment-1256200419_

PR #2685, opened 2022-10-24
add def to extends_fibers_into

shawntanzk:
Fixes https://github.com/obophenotype/uberon/issues/2677


shawntanzk:
PS would like the automated diff to run through to make sure property chain doesn't do anything strange

shawntanzk:
huh, would have expected changes, but seems like its all good :)

anitacaron (approved):

Issue #2667, opened 2022-09-29
Should fissures/sulcus/incisure be material or immaterial?


shawntanzk:
Related to https://github.com/obophenotype/uberon/pull/2633

Currently it is unclear in uberon if fissures/sulcus/incisure should be material or immaterial - do they refer to the fold itself or do they refer to the space 

Currently there are examples of both. Examples: 

immaterial: 
http://purl.obolibrary.org/obo/UBERON_0014466
http://purl.obolibrary.org/obo/UBERON_0000093

material: 
terms under surface groove(http://purl.obolibrary.org/obo/UBERON_0006846): 
http://purl.obolibrary.org/obo/UBERON_0005412
http://purl.obolibrary.org/obo/UBERON_0005478

Need input on this @cmungall & @dosumis

shawntanzk:
From UBERON call: 

In Allen Brain Atlas, it is not "represented" in the atlas - as in there is the term, but not mapped on the atlas. This should (at least in my mind) be immaterial. 
See examples by @patrick-lloyd-ray 

![Screenshot 2022-10-24 at 18 24 41](https://user-images.githubusercontent.com/76212760/197587927-6b5f6464-1429-4ee7-9623-94930222bb5d.png)
![Screenshot 2022-10-24 at 18 24 51](https://user-images.githubusercontent.com/76212760/197587956-6a003aca-9f7e-47ba-a1b4-6116eb069dc6.png)

However, @cmungall did mention that there is utility of being in the material branch as that is more used by people. 

My take: 
We follow atlas and make all the above as immaterial unless we have a use case for making it material 

Plan:
Will make a go and moving them all to immaterial and see if uberon breaks and figure out from there

shawntanzk:
Using the following in ubergraph to find all material terms with sulcus, fissure, incisure in uberon:

<details>
<summary> SPARQL query </summary>

```SPARQL
# adding prefixes used
prefix owl: <http://www.w3.org/2002/07/owl#>
prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>
prefix IAO: <http://purl.obolibrary.org/obo/IAO_>
prefix BFO: <http://purl.obolibrary.org/obo/BFO_>

# selecting only unique instances of the three variables
SELECT DISTINCT ?entity ?label WHERE
{
  # the variable property has to be definition (IAO:0000115)
  VALUES ?property {
    rdfs:label
  }
  
  # only look for uberon terms
[...]

PR #2689, opened 2022-10-25
[DRAFT] make immaterial anatomical entity immaterial

shawntanzk:
immaterial anatomical entity isn't actually immaterial
using this PR to test if I make it immaterial if it will break


shawntanzk:
weird, it seems like on OLS, immaterial anatomical entity is immaterial https://www.ebi.ac.uk/ols/ontologies/uberon/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FUBERON_0000466 though not sure where its coming from? checked uberon-full and the axiom isn't there

shawntanzk:
Closing as this was a test, follow up on https://github.com/obophenotype/uberon/issues/2667

PR #2690, opened 2022-10-25
Disjointing immaterial/material anatomical entity

shawntanzk:
Made immaterial/material anatomical entity a subclass of immaterial/material entity (seems to be in resources but not explicit, and I think its better if we are explicit here)

Made immaterial anatomical entity and material anatomical entity disjoint - I would assume it was done in the level of BFO, but doesnt seem so, so adding it on the uberon level

Related to https://github.com/obophenotype/uberon/issues/2667


gouttegd:
I have personally no objection to enforcing the disjunction between material and immaterial entities.

We have quite a few similar disjointness axioms in FBbt (including a direct equivalent to this one, between FBbt’s `material anatomical entity` and `immaterial anatomical entity`). They are very useful to catch gross classification errors, and overall I think Uberon would benefit from having more of such disjointness axioms.

I am concerned about the potential impact downstream of Uberon, though – especially for those who merge Uberon with several other ontologies. Suddenly introducing a disjointness where there was none before could reveal classification errors that went undetected until now – maybe not in Uberon itself (we seem good on that front: I tried adding the disjointness myself, it does not cause any unsats within Uberon), but in any of the ontologies Uberon may be merged with.

(This can be seen as an argument _against_ the disjointness axiom as well as an argument _in favour_ of it…)

shawntanzk:
> maybe not in Uberon itself (we seem good on that front: I tried adding the disjointness myself, it does not cause any unsats within Uberon)

Not by itself, but with changes I want to make, heaps of unsats, dealing with it now. I think ENVO might be affected by this change btws.

gouttegd (approved):
I am inclined to say “go for it” and if it causes unsats when merging with other ontologies, let’s have the other ontologies fix their inconsistencies. 😈

shawntanzk:
From https://github.com/obophenotype/uberon/pull/2691#issuecomment-1290833392 -> looks like disjointness between material and immaterial does come in later in the process, I am for keeping this, cause it makes it easier to fix at edit stage rather than later on with merges and stuff.

shawntanzk:
OMG finally figured out all the unsats from https://github.com/obophenotype/uberon/pull/2691
I think this is decently safe to merge given all these get inferred down the pipeline anyway, and this just makes it explicit in the edit file (which I think is super useful), but not 100% comfortable merging until either @dosumis or @cmungall gives a thumbs up given its such high level a change

shawntanzk:
> FYI, you can add the approval required label to this PR to explicitly state your intention of having it reviewed and not merging it without proper approval. The label will cause the stale-issues.yml GitHub Action workflow to ignore the PR.

brilliant! will do :) such a well thought out action :D

dosumis (approved):
Looks good to me.  Are the changes to prevent inconsistency in another branch?

shawntanzk:
> Are the changes to prevent inconsistency in another branch?

Not at all, they cause unsats in another branch if anything lol (though not really, cause in the QC process these get placed in), I think this is more of me trying to more easily figure out unsats - atm the material immaterial disjoint comes in a lot later in the process of QC, having this disjointness here would mean we can see it straight in protege when editing (I think) - it has made figuring out unsats stuff a lot easier for me at least.

PR #2693, opened 2022-10-26
Fix various classification issues

gouttegd:
This PR fixes the incorrect classifications of `imaginal disc` (which should be a `multi-tissue structure` and not an `epithelial sac`, since it is made of more than one epithelium) and `primary nerve cord` (which should be a `multicellular anatomical structure` instead of a `central nervous system cell part cluster`, since it is made of _cells_, not of _cell parts_).

Those incorrect classifications have no impact in Uberon itself, but are one cause of unsatisfiable classes when merging Uberon and FBbt.


shawntanzk:
> primary nerve cord (which should be a multicellular anatomical structure instead of a central nervous system cell part cluster, since it is made of cells, not of cell parts).

I don't know if this is right - does the nerve cord contain full cells? I do get that the PMID does reference it as a cluster of neurons, but given nerves are the axons, I assume nerve cord does not contain only full cells too (note: it also does not not contain full cells, eg glial and all). PS I feel like I've had this conversation recently before, but I cant find the ticket >.<

gouttegd:
I have no reference at hand, but the ventral nerve cord in flies definitely contain (among other things) full cells: some interneurons are fully contained (soma + neurites) within the confines of the cord.

It _also_ contains cell parts, sure (such as the soma of neurons whose neurites are outside the cord), but that does not preclude the cord from being better classified as a `multicellular anatomical structure` than as a `multi cell part structure`: a `multicellular anatomical structure` can be composed of both full cells and cell parts, but a `multi cell part structure` is explicitly defined as being only made of cell parts and _no_ complete cells.

(Also, as you note, the text definition of the cord as being a “cluster of neurons” is wrong, as there’s definitely more than neurons in it. I choose to leave that out of this PR as it had no impact on the classification, but that’s something that will need fixing as well.)

shawntanzk:
> ventral nerve cord in flies definitely contain (among other things) full cells

I don't doubt this, this was a conversation I was having before, I think all of them have full cells, if not neurons, probably full glial cells. 

> a multicellular anatomical structure can be composed of both full cells and cell parts

This makes sense

shawntanzk:
> imaginal disc (which should be a multi-tissue structure and not an epithelial sac, since it is made of more than one epithelium) 

btws I'm not an insect expert, but quick reading makes me feel like it still needs to relate to epithelial sac somehow? like I'm trying to understand what you mean by made of more than one epithelium? 

First line of https://www.cell.com/current-biology/pdf/S0960-9822(10)00291-5.pdf

> What is an imaginal disc?
An imaginal disc is a sac-like
epithelial structure found inside
the larva of insects that undergo
metamorphosis.

gouttegd:
> it still needs to relate to epithelial sac somehow?

Not with the way `epithelial sac` is currently defined and classified in Uberon.

An imaginal disc does have a “sac-like” shape, but the two “sides” of the sac are made of two distinct epithelia with different properties: a columnar epithelium on one side, and a squamous epithelium on the other side. So it really is a multi-tissue thing made of several epithelia, not a single epithelium.

But `epithelial sac` in Uberon is defined and classified as being a (single) epithelium. It’s not “a sac made of epithelia”, but rather “an epithelium that happens to have the shape of a sac” – that does not fit what an imaginal disc is.

shawntanzk (approved):
after explanation, all makes sense, thanks :) - ps maybe add some comments where needed?

gouttegd:
(I think it‘s a pretty nice example of why you can’t always trust term labels when you’re looking for a term suitable for your purpose… You can find a term with a label that seems to match exactly what you’re looking for, but that is not actually what you need/want.

I might start collecting examples like that – make a kind of “best-of” of all the classification errors that were caused by reliance on term labels. :D )

PR #2694, opened 2022-10-26
Fix neural glomerulus

gouttegd:
This PR re-classifies `neural glomerulus` to make it `part_of some neuropil` rather than a subclass of `neuropil`, on the basis of both the text definition and how the equivalent FBbt term is classified.

It also remove cross-references that pointed to resources that no longer exist, and the `depicted by` annotation which pointed to an image depicting a _renal_ glomerulus.


src/ontology/uberon-edit.obo
```diff
@@ -11741,13 +11741,12 @@ property_value: homology_notes "The bilaterian gut is typically a complete tube
 [Term]
 id: UBERON:0001047
 name: neural glomerulus
-def: "Large synaptic bouton found in areas like olfactory bulb, cerebellar cortex and thalamus, that contacts many post-synaptic targets[NIF]." [http://ccdb.ucsd.edu/SAO/images/glomerulus_sw_512x512.jpg, http://flybrain.uni-freiburg.de/Flybrain/html/terms/terms.html]
-comment: isa or partof neuropil? TODO. Add to GO-CC?
+def: "Large synaptic bouton found in areas like olfactory bulb, cerebellar cortex and thalamus, that contacts many post-synaptic targets[NIF]." [DOI:10.1016/B978-0-12-374245-2.00007-3]
```
shawntanzk, line 11744:
```suggestion
def: "A large synaptic bouton found in areas like olfactory bulb, cerebellar cortex and thalamus, that contacts many post-synaptic targets." [DOI:10.1016/B978-0-12-374245-2.00007-3, ILX:0728797]
```

src/ontology/uberon-edit.obo
```diff
@@ -11741,13 +11741,12 @@ property_value: homology_notes "The bilaterian gut is typically a complete tube
...
 subset: uberon_slim
 synonym: "glomerulus" BROAD [FBbt:00005386, NIF_Subcellular:sao587858733]
 xref: NIF_Subcellular:sao587858733
```
shawntanzk, line 11747:
```suggestion
xref: ILX:0728797
```

shawntanzk: NIF_Subcellular xref doesnt resolve, and NIF_Subcellular prefixes eventually resolves to ILX anyway, so this is probably more appropriate an xref regardless

shawntanzk (approved):
Agree with the change, see inline for some minor fixes to xrefs 
Will approve now, logical changes are sensible and all, but would like those xref changes made :)

Issue #2491, opened 2022-05-24
Anatomical cluster seems misplaced


sebastianduesing:
_Anatomical cluster_ is defined as a type of anatomical group (specifically, "Anatomical group that has its parts adjacent to one another."). However, it is not a subclass of _anatomical group_; rather, _anatomical group_ is a subclass of _anatomical structure_, which is a subclass of _material anatomical entity_, which in turn is a sibling class of _anatomical cluster_. Either _anatomical cluster_ should be moved to be a subclass of _anatomical group_, or its definition should be changed so that it isn't defined as a type of anatomical group while not actually being one.

PR #2695, opened 2022-10-26
Fix classification for disconnected entities

### LINKED ISSUE #2682: Inconsistent distinction between “connected” and “non-connected” anatomical entities
Uberon is inconsistent in the way it distinguishes between anatomical entities that are made of parts that are all connected to each other (hereafter “connected entities”) and entities that are made of parts that are not necessarily all connected (hereafter “non-connected entities”).

This does not cause any obvious problems within Uberon itself, only because the distinction between these two kinds of entities is not strictly enforced. This does, however, cause problem when Uberon is merged with FBbt (as happens e.g. in Ubergraph), because FBbt does enforce a strict separation between connected and non-connected entities.

In FBbt, we have:
* `anatomical structure` for anatomical entities made of connected parts, and
* `disconnected anatomical group` for anatomical entities made of non-connected parts (this used to be called `anatomical group` – it was recently renamed to add the “disconnected” qualifier to make it more apparent that it is a non-connected entity).

Many of FBbt’s classes are below one of those two classes. We also have a disjointness axiom between them, so that nothing can be simultaneously made of parts that are all connected and made of parts that are not all connected.

Uberon is superficially similar to FBbt in that it also has:
* `anatomical structure` for entities made of connected parts;
* `disconnected anatomical group` for entities made of non-connected parts.

However, several terms are classified under `anatomical structure` even though they clearly refer to entities that are _not_ made of connected parts. Many of those terms are mapped to equivalent FBbt terms, which then results in unsatisfiable classes upon merging Uberon and FBbt.

The main offenders are:
* `non-connected functional system`: the label itself clearly indicates this is a non-connected entity, however this is classified as a `anatomical system`, which is itself a `anatomical structure` (so a connected entity);
* `insect non-connected developing system`: again, that’s clearly a non-connected entity, but is classified as a `anatomical group`, which is a `anatomical structure` (so a connected entity);
* `insect synaptic neuropil block`: defined as a “closely associated group of synaptic neuropil domains”, suggesting it should be a non-connected entity (the equivalent FBbt term is indeed a non-connected entity), but classified as a `anatomical group` (so a connected entity);
* `insect synaptic neuropil`: classified in Uberon as a `anatomical group` (so a connected entity), but the corresponding FBbt term is classified as a more generic `anatomical entity`, which states nothing about the “connectedness” of the entity (we have subclasses for both connected and non-connected neuropils);
* `anatomical cluster`: textually defined as a “anatomical group that has its parts adjacents to one another”, but classified as a `anatomical entity` which states nothing about the “connectedness”; the equivalent FBbt term is classified as a `disconnected anatomical group`, which I believe is correct (if we accept that “adjacents” does not imply “connected”);
* `sensory system`: defined as the “anatomical system that overlaps the nervous system and is responsible for receiving and processing sensory informations”, which does not say a lot about the connectedness, but has exact synonym “set of sense organs”, which to me suggests it should be a disconnected entity (the various sense organs are not all connected to each other) – but `anatomical system` is a `anatomical structure`, which makes `sensory system` a connected entity;
* `endocrine system`: exactly the same situation as for `sensory system`.

Proposed actions:
- re-classify `non-connected functional system` as a `disconnected anatomical group`, to make it a non-connected entity as it should be;
- likewise for `insect non-connected developing system` and `insect synaptic neuropil block`;
- re-classify `insect synaptic neuropil` to `material anatomical entity`, so that the classification does not imply either a “connected” or a “non-connected” status (as in FBbt);
- re-classify `anatomical cluster` as a `disconnected anatomical group`;
- possibly obsolete `anatomical group`, which is currently only used as a superclass of `insect non-connected developing system`, `insect synaptic neuropil block`, and `insect synaptic neuropil` – and therefore would end up with no children once those terms are re-classified as proposed;
- re-classify `sensory system` and `endocrine system` as `non-connected functional system`.

Of note, there are probably other subclasses of `anatomical system` that should probably also be re-classified as `non-connected functional system`, here I only listed those that cause problems upon merging Uberon and FBbt. More generally, `anatomical system` is defined as a “_connected_ anatomical structure that has multiple organs as parts”. I am not sure all the systems that are currently under `anatomical system` are all made of _connected_ parts – at the very least for example, `immune system` should probably be a `non-connected functional system`.

gouttegd:
This PR re-classifies several terms to fix inconsistencies regarding the “connectedness” of anatomical entities.

We distinguish between anatomical entities that are made of parts that are all connected (`anatomical structure`) and entities that are made of parts that are not all connected (`disconnected anatomical group`). But we do not enforce the distinction, and as a result we have terms referring to disconnected entities that are classified as (connected) `anatomical structure`.

This cause massive satisfiability problems when merging Uberon and FBbt, because FBbt does enforce a strict separation (with an explicit disjointness axiom) between `anatomical structure` and `disconnected anatomical group`.

This PR implements the fixes proposed in #2682:

* classify `non-connected functional system` as a `disconnected anatomical group`;
* classify `anatomical cluster` as a `disconnected anatomical group`;
* classify `insect synaptic neuropil block` as a `disconnected anatomical group`;
* classify `insect synaptic neuropil` as a `material anatomical entity`;
* classify `endocrine system` and `sensory system` as a `non-connected functional system`.

In addition (not discussed in #2682):

* classify `immune system` as a `non-connected functional system` (on the same rationale as for `endocrine system` and `sensory system`);
* classify `multi cell part structure` as an `anatomical structure`.

closes #2682

### DIFF
--- a/src/ontology/bridge/uberon-bridge-to-mba.obo
+++ b/src/ontology/bridge/uberon-bridge-to-mba.obo
@@ -967,14 +967,17 @@ property_value: IAO:0000589 "efferent cochleovestibular bundle" xsd:string
 property_value: status "Verified" xsd:string {source="https://ror.org/03cpe7c52"}
 
 [Term]
-id: MBA:1079
-intersection_of: UBERON:0002985
+id: MBA:1078
 intersection_of: UBERON:8440077
 intersection_of: BFO:0000050 NCBITaxon:10090
-intersection_of: BFO:0000050 NCBITaxon:10090
-property_value: IAO:0000589 "medial geniculate complex, ventral part" xsd:string
 property_value: IAO:0000589 "rhinal incisure" xsd:string
 property_value: status "Curated" xsd:string {source="https://orcid.org/0000-0001-7258-9596"}
+
+[Term]
+id: MBA:1079
+intersection_of: UBERON:0002985
+intersection_of: BFO:0000050 NCBITaxon:10090
+property_value: IAO:0000589 "medial geniculate complex, ventral part" xsd:string
 property_value: status "Verified" xsd:string {source="https://ror.org/03cpe7c52"}
 
 [Term]
@@ -5826,6 +5829,13 @@ relationship: BFO:0000050 NCBITaxon:10090
 relationship: BFO:0000050 UBERON:0001393
 property_value: IAO:0000589 "posterior auditory area, layer 4" xsd:string
 property_value: status "Verified" xsd:string {source="https://ror.org/03cpe7c52"}
+
+[Term]
+id: MBA:76
+intersection_of: UBERON:8440074
+intersection_of: BFO:0000050 NCBITaxon:10090
+property_value: IAO:0000589 "interstitial nucleus of the vestibular nerve" xsd:string
+property_value: status "Curated" xsd:string {source="https://orcid.org/0000-0001-7258-9596"}
 
 [Term]
 id: MBA:763
--- a/src/ontology/bridge/uberon-bridge-to-mba.owl
+++ b/src/ontology/bridge/uberon-bridge-to-mba.owl
@@ -4268,6 +4268,32 @@
     
 
 
+    <!-- http://purl.obolibrary.org/obo/MBA_1078 -->
+
+    <owl:Class rdf:about="http://purl.obolibrary.org/obo/MBA_1078">
+        <owl:equivalentClass>
+            <owl:Class>
+                <owl:intersectionOf rdf:parseType="Collection">
+                    <rdf:Description rdf:about="http://purl.obolibrary.org/obo/UBERON_8440077"/>
+                    <owl:Restriction>
+                        <owl:onProperty rdf:resource="http://purl.obolibrary.org/obo/BFO_0000050"/>
+                        <owl:someValuesFrom rdf:resource="http://purl.obolibrary.org/obo/NCBITaxon_10090"/>
+                    </owl:Restriction>
+                </owl:intersectionOf>
+            </owl:Class>
+        </owl:equivalentClass>
+        <obo:IAO_0000589 rdf:datatype="http://www.w3.org/2001/XMLSchema#string">rhinal incisure</obo:IAO_0000589>
+        <oboInOwl:status rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Curated</oboInOwl:status>
+    </owl:Class>
+    <owl:Axiom>
+        <owl:annotatedSource rdf:resource="http://purl.obolibrary.org/obo/MBA_1078"/>
+        <owl:annotatedProperty rdf:resource="http://www.geneontology.org/formats/oboInOwl#status"/>
+        <owl:annotatedTarget rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Curated</owl:annotatedTarget>
+        <oboInOwl:source rdf:datatype="http://www.w3.org/2001/XMLSchema#string">https://orcid.org/0000-0001-7258-9596</oboInOwl:source>
+    </owl:Axiom>
+    
+
+
     <!-- http://purl.obolibrary.org/obo/MBA_1079 -->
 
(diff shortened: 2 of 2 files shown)


gouttegd:
The CI failure check is caused by the unsatisfiability of `MBA:1079`, which is defined in the `uberon-bridge-to-mba` bridge as being simultaneously equivalent to `rhinal incisure` (an `anatomical space`, so an immaterial entity) and to `ventral nucleus of medial geniculate body` (an `nucleus of brain`, so ultimately a `multi cell part structure`):

```


`multi cell part structure` used to be a `anatomical entity` (with no statement as to its “materialness”), but with this PR is now a `anatomical structure`, so a material entity.

Therefore `MBA:1079` is now both an immaterial entity and an immaterial entity, hence the unsatisfiability.

@shawntanzk You’re the one who last updated the MBA bridge, could you have a look? It does not seem right that this `MBA:1079` should be both a `rhinal incisure` and a `ventral nucleus of medial geniculate body` (whatever that is ^^ )…

shawntanzk:
Yeah I'll look into it, nothing should have two equivalent like that with the new bridge. Will see if I can do checks on our side too. Thanks for bringing this up :)

gouttegd:
I note that `rhinal incisure` is mapped (has a xref) to `MBA:1078`, while `ventral nucleus of medial geniculate body` has a xref to `MBA:1079`.

I suspect some kind of glitch where the bridge generation process produced an axiom bridging `MBA:1079` both to its correct mapped term _and_ to the mapped term of `MBA:1078`, while no bridging axiom were produced for `MBA:1078`.

shawntanzk:
yeah sorry this is my fault, I made a mistake on curation - will fix :) I'll try to work with huseyin to get a check in on our side next week so this wont happen again

gouttegd:
I’d say this proves you were right about wanting a disjointness axiom between material and immaterial entities – it’s that axiom that caught the problem here. :)

shawntanzk:
strange - this disjointness should have come in later in the QC process anyway. But I guess it's good that this is in earlier - might be one of those weird things where it somehow slips through the cracks of the QC and having disjointness early is better :)

gouttegd:
So the CI check is _still_ failing because it merges this PR with the master branch on commit ae8bf1243b4ca26920320cddf93190b38d8d4f7c, that is, as it was _before_ the fix in #2701 was merged, even though it was merged more than one hour ago already.

I guess there’s must be some kind of delay before a change in master is taken into account… I’ll investigate that later.

shawntanzk (approved):
I mean while the classifications aren't wrong, I think as @cmungall mentioned in the ticket, to remove the distinction between “connected” and “non-connected” entities entirely is sensible, cause non of the previous classifications should be wrong too. I'm guessing there's a specific reason (was it ubergraph) that you wanted this through, so I won't hold it back, but given the high level terms involved, I'd be more comfortable if @cmungall also signed off on this

gouttegd:
I have no objection on removing the distinction between between “connected” and “non-connected”, but then there wouldn’t be any point anymore in having both `anatomical structure` and `disconnected anatomical group` in Uberon. Removing the distinction would mean either removing the two classes and making everything simply a `anatomical entity`, or keeping only `anatomical structure` (with a definition that does not imply the components are connected) and moving everything under `disconnected anatomical group` to `anatomical structure`. Why not, but that’s a pretty big change, which could take some time to agree on and to implement (and could have consequences way beyond Uberon and FBbt, if, as I suspect, a similar distinction currently also exists in other taxon-specific ontologies).

In the meantime, we do have this distinction in Uberon (though it’s not enforced by means of a disjointness axiom, as it is on FBbt), so while it is there I think we should at least make sure our various classes are correctly classified under one of the two categories. Currently, several classes are _not_ correctly classified, and this is a direct cause of problems in Ubergraph.

In essence, what I propose is to fix the existing classifications under this distinction for now, and _then_ to worry about whether the distinction should even exist at all.

anitacaron:
Related to #2491

gouttegd:
It’s been one month, so I am planning to merge this by the end of this week, barring any objection.

PR #2702, opened 2022-10-31
fixes#2400-NTR-fish-terms

### LINKED ISSUE #2400: NTR: fish terms
**Comment:** The following terms are needed to indicate the origin of the tissues used establish cell lines annotated in Cellosaurus. Many thanks

**Preferred term label:** meningeal myeloid tissue
**Synonyms** hemopoietic focus
**Definition** The meningeal myeloid tissue is bone marrow-like (myeloid) structure typical of sturgeon. It is mainly granulopoietic, but it also contains lymphoid elements. [PMID: 33066257; PMID: 3765338; DOI:10.1002/ar.1090880307]
**Parent term** hematopoietic tissue UBERON:0012429
**Your nano-attribution (ORCID)** 0000-0002-0819-0473

**Preferred term label:** milt
**Definition** Milt is the seminal fluid of fish, mollusks, and certain other water-dwelling animals which reproduce by spraying this fluid, which contains the sperm, onto roe (fish eggs). [https://en.wikipedia.org/wiki/Milt]
**Parent term** male reproductive gland secretion UBERON:0006536
**Your nano-attribution (ORCID)** 0000-0002-0819-0473

**Preferred term label:** ovarian fluid
**Definition** Secretion product of the ovaries released together with the eggs. [https://hal.archives-ouvertes.fr/hal-00899767/document]
**Parent term** egg cell CL:0000025
**Your nano-attribution (ORCID)** 0000-0002-0819-0473

paulacalipho:
fixes #2400 NTR-fish-terms


shawntanzk:
don't think I can help with content, I know nothing about fish >.< but this is something I noticed: 

> [UBERON_8480039](http://purl.obolibrary.org/obo/UBERON_8480039) SubClassOf [in taxon](http://purl.obolibrary.org/obo/RO_0002162) some [Actinopterygii](http://purl.obolibrary.org/obo/NCBITaxon_7898)

>[UBERON_8480039](http://purl.obolibrary.org/obo/UBERON_8480039) SubClassOf [in taxon](http://purl.obolibrary.org/obo/RO_0002162) some [Mollusca](http://purl.obolibrary.org/obo/NCBITaxon_6447)

There probably shouldn't be two in taxon -> I think this either should use a higher up the taxon tree or a `present in taxon` annotation instead (see taxon guide here if you need help with adding taxon restriction stuff https://github.com/OBOAcademy/obook/blob/8496fe87df5ed0640d4467163c1650b1c3d78817/docs/explanation/taxon-constraints-explainer.md ; ps sorry its not fully in obook yet, in PR stage atm)

cmungall:
good catch @shawntanzk - but why didn't the pipeline catch this?

shawntanzk:
> good catch @shawntanzk - but why didn't the pipeline catch this?

Is there suppose to be a check for this, or is this something we need to look into? or is it something that certain disjointness should have pinged and we have to go figure out why the pipeline did not disjoint them?

balhoff:
@shawntanzk this class should have been found unsatisfiable by the reasoner. It means taxon constraints are not set up correctly. Four things are needed:
- taxon constraints in ontology
- robot expansions (only for never_in_taxon)
- import of taxonomy hierarchy
- import of taxonomy in_taxon sibling disjointness

shawntanzk:
thanks @balhoff - I will write a ticket and hopefully we can get to the bottom of if with the tech support team :)

src/ontology/uberon-edit.obo
```diff
@@ -224906,6 +224906,35 @@ intersection_of: part_of UBERON:0003889 ! fallopian tube
 relationship: dc-contributor http://orcid.org/0000-0002-0819-0473
 property_value: dcterms-date 2022-10-21T08:55:11Z xsd:dateTime
 
+[Term]
+id: UBERON:8480038
+name: meningeal myeloid tissue
+def: "Lymphomyeloid tissue that consist mainly in granulopoietic cells but that also contains erythroid and lymphoid cells in different stages of maturation." [PMID:25130144]
```
cmungall, line 224912: Thank you for providing a reference for your definition!

In uberon and other OBO ontologies, we prefer sources that are more focused around the definition itself, rather that experimental results papers that might embed the definition and give another sources.

Here we have:

_The meningeal tissue is a unique hemopoietic site in sturgeon
located over the medulla oblongata between the brain and cartilaginous
skull capsule, in a saddle-like manner [26,35]. The
meningeal tissue is mainly granulopoietic, with erythroid and
lymphoid cell lines in different stages of maturation and resembles
the hemopoietic bone marrow in mammals, histologically [26]._

F€ange R. Lymphoid organs in sturgeons (Acipenseridae). Vet Immunol
Immunopathol 1986;12:153e61.

Scharrer E. The histology of the meningeal myeloid tissue in the Ganoids Amia
and Lepisosteus. Anat Rec 1944;88:291e310.

I might prefer to have one of the two cited papers above as the cited reference

src/ontology/uberon-edit.obo
```diff
@@ -224906,6 +224906,35 @@ intersection_of: part_of UBERON:0003889 ! fallopian tube
...
+property_value: curator_notes "May be present in Amia and Lepisosteus, Polyodon and Acipenser, but not Polypterus." xsd:string {xref="https://doi.org/10.1111/j.1096-3642.1998.tb02524.x"}
+property_value: dcterms-date 2022-10-31T07:50:56Z xsd:dateTime
+
+[Term]
+id: UBERON:8480039
+name: milt
+def: "Seminal fluid of fish, mollusks, and certain other water-dwelling animals which reproduce by spraying this fluid, which contains the sperm, onto the egg cells." [https://en.wikipedia.org/wiki/Milt]
+is_a: UBERON:0006536 ! male reproductive gland secretion
+relationship: dc-contributor http://orcid.org/0000-0002-0819-0473
+relationship: in_taxon NCBITaxon:6447 ! Mollusca
```
cmungall, line 224925: as mentioned separately, this should have been caught by our pipeline and the PR flagged. Nothing can be only in molluscs and fish.

I wonder if we want to split this into two concepts?

src/ontology/uberon-edit.obo
```diff
@@ -224906,6 +224906,35 @@ intersection_of: part_of UBERON:0003889 ! fallopian tube
...
+is_a: UBERON:0034769 ! lymphomyeloid tissue
+relationship: dc-contributor http://orcid.org/0000-0002-0819-0473
+relationship: in_taxon NCBITaxon:7898 ! Actinopterygii
+property_value: curator_notes "May be present in Amia and Lepisosteus, Polyodon and Acipenser, but not Polypterus." xsd:string {xref="https://doi.org/10.1111/j.1096-3642.1998.tb02524.x"}
+property_value: dcterms-date 2022-10-31T07:50:56Z xsd:dateTime
+
+[Term]
+id: UBERON:8480039
+name: milt
+def: "Seminal fluid of fish, mollusks, and certain other water-dwelling animals which reproduce by spraying this fluid, which contains the sperm, onto the egg cells." [https://en.wikipedia.org/wiki/Milt]
```
cmungall, line 224922: We like to have a genus-differentia definition, where the genus is an existing term.

Here we might write as

"Seminal fluid which <differentiating feature>"

Where the genus is an is-a parent or ancestor

Here we run into a challenging issue. We have "seminal fluid" in uberon already, you opted not to use it in your is-a, likely wisely, as the existing term may be vertebrate specific (not totally clear). But we should be consistent in terminology and use the same term in your is-a that is in your definition

src/ontology/uberon-edit.obo
```diff
@@ -224906,6 +224906,35 @@ intersection_of: part_of UBERON:0003889 ! fallopian tube
...
+is_a: UBERON:0006536 ! male reproductive gland secretion
+relationship: dc-contributor http://orcid.org/0000-0002-0819-0473
+relationship: in_taxon NCBITaxon:6447 ! Mollusca
+relationship: in_taxon NCBITaxon:7898 ! Actinopterygii
+property_value: dcterms-date 2022-10-31T07:54:42Z xsd:dateTime
+
+[Term]
+id: UBERON:8480040
+name: ovarian fluid
+def: "Fish ovary secretion released with eggs." [https://hal.archives-ouvertes.fr/hal-00899767/document]
```
cmungall, line 224932: note on definition source: We always prefer IDs like PMIDs or URLs of primary literature databases rather than secondary archives.

In this case, the relevant URL is
https://pubmed.ncbi.nlm.nih.gov/8526977/

or simply PMID:8526977

cmungall: is ovarian fluid only secreted in fish?

src/ontology/uberon-edit.obo
cmungall: Wikipedia is sometimes a decent source proxy but in this case the wikipedia page is not really helpful and is more about culinary aspects.

To what extend is a new concept really required here vs just having a synonym on seminal fluid?

Having terms that group across phyla is often a sign we want two different concepts, or it's simply different nomenclature on different species for a more general concept.

cc @ddooley

cmungall (approved):

src/ontology/uberon-edit.obo
paulacalipho: @cmungall I let you decide what is more correct:
1) add as synonym of seminal fluid. 
Is possible to indicate that is a synonym/term used in fish?
2) create a new concept. 
Is this def. better?
"Secretion from the glands of the male genital tract containing sperm that is sprayed onto the egg cells in certain water-dwelling animals."
Difficult to find a good source of definition. What about PMID:31287811 ?

src/ontology/uberon-edit.obo
paulacalipho: @cmungall I do not know if ovarian fluid is only secreted in fish
@cmungall I will make the change about the reference

src/ontology/uberon-edit.obo
paulacalipho: @cmungall I found mainly references about fish seminal fluid. Should I only use 'in taxon' some Actinopterygii and give it up Mollusca?

src/ontology/uberon-edit.obo
paulacalipho: @cmungall OK, I will change to one of these

src/ontology/uberon-edit.obo
paulacalipho: @cmungall I found that this term is also used in _Petromyzon marinus_ (PMID:31287811), so I think the best is to add "milt" as synonym of seminal fluid

src/ontology/uberon-edit.obo
cmungall: Agreed

cmungall (approved):

PR #2705, opened 2022-10-31
Remove OpenCyc xrefs

### LINKED ISSUE #2434: Remove OpenCyc xrefs
As far as I can tell, OpenCyc has not existed for over a decade. Its identifiers look like some kind of hash, but are very confusing and not actionable. I think it would reduce complexity and confusion if these could be removed.

anitacaron:
Fixes #2434


shawntanzk (approved):
I wonder if they should be replaced with anything? I mean either way they should be removed, but yeah... @cthoyt any ideas?

cthoyt:
I'm not aware of any follow-ups to OpenCyc. Best to focus on connecting UBERON to modern resources IMO

PR #2706, opened 2022-10-31
Add date check

shawntanzk:
testing here before ODK


shawntanzk:
Ok not quite sure why this is passing, looking at this in ubergraph: https://api.triplydb.com/s/L3LIEli4Q 
It should be failing cause there are stuff that comes out right? Am I missing something @anitacaron?

shawntanzk:
My mistake, forgot to remove the defined by line used in ubergraph - this should work and ping up error

shawntanzk:
added to next version of ODK https://github.com/INCATools/ontology-development-kit/pull/727

PR #2721, opened 2022-11-22
fixes#2401-arthropod-terms

paulacalipho:
fixes#2401-arthropod-terms


src/ontology/uberon-edit.obo
```diff
@@ -224361,6 +224375,18 @@ relationship: dc-contributor http://orcid.org/0000-0002-0819-0473
 relationship: in_taxon NCBITaxon:7898 ! Actinopterygii
 property_value: dcterms-date 2022-10-31T07:56:47Z xsd:dateTime
 
+[Term]
+id: UBERON:8480041
+name: testicular sheath
+def: "Bilayer sheath that covers both the insect male mature gonad and a portion of the genital disc. It is composed of an outer pigment cell layer, an inner muscle cell layer, and their respective basal laminae." [PMID:9553045]
```
gouttegd, line 224381: I am not sure it is exact that the sheath covers “a portion of the genital disc”. The Kozopas et al. (1998) reference says “covers both the mature gonad (the testis) and a portion of the genital disc (the seminal vesicle)”, but while the seminal vesicle develops from the genital disc, it’s not part of the genital disc. Here, the authors probably meant “a portion of genital disc-derived structures (the seminal vesicle)”.

src/ontology/uberon-edit.obo
```diff
@@ -224361,6 +224375,18 @@ relationship: dc-contributor http://orcid.org/0000-0002-0819-0473
...
+synonym: "testis sheath" EXACT [] {http://purl.obolibrary.org/obo/IAO_0000232="https://www.fortunejournals.com/ijpaes/admin/php/uploads/798_pdf.pdf"}
+is_a: UBERON:0004923 ! organ component layer
+relationship: dc-contributor http://orcid.org/0000-0002-0819-0473
+relationship: in_taxon NCBITaxon:50557 ! Insecta
+relationship: part_of UBERON:0003135 ! male reproductive organ
```
gouttegd, line 224386: This could be made more specific, part of some `testis`.

src/ontology/uberon-edit.obo
```diff
@@ -224361,6 +224375,18 @@ relationship: dc-contributor http://orcid.org/0000-0002-0819-0473
...
+relationship: part_of UBERON:6001785 ! insect male genital disc
```
gouttegd, line 224387: See my comment about the definition above. I don’t think it is correct to make the sheath a part of the genital disc. It is partially derived (at least for the muscular component) from the disc and it may cover some structures that originate from the disc (such as the seminal vesicle), but it is not itself part of the disc.

You could envision making it part of `seminal vesicle`, but our current definition for that term is a little bit too vertebrate-specific. For now I would recommend making it solely part of `testis`.

gouttegd (changes requested):

src/ontology/uberon-edit.obo
paulacalipho: @gouttegd thanks for reviewing. I modified the definition according to your comment. If you think is best to remove this part of the sentence let me know

src/ontology/uberon-edit.obo
paulacalipho: @gouttegd I made it only part of testis

gouttegd (approved):
Looks all good to me. 👍

Issue #2729, opened 2022-12-02
Change label for UBERON:0000946 'cardial valve' to 'cardiac valve'


bvarner-ebi:
The label for UBERON:0000946 is currently 'cardial valve'. As the text definition seems to describe what is more commonly referred to as a "cardiac valve", consider making that the label.

Indeed, the dbxref (PMID:7926784) only refers to "cardiac valve"; the term "cardial valve" is not found in that paper.

"cardiac valve" is currently a related synonym. This should be removed if updating the label to "cardiac valve". If a reference can be found that supports it, "cardial valve" can be added as a synonym.

PR #2730, opened 2022-12-05
Addresses #2729 change cardial to cardiac

bvarner-ebi:
Change label for UBERON:0000946 'cardial valve' to 'cardiac valve' #2729


dosumis (approved):
Looks good to me.  There is an unrelated term 'adult cardial valve' in FBbt, but this does not seem to have been a source for the Uberon term, which groups various heart valves.

PR #2701, opened 2022-10-27
Fix MBA bridge miscuration

shawntanzk:
@gouttegd - this PR should fix the Rhinal Incisure miscuration
Sorry about that >.< 
Will let QC run through and all, in course all day, so please do just merge this in when you want :) 
Related to https://github.com/obophenotype/uberon/pull/2695#issuecomment-1292500068

PR #2732, opened 2022-12-07
Update MBA bridges

shawntanzk:
Verified MBA bridges and this should be a stable version for now.

### DIFF
--- a/src/ontology/bridge/uberon-bridge-to-mba.obo
+++ b/src/ontology/bridge/uberon-bridge-to-mba.obo
@@ -967,15 +967,18 @@ property_value: IAO:0000589 "efferent cochleovestibular bundle" xsd:string
 property_value: status "Verified" xsd:string {source="https://ror.org/03cpe7c52"}
 
 [Term]
-id: MBA:1079
-intersection_of: UBERON:0002985
+id: MBA:1078
 intersection_of: UBERON:8440077
 intersection_of: BFO:0000050 NCBITaxon:10090
+property_value: IAO:0000589 "rhinal incisure" xsd:string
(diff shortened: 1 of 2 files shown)


src/ontology/bridge/uberon-bridge-to-mba.obo
```diff
@@ -966,19 +966,16 @@ relationship: BFO:0000050 UBERON:0001648
...
-intersection_of: UBERON:8440077
-intersection_of: BFO:0000050 NCBITaxon:10090
-property_value: IAO:0000589 "rhinal incisure" xsd:string
-property_value: status "Curated" xsd:string {source="https://orcid.org/0000-0001-7258-9596"}
-
 [Term]
 id: MBA:1079
 intersection_of: UBERON:0002985
+intersection_of: UBERON:8440077
+intersection_of: BFO:0000050 NCBITaxon:10090
```
gouttegd, line 973: I believe this is a regression to the same bug that was spotted in [PR#2695](https://github.com/obophenotype/uberon/pull/2695#issuecomment-1292500068) and fixed in #2701 ...

gouttegd (changes requested):

src/ontology/bridge/uberon-bridge-to-mba.obo
shawntanzk: argh, sorry about that, I thought I fixed it in the other repo... I'll look into it. Thanks :)

shawntanzk: oh I know what happened, copied over verified file which happened after fix, so needed to re-fix it >.< will make the changes

gouttegd (approved):

PR #2733, opened 2022-12-08
Add upper and lower layers of the cortex

### LINKED ISSUE #2722: Upper and Lower Cortex
It would be useful to have upper and deep layers as defined by https://www.biorxiv.org/content/10.1101/2022.10.12.511898v1.full.pdf
```
Although cortical layers two through four are
commonly considered “upper” and layers five and six
“deep,” both upper-layer IT and deep-layer IT superclusters 
contained putative Layer-4 clusters 
```
Axiomatization with layer 4 might be tricky? 
Perhaps doing it like mix cortical layers that do not have logical axioms would make sense

shawntanzk:
#gogoeditdiff
Not my favourite definitions and would like feedback 
Fixes #2722


patrick-lloyd-ray (approved):
I think the definitions are fine -- the "surface"/"deep" modifiers may not be the most accurate but they are intuitive and express what is meant by the terms well.

Issue #2723, opened 2022-11-28
Remove property chain injection


anitacaron:
Action item:

Add back the property chain for now and create a ticket to remove it when we figure it out

@anitacaron will open another ticket to remove the injection when things are fully decided

_Originally posted by @shawntanzk in https://github.com/obophenotype/uberon/issues/2664#issuecomment-1297157894_

The property chain was added in the `components/develops-from-chains.owl`

matentzn:
- [ ] Also lets remember to remove the import in edit, if any

cmungall:
Can we see the reasoned diff?

Presumably has-developmental-contribution-from entailments will be unchanged and the classification under probe/canary classes will be unchanged?
https://github.com/obophenotype/uberon/issues/2664#issuecomment-1271631024

anitacaron:
It's in the new PR https://github.com/obophenotype/uberon/pull/3197#issuecomment-1927019687

PR #2735, opened 2022-12-09
Inject property chain in develops from

anitacaron:


matentzn (approved):
This PR is introducing a non-change. Which is needed for backwards compatibility moving forward. This will be undone again when we are ready, see https://github.com/obophenotype/uberon/issues/2723

PR #2736, opened 2022-12-09
#gogoeditdiff trigger on new comment

shawntanzk:
This PR makes the diff action trigger with new comments

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -2,13 +2,10 @@ name: 'Create ROBOT diffs on Pull requests'
 
 on: issue_comment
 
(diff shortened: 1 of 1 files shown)


shawntanzk:
#gogoeditdiff

shawntanzk:
#gogoeditdiff

shawntanzk:
#gogoeditdiff

shawntanzk:
according to https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issue_comment this needs to be on default branch, so I'm gonna push this through and test from there, will revert if there is issues

Issue #2714, opened 2022-11-17
[NTR] Moderator band a.ka. Septomarginal trabecula (heart)


emquardokus:
**Preferred term label:**
Moderator band

**Synonyms**
Septomarginal trabecula

**Definition (free text, please give PubMed ID)**
https://en.wikipedia.org/wiki/Moderator_band_(heart)
The moderator band is located in the right ventricle.[[2]](https://en.wikipedia.org/wiki/Moderator_band_(heart)#cite_note-:0-2)[[3]](https://en.wikipedia.org/wiki/Moderator_band_(heart)#cite_note-:1-3) The moderator band connects the base of the anterior papillary muscles to the ventricular septum.[[2]](https://en.wikipedia.org/wiki/Moderator_band_(heart)#cite_note-:0-2)

Haroun, Heshmat. “Comparative Anatomy of the Cardiac Septomarginal Trabecula (Moderator Band).” Anatomy Physiology & Biochemistry International Journal 2 (May 6, 2017). https://doi.org/10.19080/APBIJ.2017.02.555595.
" In man, it connects the interventricular septum with the anterior wall of the right ventricle at the base of the anterior papillary muscle. It is also present in non-human mammals where it stretches between the bases of the muscles papilaris subarteriosus and magnus. Its morphology and structure varies in different species. It carries the continuation of the right bundle branch of the conducting system of the heart."

Lee JY, Hur MS. Morphological classification of the moderator band and its relationship with the anterior papillary muscle. Anat Cell Biol. 2019 Mar;52(1):38-42. doi: 10.5115/acb.2019.52.1.38. Epub 2019 Mar 29. PMID: 30984450; PMCID: PMC6449581.

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
heart right ventricle UBERON:0002080 http://purl.obolibrary.org/obo/UBERON_0002080

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0001-7655-4833

bvarner-ebi:
@emquardokus,
thank you for the new term.

I propose the following modelling. Kindly comment if you propose revisions.

preferred term label:
septomarginal trabecula

synonym:
moderator band (Exact)
https://doi.org/10.1016/j.morpho.2010.03.004)

text defintion:
[...]

PR #2741, opened 2022-12-13
Addresses #2714 NTR septomarginal trabecula

bvarner-ebi:
Addresses #2714 NTR septomarginal trabecula
#gogoeditdiff


src/ontology/uberon-edit.obo
```diff
@@ -224033,6 +224033,17 @@ is_a: UBERON:0000029 ! lymph node
 relationship: dc-contributor http://orcid.org/0000-0002-1773-2692
 property_value: dcterms-date 2022-05-26T11:59:51Z xsd:dateTime
 
+[Term]
+id: UBERON:8470002
+name: septomarginal trabecula
+def: "A muscular trabecula in the right ventricle of the heart that links the lower part of the interventricular septum to the wall and to the base of the anterior papillary muscle of the right ventricle. The morphology and structure of the septomarginal trabecula varies in different species." [https://doi.org/10.1016/j.morpho.2010.03.004, https://doi.org/10.19080/APBIJ.2017.02.555595, https://orcid.org/0000-0001-7655-4833, PMID:30984450]
```
shawntanzk, line 224039: >The morphology and structure of the septomarginal trabecula varies in different species.

Not 100% sure this is useful in the def and might belong more in the comment, but I think its no the worst thing to have here

shawntanzk: My preference would be to change the DOIs to PMID, but small thing. If using doi, probably better to use DOI: instead of the whole link. Next version of protege will resolve it, and DOI is registered in bioregistry and all.

src/ontology/uberon-edit.obo
```diff
@@ -224033,6 +224033,17 @@ is_a: UBERON:0000029 ! lymph node
...
+comment: In humans, the septomarginal trabecula connects the interventricular septum with the anterior wall of the right ventricle at the base of the anterior papillary muscle. It is also present in non-human mammals where it stretches between the bases of the muscles papilaris subarteriosus and magnus. In some instances it carries the continuation of the right bundle branch of the conducting system of the heart. It is constituted of ordinary cardiac myocytes around and in between Purkinje myocytes. {xref="https://doi.org/10.1016/j.morpho.2010.03.004", xref="https://doi.org/10.19080/APBIJ.2017.02.555595"}
+synonym: "moderator band" EXACT [https://doi.org/10.1016/j.morpho.2010.03.004]
+is_a: UBERON:0000440 ! trabecula
+relationship: dc-contributor https://orcid.org/0000-0002-1773-2692
+relationship: located_in UBERON:0002080 ! heart right ventricle
```
shawntanzk, line 224044: Relation guide would be useful here >.< I'm not 100% sure about located in and would need context that maybe @dosumis could provide. (I think it is safe but would rather get someone who knows better)
heart right ventricle seems to be a bit confusing in how its modelled with the comment on cardiac chamber:
"generic enough to cover FBbt:00003156 heart chamber but this is a cavity. GO defines it as the cavity. TODO - move subclasses. Note this also includes sinus venosus" yet it *seems* material? 
Located in seems to not want it to be a spatial region having both range and domain being
```
'independent continuant'
 and (not ('spatial region'))
```
but I guess not doesn't work with ELK so that might not have been really enforced?

shawntanzk (commented):
see inline

src/ontology/uberon-edit.obo
bvarner-ebi: > My preference would be to change the DOIs to PMID, but small thing. If using doi, probably better to use DOI: instead of the whole link. Next version of protege will resolve it, and DOI is registered in bioregistry and all.

I don't see PMIDs for these references. I prefer the whole link for DOIs as it allows 1-click to access the reference.

src/ontology/uberon-edit.obo
bvarner-ebi: A ventricle is the hollow part of an organ (in this case the heart). That is why I chose "located in" as opposed to 'part of'.

src/ontology/uberon-edit.obo
shawntanzk: https://pubmed.ncbi.nlm.nih.gov/20359929/ for https://doi.org/10.1016/j.morpho.2010.03.004
https://pubmed.ncbi.nlm.nih.gov/22348232/ for https://doi.org/10.19080/APBIJ.2017.02.555595
DOI resolving will happen in next version of protege, and we try to use namespaces stuff where possible to keep things neat :) but yeah we will probably do a full programmatic change for DOI with the next update (I hope)

src/ontology/uberon-edit.obo
shawntanzk: I agree with you, just not sure if it is modelled that way in uberon (but i might be wrong) it's not super clear to me from a brief look, but I think if you trace back, it leads to something material.

shawntanzk: Screenshot 2022-12-13 at 16 19 46

src/ontology/uberon-edit.obo
bvarner-ebi: > https://pubmed.ncbi.nlm.nih.gov/20359929/ for https://doi.org/10.1016/j.morpho.2010.03.004
> https://pubmed.ncbi.nlm.nih.gov/22348232/ for https://doi.org/10.19080/APBIJ.2017.02.555595

Agree and will replace the first line.
The second line are not the same.

shawntanzk: whoops, my fault for multitasking lol

shawntanzk: btws could we avoid using the 2nd citation please? Publisher is potentially predatory and I'd rather not give them visibility :)

bvarner-ebi: > btws could we avoid using the 2nd citation please? Publisher is potentially predatory and I'd rather not give them visibility :)

I don't know anything about that and used the reference when creating the class as it was provided by the requester. If you want to remove it, please do so in a separate ticket and communicate your intention with the requester (@emquardokus).

shawntanzk: I don't think a separate ticket is correct as it is pertinent to the new term addition. I see that @emquardokus is already a reviewer, so maybe best to have a discussion on if we can use another citation here instead.

> I don't know anything about that

see https://beallslist.net/ 

Screenshot 2022-12-13 at 16 48 02

src/ontology/uberon-edit.obo
bvarner-ebi: At initial glance, this appears poorly modelled.

Text definition of ['cardiac chamber'](http://purl.obolibrary.org/obo/UBERON_0004151):
"A cardiac chamber surrounds an enclosed cavity within the heart."

['cardiac ventricle'](http://purl.obolibrary.org/obo/UBERON_0002082), which should be the cavity, is modelled as a subclass of 'cardiac chamber'.
If the ventricle is surrounded by the chamber it does not seem correct to model the ventricle as a type of chamber.

src/ontology/uberon-edit.obo
bvarner-ebi: This website is run by an anonymous source and provides no evidence for its accusations. Any publisher "may be" (insert adjective here) or is "potentially" (insert adjective here).

@emquardokus, appreciate your feedback here.

shawntanzk: Bealls list is pretty accepted by academics, though they do have frontiers in, which is very on the fence lol. I mean its a personal preference to avoid, but if its too much trouble just keep it in. Journals being predatory does not = paper not being reliable or good.

src/ontology/uberon-edit.obo
bvarner-ebi: Proposal:

use 'part of' some 'cardiac chamber' 
instead of 
'located in' in some 'cardiac ventricle'

and address ventricle modelling in separate ticket.

shawntanzk: I think it will eventually hit the same problem conceptually, but part_of is a clear pattern and I think that would be safe. 100% agree that tricky modelling thing should be in a separate ticket and not delay this unnecessarily :)

bvarner-ebi: Replaced 'located in' axiom with 'part of' axiom.
Also updated text def to make more general given the various morphologies reported and added another reference.

src/ontology/uberon-edit.obo
bvarner-ebi: Thanks for sharing this, @shawntanzk. It's not an issue of being too much trouble; rather, there does not seem to be compelling evidence which detracts from the validity of this particular reference.

shawntanzk: Don't have an issue with the reference itself, just the journal. It is hard to draw a line on if a journal is predatory so its tricky. I mean even nature does participate in some practices I would consider predatory (e.g. scientific reports) , yet these journals have really good papers in them. I think its partly my "spider senses" that tingled so I went to look the publisher up, and bealls list tends to be the first place to do so. Anyway here's the wiki page that might have some more evidence (with citations in them) https://en.wikipedia.org/wiki/Juniper_Publishers.

emquardokus: Thanks @shawntanzk, I'd never seen the Bealls list before, good to know!  I was primarily trying to find comparative anatomy since uberon is a multi-species ontology and this usually comes up in discussion if there is any questions.  If you think the journal publisher is too sketchy, please remove; however, it was the most comprehensive comparative anatomy paper I could find.  @bvarner-ebi Here's a different one to pigs, if you wanted to substitute: Gulyaeva AS, Roshchevskaya IM. Morphology of moderator bands (septomarginal trabecula) in porcine heart ventricles. Anat Histol Embryol. 2012 Oct;41(5):326-32. doi: 10.1111/j.1439-0264.2012.01139.x. Epub 2012 Feb 20. PMID: 22348232. 
I'm Okay with either choice here.

emquardokus (approved):
Approve of moderator band naming change and definition. I've added a different citation you may want to use instead of the Juniper press article for comparative anatomy, although it is more narrow to only pigs. Otherwise, looks good. thank-you

PR #2742, opened 2022-12-13
2022-12-13 Release

anitacaron:
We plan to have another release on the 15th of December with the base approach only.


bvarner-ebi (approved):
Spot checked uberon.obo and skimmed some of the other files. Expected changes found.

shawntanzk (approved):

PR #2745, opened 2022-12-15
Make gogoeditdiff manual

shawntanzk:
This PR makes it so that gogoeditdiff will only be triggered with a comment saying #gogoeditdiff With new pushes, this will not change until someone comments #gogoeditdiff again.
This action will only work when it is merged into master, but as a proof of concept, see my fork: https://github.com/shawntanzk/uberon/pull/2

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -1,19 +1,18 @@
 name: 'Create ROBOT diffs on Pull requests'
 
-on:
-  # Triggers the workflow on pull request events for the master branch
-  pull_request:
-    branches: [ master ]
-    paths:
-      - 'src/ontology/uberon-edit.obo'
+on: 
 
-  # Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch:
+  issue_comment:
+        types: [created]
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
   diff-reports:
+    if: ${{ github.event.issue.pull_request }}
     runs-on: macos-latest
+    permissions:
+      contents: read
+      pull-requests: read
     steps:
       - uses: khan/pull-request-comment-trigger@v1.1.0
         id: check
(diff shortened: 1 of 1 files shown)


ubyndr (approved):

.github/workflows/diff.yml
```diff
@@ -1,19 +1,18 @@
...
-  pull_request:
-    branches: [ master ]
-    paths:
-      - 'src/ontology/uberon-edit.obo'
+on: 
 
-  # Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch:
+  issue_comment:
+        types: [created]
```
matentzn, line 6: Excellent, so this fires whenever a comment is made (what does `created` means, as opposed to.. `editied`?), and `if: ${{ github.event.issue.pull_request }}` makes sure that the action is only run whenever the issue is on a pull request?

.github/workflows/diff.yml
```diff
@@ -31,8 +30,12 @@ jobs:
         run: docker pull obolibrary/odklite
         if: steps.check.outputs.triggered == 'true'
       # Checks-out current branch
+      - uses: xt0rted/pull-request-comment-branch@v1
```
matentzn, line 33: `xt0rted` -> be careful with re-using actions that you do not 100% trust.. Make sure all actions you use are safe!

matentzn (approved):

.github/workflows/diff.yml
matentzn: For example, this one has too few stars to be trustworthy. They could include malware in their action, or read our tokens or what not - at least ensure this cannot be the case.

.github/workflows/diff.yml
shawntanzk: yes, it will fire when any comment is created (as opposed to edited or deleted), and will skip everything if the comment isn't in a PR.

matentzn: cool!

.github/workflows/diff.yml
matentzn: I dont know either, sorry.

shawntanzk: I can do what deepak suggested and clone it into my repo and use that, but I don't think that's safe too, it could still redirect in the actions to redirect to a remote thing. The ideal situation would be to manually do the actions by scanning the code and copying it over I guess?

gouttegd: TypeScript is not my forte, but for what it’s worth, I see nothing nefarious in `xt0rted/pull-request-comment-branch@v1`… The code is fairly small (too small to hide anything), and it only depends on modules provided by GitHub itself.

Issue #2719, opened 2022-11-18
incisor teeth


meghalithic:
**Uberon term**
central incisor tooth and lateral incisor tooth

**Suggested revision of logical definition**
both have EQ that include a set of dentition, rather than a specific tooth
In non-human mammals, the central incisor is the same as the first incisor, and the lateral incisor is the same as the second incisor. Since mammals can have up to five incisors, lateral incisor is a meaningless term.

central incisor EQ:
'incisor tooth'
 and ('in central side of' some 'incisor region of dentition')

lateral incisor EQ:
'incisor tooth'
 and ('in lateral side of' some 'incisor region of dentition')

incisor region of dentition def:
The subdivision of the dentition that consists entire of incisor teeth.

As these are specific teeth, not a set of teeth, I propose modeling the central and lateral incisor after other teeth, like the molars and premolars.

As an example:
molar tooth 1:
def: The molar tooth of the upper or lower jaw that is phylogenetically number 1.
Subclass: molar tooth

megbalk:
@anitacaron is this issue done?

megbalk:
Can we get help with it?

PR #2757, opened 2022-12-21
central and lateral incisor tooth

meghalithic:
changing definition and EQ
issue #2719 

see discussion [here](https://docs.google.com/document/d/1XvMbNvr0FEsdqGhg79BYCYEHSqUxRHMcvhbGizEAht8/edit) on 28/11/2022.


src/ontology/uberon-edit.obo
```diff
@@ -168161,6 +168162,7 @@ id: UBERON:0016455
 name: upper lateral secondary incisor tooth
 subset: pheno_slim
 synonym: "maxillary lateral secondary incisor tooth" EXACT [FMA:55724]
+synonym: "secondary upper second incisor tooth" EXACT []
```
bvarner-ebi, line 168165: Is there a dbxref to support calling this the "second" incisor tooth?

bvarner-ebi (commented):
Hi @megbalk, I don't see any edits that would "break" the ontology in this PR, but from a content perspective I recommend seeking a review from someone qualified in this subject matter.

Some of the terms do not have text definitions and the added synonyms do not have dbxrefs to check.

Regarding, "... phylogenetically number #" in the text definitions... after a cursory search online, I could not find an explanation of a phylogenetic numbering system used in this context. This may be due to a lack of domain knowledge, which is why I recommend seeking a review from someone with subject matter expertise.

The annotation property ['tooth number'](http://purl.obolibrary.org/obo/UBPROP_0000112) mentions an ancestral count and a count along the A-P axis. I don't know if the most anterior tooth in a set is always ancestrally the first to appear. Hopefully someone with domain knowledge could help.

In summary, I recommend:

- adding text definitions where they are missing
- adding dbxrefs to support the synonyms
- seeking someone with subject matter expertise in dentition to review the content

If there is a reference that you can share that describes phylogenetically numbering from an anatomical perspective, I'm happy to take a look and do another review.

megbalk:
@cmungall @RDruzinsky could I get some help here?

anitacaron:
#gogoeditdiff

anitacaron:
#gogoeditdiff

anitacaron:
@wdduncan could you please review this PR?

anitacaron:
@wdduncan you can check the nice diff view [here](https://github.com/obophenotype/uberon/pull/2757#issuecomment-1874470012)

wdduncan:
This seems fine to me.

wdduncan (approved):

PR #2763, opened 2023-01-09
2023-01-09 Release

anitacaron:
Release with UBERON base.


shawntanzk (approved):
couldnt really look through all of uberon.obo, but uberon-base.owl looks good and nothing looks crazy on uberon.obo. PS is there a reason why we check in owl for -base file instead of obo?

matentzn:
The owl file is more complete then the obo file :)

Issue #2765, opened 2023-01-13
ID ranges


megbalk:
adding ID ranges for @helenamachado

megbalk:
Datatype: idrange:32
    Annotations:
        allocatedto: "Helena Machado"
    EquivalentTo:
        xsd:integer[> 8490000, <= 8499999]

PR #2766, opened 2023-01-13
adding id ranges for @helenamachado

megbalk:
see issue #2765

### DIFF
--- a/src/ontology/uberon-idranges.owl
+++ b/src/ontology/uberon-idranges.owl
@@ -212,4 +212,9 @@ Datatype: idrange:31
     EquivalentTo:
         xsd:integer[> 8480000, <= 8489999]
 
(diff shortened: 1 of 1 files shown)


anitacaron:
Shawn is on leave. Could you check this PR, @bvarner-ebi, please?

megbalk:
@anitacaron There seems to be a conflict where @wdduncan has idrange 31, where idrange 31 is also fro Paula Duek, and has the same range of ids. I will change @helenamachado id range, but we need to resolve the number conflict.

megbalk:
also @anitacaron can I add @helenamachado to the repo?

anitacaron:
@megbalk I can resolve the conflict, and I'll change Helena's id to 33. No problem at all; you can add her to the repo.

Issue #2453, opened 2022-05-04
Logical axioms from external relationships found in edit file


anitacaron:
After I did cleansing in #2377, some logical axioms are only present in UBERON. 

[Typedef]
id: capable_of
name: capable of
xref: RO:0002215
is_a: functionally_related_to ! functionally related to

[Typedef]
id: connected_to
name: connected to
xref: RO:0002170
is_a: transitively_connected_to ! transitively_connected to

[Typedef]
id: continuous_with
name: continuous with
xref: RO:0002150
is_symmetric: true

[Typedef]
id: develops_from
name: develops from
xref: RO:0002202
is_a: has_developmental_contribution_from ! has developmental contribution from

[Typedef]
id: drains
name: drains
xref: RO:0002179
domain: UBERON:0001638 ! vein

[Typedef]
id: has_branching_part
name: has branching part
xref: RO:0002569
is_transitive: true

[Typedef]
id: has_developmental_contribution_from
name: has developmental contribution from
xref: RO:0002254
is_transitive: true

[Typedef]
id: has_skeleton
name: has skeleton
xref: RO:0002551
domain: UBERON:0000475 ! organism subdivision
range: UBERON:0010912 ! subdivision of skeleton

[Typedef]
id: immediately_posterior_to
name: immediately posterior to
xref: BSPO:0015012
is_a: adjacent_to ! adjacent_to

[Typedef]
id: immediately_superficial_to
name: immediately superficial to
xref: BSPO:0015014
is_a: adjacent_to ! adjacent_to

[Typedef]
id: in_lateral_side_of
name: in lateral side of
xref: BSPO:0000126
disjoint_from: in_central_side_of ! in_central_side_of

[Typedef]
id: in_taxon
name: in taxon
xref: RO:0002162
holds_over_chain: attaches_to in_taxon
holds_over_chain: connected_to in_taxon
holds_over_chain: connects in_taxon
holds_over_chain: continuous_with in_taxon
holds_over_chain: developmentally_induced_by in_taxon
holds_over_chain: existence_ends_during in_taxon
holds_over_chain: existence_starts_during in_taxon
holds_over_chain: extends_fibers_into in_taxon
holds_over_chain: has_potential_to_develop_into in_taxon
holds_over_chain: innervated_by in_taxon
holds_over_chain: overlaps in_taxon
holds_over_chain: produced_by in_taxon
holds_over_chain: produces in_taxon
holds_over_chain: supplies in_taxon

[Typedef]
[...]

PR #2771, opened 2023-01-16
Remove RO logical axioms

anitacaron:
Related to #2453


src/ontology/uberon-edit.obo
```diff
@@ -224665,7 +224663,6 @@ xref: BFO:0000067
 id: continuous_with
 name: continuous with
 xref: RO:0002150
-is_symmetric: true
```
bvarner-ebi, line 224668: From [ticket](https://github.com/obophenotype/uberon/issues/2453#issuecomment-1118435290):

> [Typedef]
> id: continuous_with
> name: continuous with
> xref: RO:0002150
> is_symmetric: true
>  This should be added to RO https://github.com/oborel/obo-relations/pull/610

Do not see instruction to delete this.

src/ontology/uberon-edit.obo
```diff
@@ -224996,7 +224993,6 @@ xref: RO:0002002
 id: has_branching_part
 name: has branching part
 xref: RO:0002569
-is_transitive: true
```
bvarner-ebi, line 224999: From https://github.com/obophenotype/uberon/issues/2453#issuecomment-1118435290:

> [Typedef]
> id: has_branching_part
> name: has branching part
> xref: RO:0002569
> is_transitive: true
> Looks reasonable, although potential for massive proliferation of trivial inferences.
> 
>  Add to RO https://github.com/oborel/obo-relations/pull/612

Do not see instruction to delete this.

src/ontology/uberon-edit.obo
anitacaron: I'm cleansing the logical axioms already in RO.

bvarner-ebi (approved):
Regarding 'has_skeleton', unsure from ticket if any action was desired (see comments below). Otherwise changes seem to be in line with associated issue.

> [Typedef]
> id: has_skeleton
> name: has skeleton
> xref: RO:0002551
> domain: UBERON:0000475 ! organism subdivision
> range: UBERON:0010912 ! subdivision of skeleton
> I think the domain maybe too strict. Doesn't my foot have a skeleton. Is a foot an organism subdivision?
> 
>  Get feedback from @cmungall
>  Remove axiom in UBERON? 

...
> has_skeleton domain: this may be a case of a domain trying to do the closed-world work of a DP. Although I am not sure it's particularly problematic either - foots/autopods are organism subdivisions

Issue #2775, opened 2023-01-19
temporal crest


megbalk:
**Preferred term label:**

**Synonyms**

**Definition (free text, please give PubMed ID)**
A dorsally concave laterally directed crest that terminated on the caudo-dorsal process of the zygomatic bone.

xref: https://www.researchgate.net/publication/325258124_Morphological_Study_on_the_Skull_of_Wild_Boar_of_Mizoram_Sus_scrofa

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**

**Your nano-attribution (ORCID)**
0000-0003-2699-3066
0000-0001-8415-4848

PR #2778, opened 2023-01-19
adding temporal crest

megbalk:
Issue #2775


anitacaron:
Sorry, @megbalk, I don't have the knowledge to review NTR. Could you review it, @bvarner-ebi, please?

src/ontology/uberon-edit.obo
```diff
@@ -222238,6 +222238,15 @@ synonym: "fifth molar tooth" EXACT []
 is_a: UBERON:0003655 ! molar tooth
 relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
 
+[Term]
+id: UBERON:7500128
+name: temporal crest
+def: "A dorsally concave, laterally-directed crest that terminates on the caudal-dorsal process of a zygomatic bone." [DOI:10.5455/ijlr.20170908094210]
```
bvarner-ebi, line 222244: Considering UBERON:0004661 'mandible temporal crest'... 
If this new term is not a parent class to that, consider relabelling so that this term is differentiated from 'mandible temporal crest'.

The text definition references the zygomatic bone while the logical definition invokes the temporal bone. 
See these references, although they may be specific to humans:

https://academic.oup.com/asj/article/29/2/158/270420?login=true
https://www.imaios.com/en/e-anatomy/anatomical-structure/temporal-crest-1536898724

 If the meaning of this new term is specific to a particular taxon, consider adding taxon constraints.

bvarner-ebi (changes requested):
See inline comment.

src/ontology/uberon-edit.obo
megbalk: how about cranial temporal crest since the cranium excludes the mandible?

We will add the taxon constraint for Suidae too.

bvarner-ebi: @megbalk, labelling as 'cranial temporal crest' with exact synonym "temporal crest", or vice versa, seems reasonable.

If this structure is the same as [this](https://www.imaios.com/en/e-anatomy/anatomical-structure/temporal-crest-1536898724), consider adding the Latin term as a synonym.

After a cursory search, it does look like the structure in the human reference is also called a temporal line.

megbalk: The cranial temporal crest is bounded by the superior and inferior temporal lines, but is not the same as the temporal line. It is also not the same as the structure you linked, which is on the mandible not the temporal bone. We added clarification on the position of the cranial temporal crest. @helenamachado

I hesitate to make an exact synonym with "temporal crest" since there seems to be two different places it could be (on the mandible or on the temporal bone). Does a broad synonym to "temporal bone" make more sense?

bvarner-ebi: Hi, @megbalk,

> The cranial temporal crest is bounded by the superior and inferior temporal lines, but is not the same as the temporal line.

Consider adding a dbxref to support this assertion in the text def.

src/ontology/uberon-edit.obo
```diff
@@ -222238,6 +222238,16 @@ synonym: "fifth molar tooth" EXACT []
 is_a: UBERON:0003655 ! molar tooth
 relationship: dc-contributor https://orcid.org/0000-0003-2699-3066
 
+[Term]
+id: UBERON:7500128
+name: cranial temporal crest
+def: "A dorsally concave, laterally-directed crest that terminates at the caudal-dorsal process of a zygomatic bone. It is bounded by the superior and inferior temporal lines." [DOI:10.5455/ijlr.20170908094210]
+synonym: "temporal crest" BROAD []
```
bvarner-ebi, line 222245: > Does a broad synonym to "temporal bone" make more sense?

Assuming you mean "temporal crest", perhaps **related synonym** would be appropriate. 
This is because you are defining the term based on a paper that calls it the "temporal crest", so that term means exactly what your are defining, but that term may also refer to other "temporal crests", like as you noted above, on the mandible. Considering the potential ambiguity (the term may refer to exactly what you are defining or to a completely separate structure), related may be the "safest" choice.

src/ontology/uberon-edit.obo
```diff
@@ -222238,6 +222238,16 @@ synonym: "fifth molar tooth" EXACT []
...
+intersection_of: UBERON:4200133 ! crest
+intersection_of: part_of UBERON:0001678 ! temporal bone
```
bvarner-ebi, line 222247: From https://academic.oup.com/asj/article/29/2/158/270420?login=true:

Screenshot 2023-02-10 at 10 29 40

and Netter's illustration from https://pocketdentistry.com/7-temporal-and-infratemporal-fossae/:
ch07_u07-02-9781437726633-884638337

At least in humans, all parts of the temporal crest are **not** part of the temporal bone. Your dbxref focuses on _Sus scrofa_, so if this assertion is true in that species, a taxon restraint would be necessary.

bvarner-ebi (changes requested):
See inline comments.

src/ontology/uberon-edit.obo
megbalk: What is EA?

So the issue here is the EQ? Maybe it be better to model it after temporal fossa, so that the subclass (no EQ) would be:
'crest'
'part of' some 'cranium' (note that cranium excludes mandible, unlike skull)

bvarner-ebi (approved):
EA = equivalence axiom

Consider adding dbxref to synonym. Otherwise OK.

Issue #2788, opened 2023-02-03
Update text definition "perineurium"


aleixpuigb:
**Uberon term**
UBERON:0000121 Perineurium

**Suggested revision of textual definition**
A layer of thin, concentrically arranged cells with interspersed collagen that lies within the epineurium. The perineurium is a component of the blood–nerve barrier and is essential for protecting axons and their associated Schwann cells from ionic flux, toxins and infection.
DbXref: PMID:25818566

**Context**
Adding the blood-nerve barrier function in line with the edit of [perineurial cell](https://github.com/obophenotype/cell-ontology/issues/1726).

PR #2789, opened 2023-02-06
Addresses #2788 to update text definition of perineurium

aleixpuigb:
Addresses #2788 to update text definition of perineurium


bvarner-ebi (approved):
Could not access ISBN reference.
Otherwise OK.

bvarner-ebi (approved):

PR #2802, opened 2023-02-10
Uberon ODK 1.4 plus release

matentzn:


matentzn:
@anitacaron We will not merge this; I just want to see if it is ok, then I will create a proper release on the server. You will have to review locally as diff is too big, but I would like a sanity check :)

anitacaron:
Ok, I'll review it in the afternoon.

PR #2810, opened 2023-02-14
2023-02-14 Release

anitacaron:


anitacaron:
User-friendly diff file available [here](https://github.com/obophenotype/uberon/blob/2023-02-14-release/src/ontology/diffs/uberon-diff.md)

aleixpuigb (approved):

PR #2824, opened 2023-02-27
Update imports, components and fbbt mappings

### LINKED ISSUE #2815: Imports need refreshing
The _developmentally induced by_ relation (RO:0002256) still has range and domain restrictions to _connected anatomical structure_ (CARO:0000003).

Likewise, _results in developmental progression of_ (RO:0002295) still has a domain restriction to _connected anatomical structure_ (CARO:0000003).

Those restrictions are wrong and were found to cause several unsatisfiability issues. There were fixed in RO in early January (oborel/obo-relations#659, oborel/obo-relations#661) and the latest RO release happened a few days later, so why are the faulty restrictions still present in the latest Uberon release?

Were imports not updated prior to the latest release?

anitacaron:
Fixes #2815

### DIFF
--- a/src/ontology/components/hra_subset.owl
+++ b/src/ontology/components/hra_subset.owl
@@ -469,6 +469,15 @@
     
 
 
+    <!-- http://purl.obolibrary.org/obo/UBERON_0000482 -->
+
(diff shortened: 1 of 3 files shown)


gouttegd (approved):
All changes in the mappings are the changes I expected following the recent updates in FBbt.

PR #2830, opened 2023-02-28
Remove local logical axioms from `simultaneous with`

anitacaron:
Part of the great cleasing. Context: https://github.com/obophenotype/uberon/issues/2453


anitacaron:
#gogoeditdiff

bvarner-ebi (commented):
@anitacaron, from [this comment....](https://github.com/oborel/obo-relations/pull/611#issuecomment-1447030016)

- back out the two subproperties in this PR, on reflection these are not correct
- the corresponding injections should also be removed from uberon
- make the relation symmetric
- add the textual definition [@dosumis](https://github.com/dosumis) proposes

**Are the last two edits still required?**

anitacaron:
These were done in that [PR](https://github.com/oborel/obo-relations/pull/611/files). Here only get the reference.

bvarner-ebi (approved):

Issue #2838, opened 2023-03-14
NTR subsegmental bronchus


aleixpuigb:
**Preferred term label:**
subsegmental bronchus

**Synonyms**

**Definition (free text, please give PubMed ID)**
Branch of conducting airway between the segmental bronchus and the preterminal bronchiole. The wall contains plates of cartilage and submucosal glands.
ISBN:9780702077050
ISBN: 0781785898
https://orcid.org/0000-0002-9185-3994

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
intrapulmonary bronchus (UBERON:0035767)

**Your nano-attribution (ORCID)**
http://orcid.org/0000-0001-6677-8489

**Comment**
https://orcid.org/0000-0002-9185-3994: "UBERON:0035765 contains subsegmental lymph nodes defined as "The lymph nodes around the subsegmental bronchi.      but does not have an entry for subsegmental bronchi"

PR #2839, opened 2023-03-14
Addresses #2838 NTR subsegmental bronchus

aleixpuigb:
Addresses #2838 NTR subsegmental bronchus

@anitacaron @gouttegd , all property_value have added "", is that a consequence of the new protégé 5.6.1? Example: 
````
property_value: dcterms-date 2022-07-05T11:22:05Z xsd:dateTime --> 
property_value: dcterms-date "2022-07-05T11:22:05Z" xsd:dateTime

### CODE REVIEW

**Comment by anitacaron** [comment]
@aleixpuigb, I'll fix the conflict.

**Comment by aleixpuigb** [comment]
Thank you both for your help. Another thing I noticed is that there can't be two contributors, as the QC fails. [In CL used to be possible](https://www.ebi.ac.uk/ols4/ontologies/cl/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FCL_0004240?lang=en), so I am not sure if that depends on the ontology, or if the ID can only be associated to one contributor now.

**Comment by aleixpuigb** [comment]
Thank you @anitacaron, I used the wrong contributor.

**Comment by aleixpuigb** [comment]
Thank you, @anitacaron!

**Comment by bvarner-ebi** [comment]
@JoshuaFortriede, could you comment on the text definition and comment added to this term? A cursory online search shows somewhat different definitions, e.g., some images seem to classify it as bronchiolar, others classify it as any branching order after 3, etc.

Appreciate if you can shed some light on consensus on the meaning of this term...

**Comment by aleixpuigb** [comment]
As the order of branching was not part of the term request, and given that there is no consensus in the literature, I have decided to remove the comment.

File: src/ontology/uberon-edit.obo [hunk 1]
```diff
@@ -224009,6 +224009,18 @@ is_a: UBERON:8600006 ! visceral striated muscle tissue
 relationship: dc-contributor https://orcid.org/0000-0001-6677-8489
 property_value: dcterms-date "2023-02-21T09:26:41Z" xsd:dateTime
 
+[Term]
+id: UBERON:8600009
+name: subsegmental bronchus
+def: "Branch of conducting airway between the segmental bronchus and the preterminal bronchiole. The wall contains plates of cartilage and submucosal glands." [ISBN:0781785898, ISBN:9780702077050, Wikipedia:Bronchus]
```
bvarner-ebi, line 224015: Consider replacing "the" with "a", i.e., 
Branch of conducting airway between a segmental bronchus and a preterminal bronchiole.

I believe the second sentence is applicable to all bronchi (and the trachea), even though I did not see this mentioned in the wikipedia reference and don't have ready access to the ISBNs. Therefore, the features noted in the second sentence may suggest they are distinguishing features compared to segmental bronchi mentioned in the first sentence. 

If you'd like to keep it, consider writing, "Like other bronchi, the wall of a subsegmental bronchus contains..."

src/ontology/uberon-edit.obo
```diff
@@ -224009,6 +224009,19 @@ is_a: UBERON:8600006 ! visceral striated muscle tissue
 relationship: dc-contributor https://orcid.org/0000-0001-6677-8489
 property_value: dcterms-date "2023-02-21T09:26:41Z" xsd:dateTime
 
+[Term]
+id: UBERON:8600009
+name: subsegmental bronchus
+def: "Branch of conducting airway between a segmental bronchus and a preterminal bronchiole." [ISBN:0781785898, ISBN:9780702077050, Wikipedia:Bronchus]
+comment: Like other bronchi, the walls of subsegmental bronchus contain plates of cartilage. Notably, the cartilage plates become smaller and more widely dispersed with each branching. {xref="https://doi.org/10.1016/B978-0-12-381361-9.00009-3", xref="PMID:14913504"}
```
aleixpuigb, line 224016: The doi link is behind a paywall, but the following picture is in the chapter referenced. Interestingly, here they refer the subsegmental bronchus as intrasegmental bronchus, and therefore it will be added as a related synonym.
image

In the PMID reference, the text is more general to all bronchi, but it reflects that the most distally the bronchi are, the smaller and widely dispersed the cartilage plates get.
image

bvarner-ebi: Thank you for the illustration.
Based on this, `'distally connected to' some 'lobular bronchiole'` should be `'distally connected to' some 'bronchiole'`.

bvarner-ebi (approved):

PR #2840, opened 2023-03-15
Update imports

anitacaron:
I've verified if this brings unsats to UBERON. Everything looks good.


bvarner-ebi (approved):
- Did not see unsats when viewing in Protégé
- ^^xsd:string removed from definitions, which I believe was expected
- Did not see obvious major issues

Issue #1201, opened 2016-03-01
Define design patterns for mucosal glands


cmungall:
We should have a consistent pattern for defining mucosal glands.

Currently we do this using different patterns:
- by cell type: e.g. [UBERON:0000414](http://purl.obolibrary.org/obo/UBERON_0000414) ! mucous gland is defined by having mucus-secreting cells as part
- by location: e.g. [UBERON:0011148](http://purl.obolibrary.org/obo/UBERON_0011148) ! submucosal gland

We should clearly document any built in assumptions

 #1197 as test case

gouttegd:
WARNING: This issue has been automatically closed because it has not been updated in more than 3 years. Please re-open it if you still need this to be addressed addressed addressed – we are now getting some resources to deal with such issues.

Issue #1197, opened 2016-02-04
trachea gland UBERON_0005203 subclass of UBERON_0011148


cindyJax:
Move 'trachea gland' UBERON_0005203 to be a subclass of 'submucosal gland' UBERON_0011148 instead of 'gland' UBERON_0002530.

Update text definition (suggested): 
Trachea gland is a tubuloacinar seromucous gland that is located principally in the submucosa of the trachea.  Excretory ducts from this gland pass through the lamina propria to the tracheal lumen. These glands secrete mucus, lysozyme, defensins, and other agents that help protect the lungs from particles and infectious agents.
Refs: ISBN:1451113412, PMID:15973734, MGI:Anna and http://medical-dictionary.thefreedictionary.com/tracheal+glands

cmungall:
The classification as 'submucosal gland' should be inferred, if this is purely a location-based grouping; however, following wikipedia we define it based on a combination of location (submucosa) *and* cell-type. Also, location based classification may not be as simple as it at first seems due to glands crossing layers. 

more work needed, see #1201

PR #2843, opened 2023-03-16
Fixes #2842 logical definition trachea gland

### LINKED ISSUE #2842: Logical definition trachea gland
**Uberon term**
trachea gland UBERON:0005203

**Current logical definition and issue**

image
With the current logical definition, any cell in the trachea gland is considered part of the submucosa, and therefore a connective tissue cell. Moreover, in trachea glands there are non seromucus secreting cells (basal cells and myoepithelial cells)

**Suggested revision of logical definition**
'submucosal gland' and ('part of' some trachea)

aleixpuigb:
Fixes #2842 logical definition trachea gland

### DIFF
--- a/.github/workflows/auto-pr-ontobot.yml
+++ b/.github/workflows/auto-pr-ontobot.yml
@@ -65,6 +65,8 @@ jobs:
           labels: Automated
           body: ${{ env.PR_BODY }}
           title: ${{ env.PR_TITLE }}
+          author: ${{ env.ISSUE_CREATOR }} <${{ env.ISSUE_CREATOR }}@users.noreply.github.com>
+          committer: ${{ env.ISSUE_CREATOR }} <${{ env.ISSUE_CREATOR }}@users.noreply.github.com>
           base: ${{ github.head_ref }}
           branch: ${{ env.branch-name }}
-          token: ${{ secrets.GH_TOKEN }}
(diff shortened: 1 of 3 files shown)


bvarner-ebi (approved):
Approving with following caveats:

- Previous, seemingly unresolved, concerns with patterning: https://github.com/obophenotype/uberon/issues/1201 , https://github.com/obophenotype/uberon/issues/1197

- Similar terms are labeled 'x submucosal gland', e.g., 'bronchus submucosal gland'.
The PMID dbxref uses the term 'tracheal submucosal gland', which may be the ideal candidate for the label.

aleixpuigb:
> Previous, seemingly unresolved, concerns with patterning: https://github.com/obophenotype/uberon/issues/1201 , https://github.com/obophenotype/uberon/issues/1197

According to this, 'has part' some 'seromucus secreting cell' is correct, and I think I got it wrong. To clarify: if `a` has_part some `b`, can `c` be part of a? If yes, then there's no need to remove it and a new ticket should be created for all submucosal glands.

> Similar terms are labeled 'x submucosal gland', e.g., 'bronchus submucosal gland'.
The PMID dbxref uses the term 'tracheal submucosal gland', which may be the ideal candidate for the label.

I agree, I will open a new ticket to change the name and definition.

PR #2861, opened 2023-04-02
Remove two stray description properties

matentzn:
See https://github.com/INCATools/ubergraph/issues/121


matentzn:
Thanks a lot for checking @anitacaron!

Issue #2819, opened 2023-02-16
Review classification under "anatomical cluster"


gouttegd:
It has already been noted elsewhere (https://github.com/obophenotype/cell-ontology/issues/1782) that several terms currently classified under _anatomical cluster_ do not belong there. Time to review and fix those.

One of those terms at least is a cause of unsats as a result of that bogus classification: _hard palate_ (UBERON:0003216).

cmungall:
I think these kinds of very abstract grouping classes should be modeled as properties or characteristics of some kind. See:

https://github.com/BFO-ontology/BFO/issues/221

but happy to align with fbbt as an intermediate step

gouttegd:
In Uberon _disconnected anatomical group_ is a subclass of `'has quality' some disconnected` (PATO:0010001), and _anatomical structure_ is a subclass of `'has quality' some 'maximally connected'` (PATO:0010000).

So you would suggest to find (or create if they don’t already exist in PATO or elsewhere) more qualities to more finely distinguish the different types of entities? E.g., something like `'has quality' some 'closely grouped'` or similar for _anatomical cluster_, if we follow FBbt’s meaning of the term?

cmungall:
The PATO term seems a bit vague here. I think if we feel that clustered in the FBbt concept is a useful concept for AOs then it's worth making a PATO subclass?

It looks like many of the existing 'clusters' are essentially mereological sums / "set" terms. I think it's worth having a DP for these but I don't think that is the intent of the CARO class

gouttegd:
Just to be clear, _closely grouped_ does **not** exist in PATO, I just made it up for the example because it doesn’t look like there already is a quality in PATO that we could use to qualify “clusters”.

I’d be happy with a _clustered_ quality instead, with an agreed upon definition.

gouttegd:
Here’s the rest of the problems with the _anatomical cluster_ hierarchy.

Barring any objection, I’ll submit a PR to fix the issues highlighted in the first two sections below (cross-references and subclasses that belong elsewhere). The other issues will require more time and more input.

## Obsolete cross-references

[...]

PR #2862, opened 2023-04-03
Fix hierarchy under anatomical cluster (part 1)

gouttegd:
This PR is a first step towards fixing the hierarchy under [anatomical cluster](http://purl.obolibrary.org/obo/UBERON_0000477), as discussed in #2819 (especially [this comment](https://github.com/obophenotype/uberon/issues/2819#issuecomment-1490700354).

It clarifies the definition of _anatomical cluster_ to align it with the one used in FBbt, and moves out of the hierarchy all the terms that undoubtedly do not fit the definition.

Importantly, the re-classification of [hard palate](http://purl.obolibrary.org/obo/UBERON_0003216) as something else than an anatomical cluster fixes the unsatisfiability of [uvular muscle](http://purl.obolibrary.org/obo/UBERON_0010235) and related terms, because that muscle `has_muscle_insertion` to the hard palate, and the `has_muscle_insertion` relation applies to connected anatomical structures only (which a cluster is not).


src/ontology/uberon-edit.obo
```diff
@@ -7238,27 +7238,15 @@ relationship: present_in_taxon NCBITaxon:4751 ! Fungi
...
 name: anatomical cluster
-def: "Anatomical group that has its parts adjacent to one another." [http://orcid.org/0000-0001-9114-8737]
+def: "Anatomical group whose component anatomical structures lie in close proximity to each other." [FBbt:00007277]
 subset: common_anatomy
 subset: upper_level
-xref: AAO:0010009
-xref: AEO:0000041
-xref: BILA:0000041
-xref: CARO:0000041
 xref: EHDAA2:0003041
```
bvarner-ebi, line 7244: Could not find this in EHDAA2 using OLS. Should it be removed as well?

src/ontology/uberon-edit.obo
```diff
@@ -209360,7 +209348,7 @@ property_value: provenance_notes "This class was sourced from an external ontolo
 id: UBERON:3000809
 name: accessory articulation
 def: "Paired, additional vertebral articulations located medial to the intervertebral space, adjacent to the vertebral centrum." [AAO:Cannatella_1985]
-is_a: UBERON:0000477 ! anatomical cluster
+is_a: UBERON:0034921 ! multi organ part structure
```
bvarner-ebi, line 209351: Unsure if this refers to something that is "multi organ".

bvarner-ebi (approved):
Please see inline comments. 
Otherwise OK.

src/ontology/uberon-edit.obo
gouttegd: I can’t find it through Ontobee or BioPortal as well, so I agree, I’ll remove it.

src/ontology/uberon-edit.obo
gouttegd: Each vertebra is individually considered as a separate “organ” (a _vertebra_ is a _vertebral element_, which is a _skeletal element_, which is an _organ_), so I think it is consistent to say that an articulation between two vertebrae is a “multi organ part structure“ – it involves subparts of two different organs, which fits the definition of “multi organ part structure“.

But this makes me realise that this term can be classified even more precisely under _articulation_ (UBERON:0004905), which is a subclass of _multi organ part structure_. I’ll do that now.

bvarner-ebi (approved):

PR #2870, opened 2023-04-19
Remove reference to 'primary neuron'

### LINKED ISSUE #2869: 'neurectoderm' should not refer to 'primary neuron'
CL is planning to obsolete [primary neuron](http://purl.obolibrary.org/obo/CL_0000530), given the absence of a consensual meaning for such a term across species (obophenotype/cell-ontology#1931).

Uberon is using that term in its definition of [neurectoderm](http://purl.obolibrary.org/obo/UBERON_0002346), which should be amended accordingly.

gouttegd:
CL is obsoleting 'primary neuron', so Uberon cannot depend on it. Depending on an undefined term with no agreed upon meaning is a bad idea anyway.

closes #2869


gouttegd:
Note: Need to be sure the ZFIN folks are on board with the obsoletion of _primary neuron_ (https://github.com/obophenotype/cell-ontology/pull/1950) (i.e. with the idea that CL should _not_ be “the source of all cells”) before proceeding here.

gouttegd:
In the end we’re likely _not_ going to be obsoleting “primary neuron” in CL (see recent discussion on the linked CL ticket), but even then it’s going to be re-defined to become teleost-specific, so Uberon’s “neurectoderm” should still not depend on it and this PR remains relevant.

PR #2893, opened 2023-05-14
Revert to new serialisation format.

gouttegd:
Recent edits were performed with an older Protégé version, which caused the -edit file to be written using the serialisation behaviour of the old OWL API 4.5.6. This PR does nothing else but reverting the -edit file to the serialisation behaviour of OWL API 4.5.25 as used by Protégé 5.6.


gouttegd:
Merging without review as this solely a serialisation change, and it fixes a revert to the old serialisation that should never have occurred in the first place.

Issue #2885, opened 2023-05-09
Term for liver right lateral lobe requested


jrsjrs:
**Preferred term label:**
liver right lateral lobe

**Synonyms**
liver right posterior lobe

**Definition (free text, please give PubMed ID)**
The lateral or posterior sector of the right lobe of the liver.  The right lobe is divided into medial and lateral sectors by an oblique line that passes anteroposteriorly from the midpoint of the right lobe to the vena caval groove. (from https://www.kenhub.com/en/library/anatomy/functional-division-of-the-liver)

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
right lobe of liver (UBERON:0001114)

**Your nano-attribution (ORCID)**
ORCID:0000-0002-6443-9376

aleixpuigb:
Thank you for your term @jrsjrs. Kindly consider the following couple of questions and suggestions that I have regarding this term.

- Doing a cursory review, I've noticed that the majority of published work uses the posterior annotation, while lateral is more used for the left lobe. Is that correct, or `liver right lateral lobe` is more used and that was my perception?

- The second sentence of the definition describes the right lobe of liver. I propose the following definition: "The lateral (posterior) sector of the right lobe of the liver." Would that be sufficient?

- I would also like to include [this reference]( https://doi.org/10.1002/ca.22350) for the textual definition and this reference for the synonym: PMID: [30992874](https://pubmed.ncbi.nlm.nih.gov/30992874).

- The medial (anterior) sector is not in Uberon, would you mind that it is also created in this ticket?

jrsjrs:
*   Doing a cursory review, I've noticed that the majority of published work uses the posterior annotation, while lateral is more used for the left lobe. Is that correct, or liver right lateral lobe is more used and that was my perception?
     *   I am working on a curation project for the Somatic Cell Genome Editing Consortium and have requested this term to be used for a data submission that came to us as part of that project.  I can't comment on whether the more generally used term is "posterior…" or "lateral…"  I can only say that the data submitted to us used the terms left lateral lobe and right lateral lobe.   If you decide the term should be "liver right posterior lobe", that should be fine for our purposes as long as "liver right lateral lobe" is included as a synonym (preferably an exact synonym).
  *   The second sentence of the definition describes the right lobe of liver. I propose the following definition: "The lateral (posterior) sector of the right lobe of the liver." Would that be sufficient?
     *   Yes, that is fine.
[...]

Issue #2832, opened 2023-03-03
NTR: brain regions


paulacalipho:
Term: left fronto parietal lobe
Synonyms: 
exact: left fronto-parietal lobe PMID:3703146, PMID:27003199, PMID:23808679, PMID:30797641
Definition: The left frontal and parietal lobes of the brain hemisphere.
SubClass Of: ‘organ part’ UBERON:0000064 and ('part of' some ‘left cerebral hemisphere’ UBERON:0002812)
https://orcid.org/0000-0002-0819-0473

Term: right fronto parietal lobe
Synonyms: 
exact: right frontal parietal lobe PMID: 29678709
exact: right frontal-parietal lobe PMID: 22568892
exact: right fronto-parietal lobe PMID: 34234736, PMID:34736485, PMID:30169667
Definition: The right frontal and parietal lobes of the brain hemisphere.
SubClass Of: ‘organ part’ UBERON:0000064 and ('part of' some ‘right cerebral hemisphere’ UBERON:0002813)
https://orcid.org/0000-0002-0819-0473

Term: left occipital parietal lobe
Synonyms: 
exact: left parietal and occipital lobes PMID:30383697
exact: left parietal-occipital lobe PMID:22799762, PMID:23506194
exact: left parieto-occipital lobe PMID:27028405, PMID:36462980, PMID:36542021, PMID:36198519, PMID:36451532, PMID:36090026, PMID:25389451
broad: left occipital-parietal region PMID: 26300758, PMID:17215478, PMID:35257944
broad: left occipital/parietal region PMID:36561241, PMID:22078521
broad: left occipital parietal region PMID:26027887
broad: left occipito-parietal region PMID:27303466
Definition: The left occipital and parietal lobes of the brain hemisphere.
SubClass Of: ‘organ part’ UBERON:0000064 and ('part of' some ‘left cerebral hemisphere’ UBERON:0002812)
Mapping to other ontologies: Left parietal and occipital lobes, CS C1267724
https://orcid.org/0000-0002-0819-0473

Term: right occipital parietal lobe
Synonyms: 
broad: right occipital-parietal region PMID:35257944
broad: right occipital parietal region PMID:31053785, PMID:22711620
Definition: The right occipital and parietal lobes of the brain hemisphere.
SubClass Of: ‘organ part’ UBERON:0000064 and ('part of' some ‘right cerebral hemisphere’ UBERON:0002813)
Mapping to other ontologies: Right parietal and occipital lobes CS C1267723
[...]

PR #2894, opened 2023-05-15
Addresses #2885 NTR liver right lobe sections

aleixpuigb:
Addresses #2885 NTR liver right lobe sections.


aleixpuigb:
I have decided to use posterior/anterior nomenclature (in the left lung is lateral and medial) as it seems the standard nomenclature after a cursory review.
image PMID: [30992874](https://pubmed.ncbi.nlm.nih.gov/30992874)

Lateral and medial nomenclature was added as exact synonyms, based on this tesxt in the reference:

image

Of note, some child classes of the left and right liver lobe need revision in a separate ticket.

src/ontology/uberon-edit.obo
```diff
@@ -224175,6 +224175,25 @@ relationship: has_part CL:0000067 ! ciliated epithelial cell
...
+synonym: "liver right lateral lobe" EXACT [PMID:24453062]
+relationship: dc-contributor https://orcid.org/0000-0001-6677-8489
+relationship: dc-contributor https://orcid.org/0000-0002-6443-9376
+relationship: part_of UBERON:0001114 ! right lobe of liver
+property_value: dcterms-date "2023-05-15T10:46:34Z" xsd:dateTime
+
+[Term]
+id: UBERON:8600016
+name: liver right anterior lobe
+def: "The anterior sector of the right lobe of the liver." [PMID:30992874]
```
bvarner-ebi, line 224191: Regarding text def for both: sector --> segment

bvarner-ebi: @aleixpuigb, thank you for clarifying. 

The labels and synonyms include the word "lobe", but the text definition of these classes are describing sectors, not lobes. A similar comment related to brain regions: https://github.com/obophenotype/uberon/issues/2832#issuecomment-1520538497

If the class is meant to represent the sectors, consider renaming the label and synonyms.

Also consider adding synonyms that substitute "liver" with "hepatic" as included in the dbxrefs, e.g., "right hepatic anterior sector" PMID:[30487943](https://pubmed.ncbi.nlm.nih.gov/30487943)

bvarner-ebi (changes requested):
See inline comments.

src/ontology/uberon-edit.obo
```diff
@@ -224175,6 +224175,29 @@ relationship: has_part CL:0000067 ! ciliated epithelial cell
...
+relationship: dc-contributor https://orcid.org/0000-0001-6677-8489
+relationship: dc-contributor https://orcid.org/0000-0002-6443-9376
+relationship: part_of UBERON:0001114 ! right lobe of liver
+property_value: dcterms-date "2023-05-15T10:46:34Z" xsd:dateTime
+
+[Term]
+id: UBERON:8600016
+name: anterior sector of right lobe of liver
+def: "The anterior sector of the right lobe of the liver." [PMID:30992874]
+synonym: "liver right medial lobe" EXACT [https://orcid.org/0000-0002-6443-9376]
```
bvarner-ebi, line 224194: Consider making this a related, not exact, synonym. Sector vs lobe.

PR #2925, opened 2023-06-16
add logical axioms to UBERON:0039850

### LINKED ISSUE #2850: UBERON:0039850 does not have any logical axioms
Looking at uberon.owl, UBERON:0039850 seems not have any axioms. 
Have checked uberon-edit.obo and it is the same so its not a release issue but an actual issue. 
At minimum the term should have a parent class.

shawntanzk:
added basic logical axioms for the term, and importantly, parentage
Fixes #2850


shawntanzk:
#gogoeditdiff

aleixpuigb (approved):

Issue #2936, opened 2023-06-22
UBERON:7500102 missing child terms


shawntanzk:
From @aleixpuigb  
"given the definition (of UBERON:7500102), I miss some child terms that are part of the preorbital bone region (jugal bone, lacrimal bone and maxilla bone)."
See https://github.com/obophenotype/uberon/pull/2935#pullrequestreview-1492681196

PR #2416, opened 2022-04-13
preorbital edits

megbalk:
This PR relates to issue #1888 to create add "preorbital process" as a synonym to "preorbital process of the pars facialis of the maxilla"; issue #2385 to add "preorbital" as a region of the face and fix the synonym in "antorbital" from "perorbital" to "periorbital"; and issue #2386 to create the term "preorbital fossa". This was done with help from @helenamachado

PR #2942, opened 2023-06-23
add part of to 'preorbital bone region' structures

shawntanzk:
Fixes https://github.com/obophenotype/uberon/issues/2936
Note, did not do much research, just following definition as suggested by @aleixpuigb 
Honestly just had a bit of time between finishing my last task and drinks so... lol  
#gogoeditdiff


aleixpuigb:
Hi @megbalk and @helenamachado, I would like to ask your opinion about this ticket since you added this term in #2416. Given the definition, would it be ok to add the axiom `'part of' some 'preorbital bone region'` to `jugal bone`, `lacrimal bone` and `maxilla`?

megbalk:
@aleixpuigb yep! For maxilla, it is only the part anterior to the orbit. I suppose it could be also added for nasal bones, but only sometimes as it depends on the location of the nasal bones.

src/ontology/uberon-edit.obo
```diff
@@ -49873,6 +49875,7 @@ relationship: develops_from UBERON:0010334 {source="EHDAA2"} ! maxillary process
 relationship: in_lateral_side_of UBERON:0011156 {source="FMA-abduced-lr"} ! facial skeleton
 relationship: part_of UBERON:0003113 {notes="facial series", source="ISBN:0073040584"} ! dermatocranium
 relationship: part_of UBERON:0003277 {source="VHOG"} ! skeleton of upper jaw
+relationship: part_of UBERON:7500102 ! preorbital bone region
```
aleixpuigb, line 49878: I suggest 'overlaps' instead of 'part of'

aleixpuigb (changes requested):
Please, see inline comment.

shawntanzk:
#gogoeditdiff

shawntanzk:
thanks @aleixpuigb & @megbalk - not sure if the other part_of are alright, but I have changed the one you pointed out as suggested. Also realised automated diff didnt run - I dont think there will be any issue, but in case, would be more comfortable letting that run and making sure nothing weird is inferred from these additions

aleixpuigb (approved):

PR #2950, opened 2023-07-04
Add xrefs curated in Biomappings

cthoyt:
This PR adds 54 xrefs to BTO terms and one to a MeSH term that have been curated in Biomappings. It also annotates the contributor. This diff was generated programatically from tools in Biomappings.

Blocked by:
- [x] New release of ROBOT that upgrades the OWL API from 4.5.25 to 4.5.26 https://github.com/ontodev/robot/pull/1135
- [x] New release of ODK with ^ https://github.com/INCATools/ontology-development-kit/issues/925


src/ontology/uberon-edit.obo
```diff
@@ -2174,6 +2174,7 @@ def: "The surface (external) layer of ectoderm which begins to proliferate short
 subset: pheno_slim
 synonym: "surface (external) ectoderm" EXACT []
 synonym: "surface ectoderm" EXACT [MP:0012532]
+xref: BTO:0006242 {http://purl.org/dc/terms/contributor="https://orcid.org/orcid:0000-0003-4423-4370"} ! external ectoderm
```
gouttegd, line 2177: Until the OBO format has explicit support for arbitrary axiom annotations, annotations with unsupported annotation properties are best avoided. With the current OBO parser, they get mangled into monstrosities like this: `http://www.geneontology.org/formats/oboInOwl#http://purl.org/dc/terms/contributor`.

Please stick to `oboInOwl:hasDbXref` annotations for the time being.

And what is this ORCID format: `https://orcid.org/orcid:0000-0003-4423-4370`? Please use `https://orcid.org/0000-0003-4423-4370`.

gouttegd (changes requested):
Sorry for the late negative review, but we can’t accept that contribution in its current state. Please see inline comment.

src/ontology/uberon-edit.obo
cthoyt: Thanks for pointing out the issue with the ORCID IRIs - that was an oversight.

However, I have seen arbitrary axiom annotations in other ontologies that curate their edit file in OBO format such as MONDO.

I have tested further and found that switching `http://purl.org/dc/terms/contributor` for `dcterms:contributor` works as intended. Here's the OWL axiom that's generated after converting from OBO format to OWL RDF/XML:

```xml
<owl:Axiom>
    <owl:annotatedSource rdf:resource="http://purl.obolibrary.org/obo/UBERON_0000018"/>
    <owl:annotatedProperty rdf:resource="http://www.geneontology.org/formats/oboInOwl#hasDbXref"/>
    <owl:annotatedTarget>IDOMAL:0002421</owl:annotatedTarget>
    <dcterms:contributor>https://orcid.org/0000-0003-4423-4370</dcterms:contributor>
</owl:Axiom>
```

Is there another issue with arbitrary axioms in the OBO format that I'm not aware of?

gouttegd: > I have tested further and found that switching http://purl.org/dc/terms/contributor for dcterms:contributor works as intended.

No it doesn’t: it creates an annotation with a `dcterms:contributor` property but look at the `dcterms:` prefix declaration in the XML header:

```
xmlns:dcterms="http://www.geneontology.org/formats/oboInOwl#dcterms:"
```

This results in the annotation property effectively being `http://www.geneontology.org/formats/oboInOwl#dcterms:contributor`, which obviously is incorrect.

Please don’t try to circumvent the limitations of the OBO format unless you really know what you are doing.

gouttegd:
Not sure what’s going on with the QC. It was killed because OWLTools ran out of memory, I don’t think this has anything to do with your PR. I did run the test suite on your branch on my local machine and it passed fine.

src/ontology/uberon-edit.obo
cthoyt: Is there a way to include an appropriate definition of DCTERMS? otherwise I can change this to use the oboInOwl relation

gouttegd: Not with the current version of our tools. The next version of ROBOT should allow to use any annotation property you want, with the syntax you initially used (`{http://purl.org/dc/terms/contributor="..."}`).

Rumour has it that the next version of ROBOT is only a couple of weeks away from being released, so if you’d really like to use that property we can simply put that PR on hold until then.

Otherwise, for now you can use either `{source="..."}` (preferably) or `{hasDbXref="..."}` (_should_ work).

cthoyt: I think I can wait for the next version of ROBOT! This will be great. Thanks for the clarification

cthoyt: Is there a PR associated with this change in ROBOT we can link to?

gouttegd: The real change is in the OWL API. It happened here: https://github.com/owlcs/owlapi/pull/1099 and is now available in OWL API version 4.5.26.

The change in ROBOT is simply to upgrade the OWL API from 4.5.25 to 4.5.26: https://github.com/ontodev/robot/pull/1135

cthoyt: @gouttegd I updated to use the syntax you specified. With the new release of ROBOT, do you think this works now? Or are there other changes I should make?

cthoyt:
I think this will require 1) a new ODK release https://github.com/INCATools/ontology-development-kit/issues/925 and 2) updating UBERON's ODK config

src/ontology/uberon-edit.obo
gouttegd: This will work, but we must still wait for the next ODK release (which will include the latest ROBOT version), _and_ for the next Protégé release (which will use OWL API 4.5.26). Both should happen next week.

One thing you I’d like to ask you to change in the meantime, though, is removing the comment at the end of each statement (e.g., ` ! external ectoderm`). It does not really serve any purpose, and more importantly it will be automatically removed anyway the first time any editor does any change with Protégé and saves the -edit file. That’s because Protégé has no way of knowing the label of the foreign term (I assume in your case it comes from Biomappings, but Protégé is ignorant of that – for Protégé the xref is opaque).

So in order to avoid an otherwise unavoidable spurious diff in the first commit after this PR is merged, please remove those comments.

gouttegd:
Yes for ①, no for ②. No need for a ODK config update for a point release like 1.4.3. :) All that will be required is to update the QC workflow to use the latest ODK release.

src/ontology/uberon-edit.obo
cthoyt: Done in 60f2ad6df.

gouttegd: Excellent. As soon as we’ll have ODK 1.4.3 there should be no obstacle to merging this.

anitacaron:
@gouttegd, isn't this PR constantly failing because of the property issue in the OBO format?

gouttegd:
@anitacaron I don’t think so. Last QC checks passed fine.

The problem with this kind of syntax:

```
{http://purl.org/dc/terms/contributor="https://orcid.org/0000-0003-4423-4370"}
```

is that it currently yields bogus IRIs (such as `http://www.geneontology.org/formats/oboInOwl#http://purl.org/dc/terms/contributor`), and that’s why we are waiting for the next ODK, but such bogus IRIs should not, by themselves, causing the QC checks to fail.

It’s only when the property IRI contains a `#` character that there is a real problem, for example with `http://www.w3.org/2000/01/rdf-schema#seeAlso`. Because then the bogus IRI contains *two* `#` characters (`http://www.geneontology.org/formats/oboInOwl#http://www.w3.org/2000/01/rdf-schema#seeAlso`), and that’s something the RDF/XML parser doesn’t like *at all*. But the present PR does not use any such property.

gouttegd (approved):
As discussed with @cthoyt on Slack, in the future the plan is for Uberon to automatically import Biomappings-provided mappings by exploiting the Biomappings SSSOM export.

Most of the machinery to do that will be added with #3061, but since it is not ready yet, no objection to adding the xrefs directly in the -edit file for now. I’ll remove them when we’ll switch to the SSSOM import.

Issue #2952, opened 2023-07-06
Relationship between 'terminal bronchiole' and 'secondary pulmonary lobule'


aleixpuigb:
**Uberon term**
'terminal bronchiole' UBERON:0002187
'secondary pulmonary lobule' UBERON:0010369

**Suggested revision and reasons**
No relationship between these terms exists currently in uberon. The current definition of `secondary pulmonary lobule` is:
> A unit of lung supplied by three to five terminal bronchioles and contained by fibrous septa.

This definition is supported by the dbxref [PMID:3259815](https://www.ajronline.org/doi/epdf/10.2214/ajr.151.1.21):
![image](https://github.com/obophenotype/uberon/assets/94959119/0acec19e-b24c-460c-aaf5-2ecdb393839b)

and this article https://doi.org/10.1148/93.3.507:
> The secondary pulmonary lobule can be defined in two ways: [...] and (b) a unit of lung structure supplied by a cluster of three to five terminal bronchioles.

However, in the dbxref [PMID:3259815](https://www.ajronline.org/doi/epdf/10.2214/ajr.151.1.21), it also can be found:
![image](https://github.com/obophenotype/uberon/assets/94959119/1f1f9cd4-b9cc-421a-9554-8acd97044565)
![image](https://github.com/obophenotype/uberon/assets/94959119/0c9d67da-39da-47de-824c-f9698dbe5fe1)

The RO term in supplies is only applied in arteries:
> Relation between an arterial structure and another structure, where the arterial structure acts as a conduit channeling fluid, substance or energy.

And the parent property `connected to` doesn't seem adequate as from the last definition the `terminal bronchiole` is inside the `secondary pulmonary lobule` (partly or completely).

Therefore, the object properties that would be most appropriate would be `part of` if all and the whole terminal bronchioles are inside the secondary pulmonary lobules, or `overlaps` if all terminal bronchioles are partly inside the secondary pulmonary lobules.

In this [article](https://www.semanticscholar.org/paper/Thin-section-CT-of-the-secondary-pulmonary-lobule%3A-Webb/659e8cbf37f651ca37166afedc1acdde89e07b19) (behind a paywall but figures and tables are available), the diagram shows that the terminal bronchiole is inside the lobule and is the lobular bronchiole that enters in the secondary pulmonary lobule:
![image](https://github.com/obophenotype/uberon/assets/94959119/bcf81db2-f9eb-418a-ab97-8a4818b4e0af)

[...]

PR #2953, opened 2023-07-06
Addresses #2952 terminal bronchiole part of secondary pulmonary lobule

aleixpuigb:
Addresses #2952 terminal bronchiole part of secondary pulmonary lobule.
@JoshuaFortriede, as some of the information about it was a bit ambiguous (see ticket #2952), I would appreciate if you could confirm that this edits are appropriate? Thank you!

### DIFF
--- a/.github/workflows/auto-pr-ontobot.yml
+++ b/.github/workflows/auto-pr-ontobot.yml
@@ -47,7 +47,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install ontobot-change-agent==0.3.11
+          pip install ontobot-change-agent
 
       - name: Run ochange
         id: ochange
@@ -55,7 +55,8 @@ jobs:
           ochange process-issue ${{ env.resource }} \
           -r ${{ steps.gh-script-repo.outputs.result }} \
           -n ${{ steps.gh-script-issue.outputs.result }} \
-          -g ${{ secrets.GH_TOKEN }}
+          -g ${{ secrets.GH_TOKEN }} \
+          -p UBERON
 
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v4


src/ontology/uberon-edit.obo
```diff
@@ -43847,7 +43847,7 @@ property_value: structure_notes "lacks submucosa and cartilage plates; they have
 [Term]
 id: UBERON:0002187
 name: terminal bronchiole
-def: "The last conducting structure of non-respiratory bronchioles; after this point, the airways have alveoli in their walls." [ISBN:0-397-51047-0, MGI:cwg, MP:0002268]
+def: "A bronchiole that is part of a secondary pulmonary lobule and the last conducting structure of non-respiratory bronchioles; after this point, the airways have alveoli in their walls." [ISBN:0-397-51047-0, MGI:cwg, MP:0002268, PMID:16543587]
```
bvarner-ebi, line 43850: Given the ambiguity you detail in the linked issue, keeping the original definition (which doesn't seem contested) and using the relationship "overlaps" may be a less controversial solution. Will defer to @JoshuaFortriede or another SME for confirmation.

JoshuaFortriede: [Webb 2006](https://edisciplinas.usp.br/pluginfile.php/5126574/mod_folder/content/0/Thin-section%20CT%20of%20the%20secondary%20pulmonary%20lobule.pdf?forcedownload=1) (paywall article) states:

`Each secondary lobule is supplied by a small bronchiole`

and [Ohno 2021](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8165947/) states:

`The secondary pulmonary lobule is supplied by a lobular bronchiole `

These definitions match my understanding; that the primary bronchiole is the street into the neighborhood (secondary pulmonary lobule), and the terminal bronchioles are the main streets contained within that neighborhood. As such, I believe the new definition works.

I would also suggest adding that the terminal bronchiole is between the primary bronchiole and respiratory bronchiole. Specifically the terminal is distal to the primary, and proximal to the respiratory.

Here is my cheat sheet that I made to keep these airways/structures and space/segments together. The "Type" or RO might not be completely accurate.
image

aleixpuigb: Thank you for your input and clarification, @JoshuaFortriede.

> I would also suggest adding that the terminal bronchiole is between the primary bronchiole and respiratory bronchiole. Specifically the terminal is distal to the primary, and proximal to the respiratory.

In a different ticket #2957 I have already added `'terminal bronchiole' SubclassOf ' distally connected to' some 'respiratory bronchiole'`. However, [mice have none or few respiratory bronchioles](https://doi.org/10.1186/s13326-019-0209-1), and therefore 'terminal bronchioles' are connected to alveolar ducts. Therefore I will have to revert that axiom in a different ticket. On the other side, the lobular bronchiole can be found between the primary bronchiole and the respiratory bronchiole. I would like to investigate a bit more before adding that axiom.

image

bvarner-ebi: Consider using singular form in the def:

A bronchiole that is part of a secondary pulmonary lobule and the last conducting structure of **a** non-respiratory **bronchiole**; after this point, the airways have alveoli in their walls.

bvarner-ebi (approved):

Issue #2926, opened 2023-06-19
Object properties for vascular system and pulmonary system


aleixpuigb:
Context: To record relations between two blood vessels or between two airway ducts in uberon, there are two main object properties that have been used. However, in many cases the definition of these object properties do not align with their usage:

- [connected to](http://purl.obolibrary.org/obo/RO_0002170): a is connected to b if and only if a and b are discrete structure, and there exists some connecting structure c, such that c connects a and b.
     Example of misusage (mainly there is no structure c):
     - [respiratory bronchiole](http://purl.obolibrary.org/obo/UBERON_0002188): 'proximally connected to' some 'terminal bronchiole'
     - [proper hepatic artery](http://purl.obolibrary.org/obo/UBERON_0015480): 'connected to' some 'proper hepatic artery'
        
![image](https://github.com/obophenotype/uberon/assets/94959119/1e911ebe-edc0-4867-95bc-58e4b8ca3a47)

- [branching part of](http://purl.obolibrary.org/obo/RO_0002380): x is a branching part of y if and only if x is part of y and x is connected directly or indirectly to the main stem of y.
     Example of misusage (the blood vessel is not part of the other blood vessel:
     - [femoral artery](http://purl.obolibrary.org/obo/UBERON_0002060): 'branching part of' some 'external iliac artery'
     - Querying 'part of' some 'iliac artery' returns many arteries that most likely are not part of it:
       
![image](https://github.com/obophenotype/uberon/assets/94959119/84e2bd68-9183-4d71-9f04-22324930658d)

The question is, are these object properties good enough to record these relationships? Should their definition be changed? There is an object property, [connecting branch of](http://purl.obolibrary.org/obo/RO_0002252), that is not present in uberon that [could partly solve this problem](https://github.com/obophenotype/uberon/wiki/Branching-structures-Design-Pattern), Is there any issue for importing it to uberon?
```[tasklist]
### Tasks
- [ ] Import 'connecting branch of' to uberon
```

PR #2958, opened 2023-07-06
Addresses #2957 revision of distal pulmonary hierarchy

### LINKED ISSUE #2957: Edit pulmonary hierarchy
The distal pulmonary hierarchy needs the following edits: 
- Remove `bronchiole SubclassOf 'part of' some 'pulmonary lobule'`. Only distal bronchioles or respiratory bronchioles are part of a (secondary) pulmonary lobule
- Remove `'pulmonary acinus' SubclassOf 'pulmonary lobule'`. This is based on the premise that 'pulmonary acinus' is an exact synonym of 'primary pulmonary acinus'. However, according to this [reference](https://doi.org/10.53347/rID-8763), "A primary pulmonary lobule is defined as the lung unit distal to the respiratory bronchioles. It is significantly smaller than an [acinus](https://radiopaedia.org/articles/pulmonary-acinus?lang=us), and is composed of alveolar ducts, alveolar sacs and alveoli."
- Add `'pulmonary acinus' SubclassOf 'part of' some 'secondary pulmonary lobule'`. According to this [reference](https://doi.org/10.53347/rID-8761), "Secondary pulmonary lobules represent a cluster of up to 30 [acini](https://radiopaedia.org/articles/pulmonary-acinus?lang=us)"
- Remove `'pulmonary lobule' SubclassOf 'has part' some 'terminal bronchiole'`. In the ticket #2952 'terminal bronchiole' will be SubclassOf 'part of' some 'secondary bronchiole' and in the ticket #2956 'pulmonary lobule' will be merged with 'secondary pulmonary bronchiole'.
- - Add `'terminal bronchiole' SubclassOf ' distally connected to' some 'respiratory bronchiole'`. See diagram below.
- Add `'terminal bronchiole' SubclassOf ' distally connected to' some 'pulmonary acinus'`. 'terminal bronchiole' is 'distally connected to' some 'respiratory bronchiole', and respiratory bronchiole are 'part of' some 'pulmonary acinus'. See diagram
image

- Add `'pulmonary acinus' SubclassOf 'connected to' some 'terminal bronchiole'`. See previous point.

aleixpuigb:
Fixes #2957 revision of distal pulmonary hierarchy
Only hierarchy is revised, no text definitions or dbxrefs. These will be addressed in a separate ticket.


aleixpuigb:
#gogoeditdiff

src/ontology/uberon-edit.obo
```diff
@@ -43863,6 +43862,8 @@ xref: SCTID:278981008
 xref: UMLS:C0225666 {source="ncithesaurus:Terminal_Bronchiole"}
 xref: Wikipedia:Terminal_bronchiole
 is_a: UBERON:0002186 ! bronchiole
+relationship: distally_connected_to UBERON:0002188 ! respiratory bronchiole
```
bvarner-ebi, line 43865: Consider if 'continuous with' is a more accurate object property.

src/ontology/uberon-edit.obo
```diff
@@ -43863,6 +43862,8 @@ xref: SCTID:278981008
...
+relationship: distally_connected_to UBERON:0008874 ! pulmonary acinus
```
bvarner-ebi, line 43866: The text def of connected to:
"a is connected to b if and only if a and b are discrete structure, and there exists some connecting structure c, such that c connects a and b"

In this relationship, what is 'c'? Can it be 'respiratory bronchiole' if 'respiratory bronchiole' is already asserted to be part of the 'pulmonary acinus'?

bvarner-ebi (commented):
Kindly see inline comments.
Will approve if there are no concerns expressed from assignee's side.

src/ontology/uberon-edit.obo
aleixpuigb: Thank you for this comment, Bradley. There is this related ticket in Uberon #2926 and it was treated in the last Uberon call (19/06/23) about the usage of connected to and how it is not aligned with the definitions. However, it was agreed to use this object property for the pulmonary system.

bvarner-ebi (approved):

src/ontology/uberon-edit.obo
aleixpuigb: yes, in this case 'respiratory bronchiole' would be c and the usage of connected_to would be more aligned with the definition.

Issue #2961, opened 2023-07-07
[Edit] Restrict current 'parallel fiber' to teleost; [NTR] new non-teleost term 'parallel fiber, bifurcated'


bvarner-ebi:
Additional context: https://github.com/obophenotype/cell-ontology/issues/2012

The current Uberon class '[parallel fiber](http://purl.obolibrary.org/obo/UBERON_2002218)' was sourced from [TAO](https://www.ebi.ac.uk/ols4/ontologies/tao).

Not surprisingly, it contains axioms that are specific to [Teleostei](http://purl.obolibrary.org/obo/NCBITaxon_32443), such as `SubClass Of 'part of' some 'molecular layer valvula cerebelli'`. 
'[valvula cerebelli](http://purl.obolibrary.org/obo/ZFA_0000603)' is "uniquely present in ray-finned fishes".

Also, 'parallel fibers' in teleost fish [do not bifurcate in contrast to other vertebrates](https://link.springer.com/referenceworkentry/10.1007/978-94-007-1333-8_63).

[GO has the term 'parallel fiber'](http://purl.obolibrary.org/obo/GO_1990032) that ostensibly refers to non-teleost fish vertebrates as it is defined as "[resulting] from the bifurcation of a cerebellar granule cell axon in the molecular layer into two diametrically opposed branches, that are oriented parallel to the long axis of the folium".

**Action Items:**

For UBERON:2002218 'parallel fiber':
- [x] relabel 'parallel fiber' to 'parallel fiber, teleost'.
- [x] add axiom `SubClass Of 'only in taxon' some 'teleost fish'`
- [x] add text definition: "A parallel fiber results from the axon of a cerebellar granule cell ascending through the molecular layer where it synapses on the dendritic branches of Purkinje cells and stellate cells. The parallel fibers in teleost fish do not bifurcate in contrast to other vertebrates."
doi:10.1007/978-94-007-1333-8_63

Add non-teleost, new term 'parallel fiber, bifurcated' to UBERON: 
- [x] add term based on [GO term](http://purl.obolibrary.org/obo/GO_1990032). 
- [x] add annotation `never_in_taxon 'teleost fish'`
- [x]  add `Disjoint With some 'parallel fiber, teleost'`

Add a general grouping class 'parallel fiber' to include the two terms above.
- [x] add `overlaps some 'cerebellar granule cell'`

PR #2965, opened 2023-07-11
Restrict current 'parallel fiber' to teleost; [NTR] new non-teleost term 'parallel fiber, bifurcated'; [NTR] grouping class 'parallel fiber' #2961

bvarner-ebi:
Addresses #2961
- Restrict current 'parallel fiber' to teleost; 
- Add new, non-teleost term 'parallel fiber, bifurcated'
- Add new grouping class 'parallel fiber'


src/ontology/uberon-edit.obo
```diff
@@ -190087,6 +190087,24 @@ is_a: UBERON:0001145 {source="FMA"} ! ovarian vein
 intersection_of: UBERON:0001145 ! ovarian vein
 intersection_of: drains UBERON:0002119 ! left ovary
 
+[Term]
+id: UBERON:0700019
+name: parallel fiber
+def: "A parallel fiber results from the axon of a cerebellar granule cell ascending through the molecular layer where it synapses on the dendritic branches of Purkinje cells and stellate cells." [GO:1990032] {comment="doi:10.1007/978-94-007-1333-8_63"}
+comment: This a grouping class for parallel fibers in any species. If annotating a parallel fiber in any species other than teleost fish, consider using 'parallel fiber, bifurcated'.
+is_a: UBERON:0006134 ! nerve fiber
```
aleixpuigb, line 190095: After a cursory review, I found that they are in the cerebellum. Consider adding 'part of' some cerebellum https://doi.org/10.3389/fncir.2020.611841

src/ontology/uberon-edit.obo
```diff
@@ -190087,6 +190087,24 @@ is_a: UBERON:0001145 {source="FMA"} ! ovarian vein
 intersection_of: UBERON:0001145 ! ovarian vein
 intersection_of: drains UBERON:0002119 ! left ovary
 
+[Term]
+id: UBERON:0700019
+name: parallel fiber
+def: "A parallel fiber results from the axon of a cerebellar granule cell ascending through the molecular layer where it synapses on the dendritic branches of Purkinje cells and stellate cells." [GO:1990032] {comment="doi:10.1007/978-94-007-1333-8_63"}
```
aleixpuigb, line 190093: the dbxref is using a rdfs comment as an annotation property instead of database_cross_reference

src/ontology/uberon-edit.obo
```diff
@@ -190087,6 +190087,24 @@ is_a: UBERON:0001145 {source="FMA"} ! ovarian vein
...
+comment: This a grouping class for parallel fibers in any species. If annotating a parallel fiber in any species other than teleost fish, consider using 'parallel fiber, bifurcated'.
+is_a: UBERON:0006134 ! nerve fiber
+relationship: has_part CL:0001031 ! cerebellar granule cell
```
aleixpuigb, line 190096: I am not sure if I agree with this axiom. The definition of has_part is not very clear either: a core relation that holds between a whole and its part.
The way I understand is that the whole (but not all!) cerebellar granule cell are part of the fiber. However, in this case, only the axioms are part of the fiber. Overlaps is more general, but that's the object property I would consider using in this case.
However, as @dosumis [requested this axiom](https://github.com/obophenotype/cell-ontology/issues/2012#issue-1747510248), I will defer the review of it to him.

aleixpuigb (changes requested):
Please, see inline comments. 
I will defer the review of 'has part' some 'cerebellar granule cell' to requester.

src/ontology/uberon-edit.obo
bvarner-ebi: > I am not sure if I agree with this axiom. The definition of has_part is not very clear either: a core relation that holds between a whole and its part.
> The way I understand is that the whole (but not all!) cerebellar granule cell are part of the fiber. However, in this case, only the axioms are part of the fiber. Overlaps is more general, but that's the object property I would consider using in this case.
> However, as @dosumis [requested this axiom](https://github.com/obophenotype/cell-ontology/issues/2012#issue-1747510248), I will defer the review of it to him.

**overlaps** seems to be a reasonable relation in this instance. 
@dosumis, kindly advise if there is a compelling reason to use 'has part' instead of overlaps here.

aleixpuigb (approved):

src/ontology/uberon-edit.obo
bvarner-ebi: Checking a similar relationship:
 '[dentate gyrus mossy fiber](http://purl.obolibrary.org/obo/GO_0044302)' has
`SubClass Of 'part of' some 'granule cell'`.

Given that, here is the updated proposal:

'parallel fiber' `SubClass Of 'part of' some 'cerebellar granule cell'`

bvarner-ebi:
#gogoeditdiff

src/ontology/uberon-edit.obo
aleixpuigb: While the majority of parallel fibers are unmyelinated, there are some myelinated parallel fibers [PMID:15869526](https://pubmed.ncbi.nlm.nih.gov/15869526). As the myelin sheath is not part of the 'cerebellar granule cell', I would not use 'part of'.

PR #2968, opened 2023-07-12
Fix preprocessing step.

### LINKED ISSUE #2967: Remove separately maintained ro_import and dot-definitions from uberon-merged.owl
Even though Uberon has switched to the base merging mechanism several months ago already, it is still actively using the individual `ro_import.owl` module, which was last updated in January.

This is because of this step early in the pipeline:

```
	$(OWLTOOLS) --no-logging $< $(COMPONENTSDIR)/disjoint_union_over.ofn issues/contributor.owl --merge-support-ontologies -o $@ && $(SCRIPTSDIR)/expand-dbxref-literals.pl $@ > $@.tmp
	# The previous step seems to be necessary because somehow the expand-dbxref-literals script expects the owtools output.. No idea why
	$(ROBOT) expand -i $@.tmp --no-expand-term http://purl.obolibrary.org/obo/RO_0002175 \
		query \
			--update $(SPARQLDIR)/remove_axioms.ru \
			--update $(SPARQLDIR)/delete-definition-dot.ru \
		merge -i $(IMPORTDIR)/ro_import.owl -o $@
```

Among other things, this step drops every single axioms relating to RO (`--update $(SPARQLDIR)/remove_axioms.ru`), then re-imports them from the individual `ro_import.owl` module (`merge -i $(IMPORTDIR)/ro_import.owl`).

As a result, all the RO axioms coming from the (up-to-date) `merge_import.owl` module are discarded, and replaced by the RO axioms coming from the (stale) `ro_import.owl` module. Among other consequences, it means that several bogus range/domain constraints that had been fixed in RO in the recent months are creeping back in.

The rationale for this “remove-then-reimport” dance was to “remove RO labels to avoid duplicates then adds them back to ensure they are canonical”. I believe after the switch to base merging this is now completely unnecessary, in addition to being harmful.

gouttegd:
The preprocessing step of the custom pipeline is needlessly complicated in that it:

- invokes a script (`expand-dbxref-literals.pl`) that does nothing useful (it does _not_ find any CURIE to expand);
- in order to invoke that script, proecesses the edit file through OWLTOOLS (all for nothing, since the script itself does nothing);
- discards all RO axioms coming from the merged import module and reimport older versions from the stale `ro_import` module;
- removes "dot" definitions but leaves around their annotations, now attached to blank nodes.

Therefore we replace the entire step by a much simpler step that only keeps the useful bits of the old preprocessing step, that is:

- merging the edit file with the `disjoint_union_over.ofn` and `contributor.owl` components;
- expanding macros.

closes #2967


anitacaron:
> invokes a script (expand-dbxref-literals.pl) that does nothing useful (it does not find any CURIE to expand);

I suspect it's not working anymore because the script is searching for this pattern with datatype that has now changed with the new serialization.

```
if (m@(.*)<oboInOwl:(\S+) rdf:datatype="http://www.w3.org/2001/XMLSchema#string">(\S+)</oboInOwl:\S+>(.*)@) {
```

As soon as I get my docker working again, I can test it and review this PR properly.

gouttegd:
> I suspect it's not working anymore because the script is searching for this pattern with datatype that has now changed with the new serialization.

Yes. And there are been 4 releases since we switched to the new serialisation, so it’s been 4 releases without CURIE expansion in the `oboInOwl:source` and `oboInOwl:hasDbXref` tags and nobody had complained about it. This strongly suggests this expansion, even when it worked, was not needed to begin with.

gouttegd:
I already have a _draft_ of such a plugin, so by the time we get a version of ROBOT that supports plugins, we could have, yes.

anitacaron:
In the meantime, it's better if we fix the script. Can you do this in this PR, please?

gouttegd:
I can do it, yes, but not in this PR. CURIE expansion was not broken by this PR. The fix, if a fix there must be, does not belong here.

If CURIE expansion is a behaviour that is actually desired, then someone open a ticket to explicitly ask for that behaviour to be restored and I’ll fix the script right away in a new PR. (I don’t need to know _why_ CURIE expansion is needed – just that it is an actual need and not merely a leftover that lingered here just because nobody knew what it was for.)

Sorry, but the custom Makefile is full of hacks that have been there for so long that nobody remembers why they are here at all. I won’t maintain a hack just because we _think_ that maybe someone, somewhere could possibly have a need for it – even though that someone didn't notice for six months that the hack was not working anymore, despite the extensive testing that all members of the tech group did back when we switched to the new serialisation.

matentzn (approved):
I have no strong objections here, other than the fact that I would like to make sure that the issue on the expand curie script is resolved #2974 

@anitacaron I think given the argument that it "does not do anything right now" trumps the argument of proper process in this case. Please keep your rigorous reviews up because as I am getting old, I am loosing my edge and may wave things through that I shouldn't. In this case though I gave it some thought and I think we can trust @gouttegd to do his thing.

PR #2979, opened 2023-07-19
Fix invalid IRI to RO relation in preprocessing step.

gouttegd:
The recent overhaul of the preproprecessing step (#2968) introduced an invalid reference to RO:0002175. As a result, that annotation was now expanded along with the other annotations, instead of being explicitly left unexpanded as was the intention.

This PR fixes the invalid reference to restore the non-expansion of RO:00002175.


matentzn (approved):
Gr8t catch!

PR #2983, opened 2023-07-21
Fixes #2884 harmonise spine segments

### LINKED ISSUE #2884: Revise and harmonise nomenclature for spine segments
The nomenclature used for the spine is variable depending on the region:

image
image

- [ ] A consensus is needed to harmonise the nomenclature. First option is using the letter + number formula (eg C1, T3, L2, S5,... segment of X spinal cord). The second option is using ordinal numbers  (eg first lumbar spinal cord segment). I'm in favour of the first option as it is more standard, is more visual and keeps the segments in order.
- [ ] Segments lack definition, would a standard definition be sufficient (eg "The first segment of the cervical region in the spinal cord")?
- [ ] The current terms are a representation of the human spine. Should we leave it as it is or add more segments? (Some mammals have as many as 25 thoracic segments or more!)

aleixpuigb:
Fixes #2884 harmonise spine segments


src/ontology/uberon-edit.obo
```diff
@@ -112210,9 +112216,12 @@ is_a: UBERON:0007716 ! lumbar subsegment of spinal cord
 
 [Term]
 id: UBERON:0006452
-name: fourth thoracic spinal cord segment
+name: T4 segment of thoracic spinal cord
+def: "The segment of the spinal cord that corresponds to the fourth thoracic vertebra in most mammals." [PMID:19876658]
 subset: defined_by_ordinal_series
+synonym: "fourth thoracic spinal cord segment" EXACT [BIRNLEX:1378]
 synonym: "T4 segment" EXACT [FMA:71170]
+synonym: "T4 segment of spinal cord" EXACT []
```
aleixpuigb, line 112224: I have deleted this synonym, and the commit is in the branch but not in the PR was some reason.
image

bvarner-ebi (approved):

Issue #2987, opened 2023-07-26
base-file orphans a few terms 


shawntanzk:
Not sure why this is happening, but I assume it is unintended. 
It seems that terms that are under UBERON:0001062 ('anatomical entity') that are in the UBERON namespace gets the subClassOf UBERON:0001062 removed in base file orphaning them as top level terms. 
There are currently 3 terms that fall into that:
UBERON:8600016
UBERON:8600015
UBERON:0008874

shawntanzk:
@matentzn @anitacaron or @gouttegd would you mind helping me take a look at this please? Thanks :)

gouttegd:
From what I can see, it’s not a problem with the generation of the base file. Those terms are already orphaned at the source, in the `-edit` file.

shawntanzk:
@gouttegd ah dang you are right - I guess some process in materialization or something gave them parents in the full file but not the base file for some odd reason. Sorry for not checking the edit file, dont tend to use that these days >.< Will make a PR to fix it in the edit file, thanks heaps :) 
PS I thought the base file came out as a product of the full file, and given the full file has subClassOf anatomical entity, how did this get dumped in base?

gouttegd:
> I thought the base file came out as a product of the full file

No, the `-base` file comes straight from the `-edit` file (well, almost: it comes from the “preprocessed” `-edit` file):

```
# base: All the axioms as they are editted by the editors, excluding reasoning
$(ONT)-base.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC)
        $(ROBOT_RELEASE_IMPORT_MODE_BASE) \
        $(SHARED_ROBOT_COMMANDS) \
        annotate --link-annotation http://purl.org/dc/elements/1.1/type http://purl.obolibrary.org/obo/IAO_8000001 \
                --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) \
                --output $@.tmp.owl && mv $@.tmp.owl $@
```

So any relationship coming from the materialisation leading to `-full` will _not_ show up in `-base`.

PR #2988, opened 2023-07-26
Add parents to orphaned terms

shawntanzk:
Fixes https://github.com/obophenotype/uberon/issues/2987

Did not do much digging/research, just adding basic parents, more precise parentage can come in another PR if needed


src/ontology/uberon-edit.obo
```diff
@@ -128290,6 +128290,7 @@ xref: NCIT:C33426
 xref: SCTID:361365004
 xref: UMLS:C0225696 {source="ncithesaurus:Pulmonary_Acinus"}
 xref: Wikipedia:Acinus#The_Lungs
+is_a: UBERON:0034922 ! cell cluster
```
aleixpuigb, line 128293: The pulmonary acinus contain alveolar ducts and respiratory bronchioles, I don't think it should be classified as a cell cluster. Should it also be an anatomical entity?

aleixpuigb (changes requested):
Please, consider a different parent for pulmonary acinus.

src/ontology/uberon-edit.obo
shawntanzk: happy to put it as anatomical entity :) only put cell cluster cause I searched up acinus and it said cell cluster.

aleixpuigb: Yes, the pulmonary acinus is a bit different than other acini.

aleixpuigb:
I have received a notification of this commit [14dd3a6](https://github.com/obophenotype/uberon/commit/14dd3a618a7ddc9c38c11d8658eb7a175acf33a7), but I can't see it in the PR. Something similar happened to #2983. I can see the commit in the branch, but it will not merge if you mergw the PR.
image
Any idea why does that happen?

shawntanzk:
ok a bit confused, hate not having github desktop, but I have committed the requested change to the branch (https://github.com/obophenotype/uberon/commit/14dd3a618a7ddc9c38c11d8658eb7a175acf33a7) but it does not seem to be showing up in the PR - Is there something I am doing wrong?

shawntanzk:
ok fixed, sync issue apparently, see https://stackoverflow.com/questions/45626986/updating-pull-request-after-review-changes-commit-doesnt-show-up-in-the-pull-re

aleixpuigb (approved):

gouttegd (approved):

PR #3010, opened 2023-08-09
Do not (try to) rebuild the merged developmental stage ontologies.

### LINKED ISSUE #3009: Uberon should not try to build "developmental-stage-ontologies"
One step of the Uberon pipeline is this:

```Makefile
$(TMPDIR)/update-stages: $(SRC) | $(TMPDIR)
        rm -rf $(TMPDIR)/developmental-stage-ontologies && \
        cd $(TMPDIR) && \
       git clone https://github.com/obophenotype/developmental-stage-ontologies.git && \
       cd developmental-stage-ontologies/src && make OWLTOOLS=owltools all ssso-merged.obo -B && cd ../../../ && \
       touch $@
```

where we clone the [obophenotype/developmental-stage-ontologies](https://github.com/obophenotype/developmental-stage-ontologies) repo and rebuild its products, notably this `ssso-merged.obo` file.

As of today, this build is broken (consequently breaking Uberon’s build as well) because some upstream ontologies the `developmental-stage-ontologies` repo depends on are no longer available.

I think we should _not_ try to rebuild `developmental-stage-ontologies` as part of the Uberon pipeline. The Uberon pipeline is hard enough to debug, we don’t need to make it even harder by having nested pipelines inside.

**Short-term**, we should just clone the `developmental-stage-ontologies` repo and use its products as they are, without trying to rebuild them. This way, if there is a problem in the `developmental-stage-ontologies` pipeline (as there currently is), said problem will remain isolated to that repo (where hopefully it would be fixed at some point) without breaking Uberon.

If we want to have up-to-date developmental stage ontologies, all we’d need to do would be to update the `developmental-stage-ontologies` repo, outside of Uberon.

**Longer-term**, the `developmental-stage-ontologies` repo could be updated to make proper releases (kind of like what is suggested in #2047, except it should not be done in the Uberon repo), that Uberon could then just import as any other foreign ontology.

gouttegd:
After cloning the `developmental-stage-ontologies` repo, we should use the products that have been committed there as they are, without trying to rebuild them.

This may mean we'll end up with less up-to-date things (if one of the taxon-specific stage ontology has been updated since the last time product files were committed to `developmental-stage-ontologies`), but it will prevent a failure of the `developmental-stage-ontologies` pipeline to cause a failure of the Uberon pipeline. If someone wants ultra-fresh developmental stage ontologies, then someone can take care of updating the `developmental-stage-ontologies` repo.

closes #3009

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -1141,9 +1141,7 @@ $(REPORTDIR)/%-synclash: %.obo
 $(TMPDIR)/update-stages: $(SRC) | $(TMPDIR)
 	rm -rf $(TMPDIR)/developmental-stage-ontologies && \
 	cd $(TMPDIR) && \
-	git clone https://github.com/obophenotype/developmental-stage-ontologies.git && \
-	cd developmental-stage-ontologies/src && make OWLTOOLS=owltools all ssso-merged.obo -B && cd ../../../ && \
-	touch $@
+	git clone https://github.com/obophenotype/developmental-stage-ontologies.git
 
 CSTAGES := $(filter-out %bridge-to-uberon.obo, $(wildcard $(TMPDIR)/developmental-stage-ontologies/*/*-uberon.obo))
 #CSTAGES := $(wildcard $(TMPDIR)/developmental-stage-ontologies/*/*-uberon.obo)


matentzn (approved):
Assuming that I can update the developmental stage ontologies upstream.

gouttegd:
> Assuming that I can update the developmental stage ontologies upstream.

So do you prefer that I wait until you have done so (here I suppose: https://github.com/obophenotype/developmental-stage-ontologies/pull/86) before I merge this? I don’t mind, just to know.

matentzn:
I would not approve this PR if I would prefer it :) I prefer to remove obstacles from the bulldozer... :D It will be fine, don't worry. I will just merge it if I don't hear and ask for forgiveness

gouttegd:
Okie. :)

I suppose the “bulldozer” refers to my Makefile cleaning initiative… That’s quite inaccurate: I am not running a bulldozer through the Makefile, I am just detonating some well-placed demolition charges. :D

Issue #2999, opened 2023-07-28
Remove dependence of GO.xrf_abbs in pipeline


cmungall:
GO.xrf_abbs predates various prefix harmonization efforts.

We should just add bioregistry based prefix checking to OAK `validate` if it's not there already

Issue #2980, opened 2023-07-19
`components/reflexivity_axioms.owl` is unused


gouttegd:
We have an awkward step in the Makefile (and I am not even the one who says it is awkward):

```Makefile
# somewhat awkward: we temporarily inject reflexivity axioms
# **Hacking_Feb_2022** Notes start here
# Background: 
## This ends up in the release file, reflexivity_axioms.owl temp merge as consistency QC
## This is only a QC.  Rationale - this would catch some errors because expands part disjointness ()
## **Hacking_Feb_2022** TODO - rewrite as as expansions from annotation axioms using SPARQL. Expanded axioms --> component (as for taxon restrictions).

TMP_REFL=$(COMPONENTSDIR)/reflexivity_axioms.owl
DEVELOPS_FROM_CHAIN=$(COMPONENTSDIR)/develops-from-chains.owl
# see https://github.com/obophenotype/uberon/issues/2381

$(TMPDIR)/materialized.owl: $(TMPDIR)/unreasoned.owl $(TMP_REFL)
	$(ROBOT) merge -i $< -i $(DEVELOPS_FROM_CHAIN) --collapse-import-closure false \
		relax \
		materialize -T $(CONFIGDIR)/basic_properties.txt -r elk \
		reason -r elk --exclude-duplicate-axioms true --equivalent-classes-allowed asserted-only \
		unmerge -i $(TMP_REFL) \
		unmerge -i $(DEVELOPS_FROM_CHAIN) \
		annotate -O $(URIBASE)/uberon/materialized.owl -V  $(RELEASE)/materialized.owl -o $@ 2>&1 > $@.LOG
.PRECIOUS: $(TMPDIR)/materialized.owl
```

What _should_ happen in this step, among other things, is that the `components/reflexivity_axioms.owl` component is merged with the ontology, materialisation and reasoning are performed, then the axioms from the `reflexivity_axioms.owl` components are removed from the output.

That is _not_ what actually happens. Notice how the `reflexivity_axioms.owl` component, even though it is listed as a dependency of `$(TMPDIR)/materialized.owl`, is never actually merged in – only the `develops-from-chains.owl` component is merged.

This is seemingly the result of the conversion of that rule from OWLTOOLS to ROBOT, which happened in commit a3be1f63775feb0e8ce1ff3b0b03410ef236b9aa. Here is the relevant diff:

```diff
-ext.owl: $(TMPDIR)/materialized.owl $(TMP_REFL)
[...]

Issue #1501, opened 2019-05-26
Explore entry for Bgee Anatomical homology


marcrr:
Hi,

I suggest adding our new "Anatomical homology" page to the "Explore" menu on http://uberon.github.io/:

https://bgee.org/?page=anat_similarities

Marc

mellybelly:
+1

matentzn:
@marcrr we are getting to this now.

As we are in the process of streamlining quite a lot in Uberon, can you tell us which exact files in uberon BGEE is depending on? For all your use cases? uberon.owl? homology.owl? composite-metazoan.owl?

fbastian:
@matentzn we rely on composite-metazoan.owl to retrieve the full ontology, and ext.owl to compute the taxon constraints.

fbastian:
We update Uberon only for major releases of Bgee, which we just did, so I won't know for a while!

matentzn:
cc @cmungall related to #2032

PR #3028, opened 2023-08-18
Cleaning up makefile

### LINKED ISSUE #2973: Cleaning up the custom Makefile
The `uberon.Makefile` is currently a big mess.

Since burning it down to the ground and starting anew is not an option (not that I wouldn’t want to do it, but more reasonable people on this project would disagree, and that’s probably for the best), we need to improve it incrementally instead.

## Purging commented out code
It’s easy to think that code that is commented out is harmless, but I argue that it is not the case. The amount of commented code in the Makefile is one of the things that make it very hard to read. I can understand the reluctance to throw away old code just in case it could be useful again, but we should remember that everything is under version control anyway: we’ll never definitely lose anything.

So I propose to mercilessly erase commented out code in cases like the following examples:

```Make
##MAKEOBO= owltools $< --remove-axiom-annotations -o -f obo $@.tmp1 && grep -v ^property_value: $@.tmp1 | grep -v ^owl-axioms: > $@.tmp && obo2obo $@.tmp -o $@
MAKEOBO=  $(OWLTOOLS) $< --add-obo-shorthand-to-properties  -o -f obo --no-check $@.tmp1 && grep -v ^property_value: $@.tmp1 | perl -npe 's@relationship: dc-@property_value: dc-@' | grep -v ^owl-axioms: > $@.tmp && mv $@.tmp  $@

[…]

# Not used anywhere so commenting out
#$(TMPDIR)/phenoscape_homology.owl:
#       wget --no-check-certificate https://raw.githubusercontent.com/phenoscape/phenoscape-ontologies/master/demo/phenoscape_homology.owl -O $@

[…]

# First pass at making base module
# Currently this will be missing the temporary reflexivity axioms.
# should this include downward-injected axioms, e.g on ZFA?
#uberon-base.owl: $(TMPDIR)/unreasoned.owl
#       $(OWLTOOLS) $< --remove-imports-declarations --remove-axioms -t Declaration --set-ontology-id -v $(RELEASE)/$@ $(URIBASE)/uberon/$@ -o $@.tmp && mv $@.tmp $@
```

etc.

## Purging obsolete comments
In addition to commented out _code_, there are also plain text comments that are clearly outdated. Again, they are not harmless. One can lose quite a bit of time just trying to figure out whether a given comment is actually saying something useful that one should be aware of, or if it is something that is only of interest for Uberon archeologists.

For example, there’s an entire blocks of comments dedicated on _phenoscape_ and how to merge it with Uberon – but phenoscape as a separate entity is no longer a thing [since two years ago](https://github.com/obophenotype/uberon/pull/1916)!

Same punition as above: we should delete such comments. Again, they will remain in the history of the project for eternity, if really someone needed to consult them one day.

## Obsolete mirroring and import rules

There are lots of custom rules to deal with mirrors and imports. I suspect many if not most of them should no longer be necessary especially after the switch to base merging. We should review those rules to check:

1) if they are used at all; if they are not, they should be deleted;
2) if they are used, then are they doing something useful? Are they _supposed_ to be used? If no, again they should be deleted.

## Obsolete rules related to Travis

We are no longer using Travis, so there’s no reason to keep Travis-specific rules around. We should whether those rules are still used as part of the GitHub-based CI checks, and if they are not, delete them.

## Use of OWLTOOLS when a ROBOT equivalent already exists
There are rules that make use of OWLTOOLS for tasks that ROBOT is completely able to perform. For example:

```Make
$(TMPDIR)/unreasoned.owl: $(OWLSRC) $(BRIDGEDIR)/uberon-bridge-to-bfo.owl
        $(OWLTOOLS) $^ --merge-support-ontologies -o -f functional $@
```

This is merely equivalent to

```Make
$(TMPDIR)/unreasoned.owl: $(OWLSRC) $(BRIDGEDIR)/uberon-bridge-to-bfo.owl
        $(ROBOT) merge -i $< -i $(BRIDGEDIR)/uberon-bridge-to-bfo.owl --collapse-import-closure false -o $@
```

## Merging rules that no longer need to be kept separate
In the previous example, the `$(TMPDIR)/unreasoned.owl` target is only used once elsewhere in the Makefile, as a dependency to the following rule:

```Make
$(TMPDIR)/materialized.owl: $(TMPDIR)/unreasoned.owl $(TMP_REFL)
        $(ROBOT) merge -i $< -i $(DEVELOPS_FROM_CHAIN) --collapse-import-closure false \
                relax \
                materialize -T $(CONFIGDIR)/basic_properties.txt -r elk \
                reason -r elk --exclude-duplicate-axioms true --equivalent-classes-allowed asserted-only \
                unmerge -i $(TMP_REFL) \
                unmerge -i $(DEVELOPS_FROM_CHAIN) \
                annotate -O $(URIBASE)/uberon/materialized.owl -V  $(RELEASE)/materialized.owl -o $@ 2>&1 > $@.LOG
```

If `$(TMPDIR)/unreasoned.owl` is now itself generated by a ROBOT command and is not used anywhere else, then there’s no need for two separate rules and even for two separate commands! It can all be done in a single rule with a single ROBOT pipeline:

```Make
$(TMPDIR)/materialized.owl: $(OWLSRC) $(BRIDGEDIR)/uberon-bridge-to-bfo.owl $(TMP_REFL)
        $(ROBOT) merge -i $< -i $(BRIDGEDIR)/uberon-bridge-to-bfo.owl -i $(DEVELOPS_FROM_CHAIN) --collapse-import-closure false \
                relax \
                materialize -T $(CONFIGDIR)/basic_properties.txt -r elk \
                reason -r elk --exclude-duplicate-axioms true --equivalent-classes-allowed asserted-only \
                unmerge -i $(TMP_REFL) \
                unmerge -i $(DEVELOPS_FROM_CHAIN) \
                annotate -O $(URIBASE)/uberon/materialized.owl -V  $(RELEASE)/materialized.owl -o $@ 2>&1 > $@.LOG
```

Reducing the number of times we spawn a new OWLTOOLS or ROBOT process reduces the number of times the ontology needs to be parsed from disk. In the case of Uberon I am willing to bet the gain speed wouldn’t be negligeable.

## Potentially obsolete workarounds
There are several instances where we work around problems coming from imported ontologies. Nothing inherently wrong with that (sometimes we don’t have any choice), but we should check whether those workarounds are still needed.

For example:

```Make
# TODO: We should fix that upstream
# They are using too strict relationships, which are here replaced by more general ones
$(TMPDIR)/fixed-zfa.obo: $(TMPDIR)/mirror-zfa.obo
        if [ $(MIR) = true ] && [ $(IMP) = true ]; then perl -npe 's@RO:0002488@RO:0002496@;s@RO:0002492@RO:0002497@' $< > $@; fi

mirror/zfa.owl: $(TMPDIR)/fixed-zfa.obo
        if [ $(MIR) = true ] && [ $(IMP) = true ]; then $(OWLTOOLS_NO_CAT) $< -o -f ofn $@; fi

$(TMPDIR)/fixed-ehdaa2.obo: $(TMPDIR)/mirror-ehdaa2.obo | $(SCRIPTSDIR)/obo-grep.pl $(SCRIPTSDIR)/fix-ehdaa2-stages.pl
        if [ $(MIR) = true ] && [ $(IMP) = true ]; then $(SCRIPTSDIR)/obo-grep.pl -r 'id: (EHDAA2|AEO)' $< | $(SCRIPTSDIR)/fix-ehdaa2-stages.pl | grep -v ^alt_id > $@; fi
```

## Pointless targets
Consider the following rule:

```# tmp/merged.owl is now the flattening of ext.owl
# TODO: do we need this intermediate step? Used for subsets
tmp/merged.owl: ext.owl
        $(OWLTOOLS) $< --merge-import-closure --set-ontology-id -v $(RELEASE)/$@ $(URIBASE)/uberon/$@ -o $(TMPDIR)/$@ &&\
        $(ROBOT) merge -i $(TMPDIR)/$@ --collapse-import-closure false annotate --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $@
```

First, it is again using OWLTOOLS where ROBOT would do just fine. But most importantly, the only point of that rule is to merge the imports. OK, this may be “used for subsets”, except that… at this point in the pipeline the imports have already been merged! They have been merged since the early steps of the pipeline. `tmp/merged.owl` is actually the same thing as `ext.owl` (modulo the differences caused by serialisation, e.g. different IDs for blank nodes). This entire step is useless: two invocations of OWLTOOLS/ROBOT, two loads of the same ontology… for nothing.

## Bottom line
Without re-writing the Makefile scratch, I think that, step-by-step, we can considerably improvide its readibility and therefore its maintainability, all the while making the pipeline faster (by removing needless steps), without introducing any changes in the pipeline’s outputs.

gouttegd:
This PR cleans up the custom `uberon.Makefile` to, among other things:

* purge commented out code;
* purge obsoleted comments;
* remove obsolete targets;
* replace Owltools by ROBOT in most cases where it is directly possible;
* re-arrange the Makefile in a somewhat logical fashion (regrouping rules that are about the same thing);
* add comments to explain what the hell we are doing.

Overall, the goal is to make the Makefile reasonably _readable_.

This PR does _not_ change any of the main products (e.g., `uberon`, `uberon-base`, `composite-metazoan`, `subsets/human-view`) in any meaningful way, beyond some serialisation-induced differences.

Reviewers are **strongly** advised to review the PR _commit by commit_, rather than trying to review the PR-wide diff on `uberon.Makefile`. Each commit has a commit message explaining what it does (and sometimes, _why_ it does it).

closes #2973


matentzn:
## Nico ongoing review

- [X] https://github.com/obophenotype/uberon/pull/3028/commits/b01a116b67910a3c7f1d2167159594766319167a (comments with real significance were left in, so all good)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/3b49bbd01de1ec36e78900c2f7e5e9e3c518a814 (no questionable removals)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/85192ee3d03269d57ec3d48f3bcb7dbb182976c1 (no issues)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/b65fe5b7dd54bc0e20658e727a193356654121d5 (no issues)
- [ ] https://github.com/obophenotype/uberon/pull/3028/commits/2d76a2840e6914193bc9c68c4842ed22ba207726

src/ontology/uberon.Makefile
```diff
@@ -98,250 +120,43 @@ $(OWLSRC): $(SRC) $(COMPONENTSDIR)/disjoint_union_over.ofn $(REPORTDIR)/$(SRC)-g
...
+# <https://github.com/obophenotype/uberon/issues/2381>
 DEVELOPS_FROM_CHAIN=$(COMPONENTSDIR)/develops-from-chains.owl
-# see https://github.com/obophenotype/uberon/issues/2381
-
-$(TMPDIR)/materialized.owl: $(TMPDIR)/unreasoned.owl $(TMP_REFL)
-	$(ROBOT) merge -i $< -i $(DEVELOPS_FROM_CHAIN) --collapse-import-closure false \
-		relax \
-		materialize -T $(CONFIGDIR)/basic_properties.txt -r elk \
-		reason -r elk --exclude-duplicate-axioms true --equivalent-classes-allowed asserted-only \
-		unmerge -i $(TMP_REFL) \
```
matentzn, line 172: As I am going commit by commit, I am now reviewing a commit that still has these. 

I assume without reading, that you decided that "since it was never merged, only unmerged, and no-one noticed for over 1 year, it must be that it does not do any real purpose, so we drop it".

src/ontology/uberon.Makefile
```diff
@@ -98,250 +120,43 @@ $(OWLSRC): $(SRC) $(COMPONENTSDIR)/disjoint_union_over.ofn $(REPORTDIR)/$(SRC)-g
...
-		unmerge -i $(DEVELOPS_FROM_CHAIN) \
-		annotate -O $(URIBASE)/uberon/materialized.owl -V  $(RELEASE)/materialized.owl -o $@ 2>&1 > $@.LOG
-.PRECIOUS: $(TMPDIR)/materialized.owl
-
-# Used directly by Bgee, see https://github.com/obophenotype/uberon/issues/1501
-ext.owl: $(TMPDIR)/materialized.owl
```
matentzn, line 178: Given the comment above "used directly by bgee", I would prefer to change this goal to 

```
ext.owl: uberon.owl
	$(ROBOT) annotate -i $< --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $@
```

instead of deleting it. Remember users have idiosyncratic pipelines, and they may have lost over time developers etc that can debug running pipelines -> if a pipeline breaks because we remove a (I agree, redundant) release artefact, this can cause churn and reduce trust. I think its a tiny compromise to leave it in as a copy of uberon.owl.

src/ontology/uberon.Makefile
gouttegd: That particular case is described in #2980. As explained here, the component of reflexive axioms has not been used (being unmerged at the end without ever being merged in the first place) since 2021. I opened the ticket precisely to check whether my cleaning up should try to restore the pre-2021 behaviour or confirm the current situation; Chris confirmed in that ticket that the component is of no use.

(I should have mentioned ticket #2980 in the commit message, my bad.)

src/ontology/uberon.Makefile
```diff
@@ -98,250 +120,43 @@ $(OWLSRC): $(SRC) $(COMPONENTSDIR)/disjoint_union_over.ofn $(REPORTDIR)/$(SRC)-g
...
-
-TMP_REFL=$(COMPONENTSDIR)/reflexivity_axioms.owl
+# Step 2: Reasoning.
+# For the temporary injection of property chains, see
+# <https://github.com/obophenotype/uberon/issues/2381>
 DEVELOPS_FROM_CHAIN=$(COMPONENTSDIR)/develops-from-chains.owl
-# see https://github.com/obophenotype/uberon/issues/2381
-
-$(TMPDIR)/materialized.owl: $(TMPDIR)/unreasoned.owl $(TMP_REFL)
-	$(ROBOT) merge -i $< -i $(DEVELOPS_FROM_CHAIN) --collapse-import-closure false \
```
matentzn, line 168: Given the `--collapse-import-closure false` in the makefile goal, it seems to me that the intention was that ext.owl (immediately dependent on materialized.owl) is supposed to be uberon.owl _but with imports merged in_. However, downloading ext.owl it seems that merging was indeed happening somewhere upstream, like `unreasoned.owl`.

src/ontology/uberon.Makefile
```diff
@@ -1004,784 +816,470 @@ subsets/life-stages-minimal.owl: subsets/merged-partonomy.owl
...
-#
-# setting --home-dir to . will cause the dictionary .aspell.en.pws  in this directory to be used
+subsets/metazoan-view.owl: basic.owl
+	cp $< $@
 
-# interactive
-ispellcheck: $(SRC)-ispellcheck
+# The first step is a simple "merge+reason", but it still requires
+# Owltools because ROBOT has no equivalent to the -x option to simply
+# ignore unsatisfiable classes without erroring out.
```
matentzn, line 936: Does not seem like a good reason to keep it, at least not reading it this way.. Since ROBOT reason is used in the next step, what is the reason to maintain this?

src/ontology/uberon.Makefile
```diff
@@ -1998,7 +1473,6 @@ normalise_release_serialisation_ofn:
 
 normalise_release_serialisation_rdfmxml:
```
matentzn, line 1474: `normalise_release`, `normalise_release_serialisation_rdfmxml` and `normalise_release_serialisation_ofn` are remnants of an older OWLAPI serialisation and have not been needed since ODK 1.4.0.

Given these release instructions: http://obophenotype.github.io/uberon/uberon-release/

the file called /Users/matentzn/ws/uberon/src/ontology/run-release.sh is obsolete and should also be deleted.

src/ontology/uberon.Makefile
gouttegd: I would rather recommend updating Uberon’s PURL configuration to redirect `http://purl.obolibrary.org/obo/ext.owl` to `uberon.owl`.

We have a URL redirection system, we might as well use it.

src/ontology/uberon.Makefile
gouttegd: Imports are merged right at the beginning of the (original) pipeline, during the “preprocessing” step:

```Makefile
$(OWLSRC): $(SRC) $(COMPONENTSDIR)/disjoint_union_over.ofn $(REPORTDIR)/$(SRC)-gocheck $(REPORTDIR)/$(SRC)-ic>
        @echo "STRONG WARNING: issues/contributor.owl needs to be manually updated."
        $(ROBOT) merge -i $< \
                        -i $(COMPONENTSDIR)/disjoint_union_over.ofn \
                        -i issues/contributor.owl \
                expand --no-expand-term http://purl.obolibrary.org/obo/RO_0002175 \
                        -o $@
```

Now, it’s entirely possible that this was not intentional, and only caused by the fact that whoever wrote this didn’t know that `robot merge` merges the import **by default**, unless `--collapse-import-closure` is explicitly set to _false_.

But whether this was the intention or not, I wrote this PR with the aim of _not changing the current behaviour_. If the current behaviour was not actually wanted, that can be fixed in a later PR ­– where it would actually be much easier to fix with a readable, maintainable Makefile!

src/ontology/uberon.Makefile
matentzn: This is an option, albeit a little bit less safe in case a pipeline would use github links instead of PURLs.. remember my position here is not to challenge the correctness, more to reduce churn for users - even those using our resources "the wrong way".

src/ontology/uberon.Makefile
gouttegd: I’ve left all those in the PURGATORY section at the end (which is basically where I left all the cruft that I didn’t know what to do with). Happy to make another pass in that section to drop those.

src/ontology/uberon.Makefile
gouttegd: My bad, the comment is not clear enough. `-x` does not simply _ignore_ the unsatisfiable classes, it actively _removes them_ – so `robot reason` does not ever see the problematic classes. If we removed the Owltools step here, and the subset happens to contains unsats, the entire step would fail.

src/ontology/uberon.Makefile
gouttegd: I strongly object to keeping cruft in the Makefile just because _maybe_, _somewhere_, it could be that _someone_ _may_ be relying on that cruft – not if I don’t have the slightest clue to tell me that it is indeed the case.

If this breaks someone’s pipeline, they can complain about it. Then they can explain to use why they use direct GitHub URLs, and if their explanation makes sense, we can add back the `ext` product.

Using direct GitHub URLs always exposes users to unnecessary trouble and it should be clearly stated that this is unsupported.

(This applies to us: I am not particularly happy about the fact that we clone the `developmental-stages-ontologies` repo as part of this Makefile; It would be much better if that repo made some proper releases that we could then just download as any other foreign ontology.)

gouttegd: When I said below that one aim of this PR was “not changing the current behaviour”, I was speaking of the officially released products published in the official way.

If we have a PURL redirection from `ext` to `uberon`, the officially released `ext` product would appear to anyone out there to be the same as it was before (except for its ontology IRI), so this wouldn’t change the “current behaviour”.

src/ontology/uberon.Makefile
gouttegd: > whoever wrote this didn’t know that robot merge merges the import by default, unless --collapse-import-closure is explicitly set to false.

To be fair to that person: ROBOT’s inline help is probably at fault here:
```sh
$ robot merge --help
usage: robot merge --input <file> --input <file> --output <file>
[…]
 -c,--collapse-import-closure <arg>   if true, the imports closure will be
                                      merged
```
With such a phrasing, I think anyone reading this would conclude that the default value for this option is **false**, _i.e._ that the imports closure is _not_ merged by default.

One needs to consult the [online help](http://robot.obolibrary.org/merge) to learn that it is the opposite.

src/ontology/uberon.Makefile
matentzn: > not if I don’t have the slightest clue to tell me that it is indeed the case.

The phrase "Used directly by Bgee, see https://github.com/obophenotype/uberon/issues/1501" should at least approximate your definition of "slightest clue". But let's not ride on it. This is a sort of 51/49 situation for me now, so since you feel reasonably strongly about it, I am inclined to just drop my objections.

src/ontology/uberon.Makefile
gouttegd: Arf, just saw your Slack message telling to ignore the comments you are making on the PR for the time being! :D My bad, I’ll let you comment in peace and I’ll reply once you are done!

src/ontology/uberon.Makefile
```diff
@@ -498,422 +332,408 @@ imports/local-%.owl: mirror/%.owl
...
 
-DISABLE= multiply-labeled-edge valid-id-space isa-incomplete ascii-check has-definition bad-pmid ontology-declaration-check referenced-id-syntax-check owl-axiom-check is-symmetric-check
-$(REPORTDIR)/%.obo-gocheck: %.obo $(TMPDIR)/GO.xrf_abbs | $(SCRIPTSDIR)/check-obo-for-standard-release.pl
-	$(SCRIPTSDIR)/check-obo-for-standard-release.pl --xref-abbs $(TMPDIR)/GO.xrf_abbs $(patsubst %,--disable-%,$(DISABLE)) $< > $@.tmp && mv $@.tmp $@
+# Similar to %-part-of-parents above, but limited to a view
+reports/part-%-report.tsv: subsets/%-view.owl
+	$(OWLTOOLS) $< --reasoner mexr --export-parents -p BFO:0000050 -o $@.tmp && mv $@.tmp $@
 
-$(TMPDIR)/GO.xrf_abbs: $(SRC)
-	wget http://geneontology.org/doc/GO.xrf_abbs -O $@ && touch $@
```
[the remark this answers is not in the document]
gouttegd: The GO folks later [restored the URL](https://github.com/geneontology/go-ontology/issues/25854) just to make Uberon happy (nice!), but in the meantime Chris [has explicitly requested](https://github.com/obophenotype/uberon/issues/2999) that we stop using that file.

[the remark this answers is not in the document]
matentzn: OK!

src/ontology/bridge/collected-drosophila.owl
```diff
@@ -32,6 +32,6 @@
     </rdfs:comment>
     <owl:imports rdf:resource="http://purl.obolibrary.org/obo/uberon.owl"/>
     <owl:imports rdf:resource="http://purl.obolibrary.org/obo/uberon/bridge/uberon-bridge-to-fbbt.owl"/>
-    <owl:imports rdf:resource="http://purl.obolibrary.org/obo/uberon/fbbt-nd.owl"/>
+    <owl:imports rdf:resource="http://purl.obolibrary.org/obo/fbbt.owl"/>
```
matentzn, line 35: > "which we update here to use the imports/local-fbbt.owl mirror instead."

This is not immediately clear from this PR here - is this because the catalog rewires http://purl.obolibrary.org/obo/fbbt.owl to imports/local-fbbt.owl?

gouttegd: Yes – as for all the other foreign ontologies that are imported in the various `collected-*.owl` files. The pattern is always the same: the `collected-*.owl` file mentions the foreign ontology by its “normal” IRI, and the catalog redirects it to the `imports/local-*.owl` copy.

`collected-drosophila.owl` was the only collected ontology to deviate from that pattern, because it imported that `fbbt-nd.owl` instead. So here I just made `collected-drosophila.owl` follow the same pattern as all the others.

(This pattern is one of the thing that I will amend when I’ll fix the composite pipeline, by the way. I think it is needlessly confusing, because you need to look up into the catalog to actually know _which_ file is being used. But I didn’t want to do that just yet, not in this PR which is, err, long enough already?)

matentzn: Thanks!

src/ontology/uberon.Makefile
matentzn: Lets roll with this, as it is the status quo!

src/ontology/uberon.Makefile
matentzn: resolving in favour of https://github.com/obophenotype/uberon/issues/3031

matentzn (approved):
Minor comment left, please resolve as you see fit.

Monumental work, looks fantastic. Just reviewing it took me a bit more than 3 hours, I cannot imagine how many dozens of hours it must have taking to make that PR!


## Nico commit by commit review (just bookkeeping)

If the commit is checked, it means I have finished reviewing it (not that I agree with all of it).

- [X] https://github.com/obophenotype/uberon/pull/3028/commits/b01a116b67910a3c7f1d2167159594766319167a (comments with real significance were left in, so all good)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/3b49bbd01de1ec36e78900c2f7e5e9e3c518a814 (no questionable removals)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/85192ee3d03269d57ec3d48f3bcb7dbb182976c1 (no issues)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/b65fe5b7dd54bc0e20658e727a193356654121d5 (no issues)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/2d76a2840e6914193bc9c68c4842ed22ba207726 (some minor issues, including reservations about removing ext.owl, and some further opportunities of removing noise from uberon.Makefile)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/e50dd8854a475241ddc3e7a32db1a608b96e1eea (Damien made a corresponding issue, all good)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/e6e1b17b8048979502b010bda5eb4e40d9355686 (no changes, just adding a FIXME)
- [X]  https://github.com/obophenotype/uberon/pull/3028/commits/4b462a5c186dc2b02afbfe0e59670ee5404fd674 (removing uberon.yaml, which has never been used)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/87e5cdb0e691f93ea5d5b118933a58ef7fce690f (removing redundant uberon-base, all good)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/0d13d6460070a60f55e1fd19d4935cff848bf1a0 (removing custom envo import; in any case, the brokenness was fixed a while back)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/9cee90b5880e396a3f00937ae01b5b5a2f36f404 (this is the removal of ext, which I initially opposed, but coming around to now, see https://github.com/obophenotype/uberon/issues/3031)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/7f143f1da9d549ca9862281428dfce3c9bfd034f (RO MIRROR special rule, has FIXME and associated ticket. All good).
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/dd30ff3822f8d3f1f77c08c744b2997a0dfde2f0 (I didn't check that ZFA fixed all these relationships, but I trust Damien has checked it)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/235ccb588402cd29f38320af7a99894ebcb4082d (minor source of possible risk, made comment, but no need to think of it).
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/c6d75ea6b74df06bcb945debdc7cab9ec962f04f (its possible I made that mistake during the migration, and perhaps ehdaa2-local was supposed to depend on it, but, oh well).
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/435fe2ef5c51dda579dcad5c0b1edeea15cc1de6 (good)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/21fefc4c1cfbcc2d67fc8427b34b87cbeadfae44 (fantastic)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/fd37d153afd3d7ef86ab29f5957dd4da8aa9e3b0 (removing import rules, not needed)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/64da3e916df4bc82c3ed281b753c5719984f6628 (non-functional reordering, these can probably be removed at some point)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/094e00eb36cb7f304dbb5d64e4e37f2b8252654f (I spot checked the most important reports, (allcycles, gocheck),  and it seems there are no functional changes. Only reordering).
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/123aa4cf77e027236166b7a24b8f4fb5d60591f3 (mild concern, made a comment on makefile. Seems that GO file _does_ exist?)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/ff6450c322828ad6cee488cbe943c347ade96166 (this was a bigger commit, and I can't confidently say I have reviewed it line by line; but I reviewed the goals that I found most important, like `$(TMPDIR)/cl-core.obo` and `$(TMPDIR)/seed.obo`).
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/9630664e6ea9d004ac8da0b7374949e37b76027b (emap overdue removal, ok)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/62840c48bf458c0208db4900ede8c1d751481dfd (redundant rule removal from ancient times, ok)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/10b87bb43b9bced0b569548c0700d64e7ab1e32f (no functional changes, all good)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/2d1747c332bbc958c6404c10cb4dfd2bb26fac14 (beautiful reorder)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/775d8245e7b1ab56b3a4819e251d5ef803a016e9 (AWESOME! - I never really stuck long enough here to get how these checks were supposed to work)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/237020924595da21cc6e680c41d9d24c5942b72a (ok, learned something new, that I should not be using `patsub` in `make` for simple for loops)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/3751a0070fd5ad5fc871efdb9617ef99f45d1c05 (clear rewriting of bfo check. nice)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/3f6f476e0cd2780c2972ddfa689ecf261edb6cdb (Cleanup)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/0d18f9d6326adf6483ee15a0013b540ad9b07b21 (removing redundant rule, ok)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/bf30ae0ba1dceb65c8458ce3f8e6650ae2b38500 (nice simplification of developmental stage ontology use, no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/ea39a1d33b7c7426bc913d93341f386dccd7ac38 (simple cleanup, no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/c95e31333932ff133c273efb54759c7243a123c8 (reshuffling. some non functional changes, like removing uberon.owl from MBASE which is already covered by ext-weak. LGTM)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/e9fdef5cf44af7764e78b82956f9a31780a5250c (small comment about redirection of import)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/364f31a4414a625501d2573f3e16de809e72c8ff (no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/f1e3c066a30bab78a3ebb56cca7943d2b31757a1 (no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/8bc44b966f704901b162080622e8172aa662442d (no concerns, not sure what this is and if it is effectual in any way)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/327d19895c1a77f18b67fa477ece75ad7d5c7d89 (no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/827b849eba03be55887f151a3113a6ceafd6c54e (no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/6b335f4e541f93080dbe6cd87cd06697be1d89be (no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/d7bcf705d2ccd2c620d12b09cd119364417c576c (great to see the test goals all grouped together. It would be nice to simplify this moving forward).
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/69c5a08b387cd1d963a853f513100c1cb1a061a7 (no concerns, just comments)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/9b5fc7fa13049994c27aa7fee488f492b6b9cbcc (made a separate issue #3037)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/0bba7b4af47e2083255e6d4f81db1d0e966a8e9f (brave, and correct decision IMO)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/2cae55acf4334a745cb767af930e1254636b9b33 (no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/04e2c1e5c2f330aed4d728bb5839324af537f16b (no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/6e5e3c591a0f2bd49167300e2f19126bf96199ec (no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/9e7c84cbb49db7259611030dc2420ceaa3e7ac7c (I feel no offense that my hacks have been moved to PURGATORY)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/7cee50eb6cc780aa090770ddabb8e32cd9feec18 (no concerns)
- [X] https://github.com/obophenotype/uberon/pull/3028/commits/a719af2c18452c618e8973d4e60a2626dd3c4076 (I don't like this but nothing related to PR)

PR #3029, opened 2023-08-19
Undo using dev in QC and change "see also" in synonym annotation to xref

### LINKED ISSUE #3025: Revert ODK image to latest used QC GH Action
Because some ongoing changes are not yet available in the latest ODK image, I updated the container to use the dev version, which is unstable.

Revert the commit https://github.com/obophenotype/uberon/pull/3021/commits/5f9733c4c592f2711400363c39f095341d48698d

matentzn:
Fixes #3025 

This commit undoes the evil (my bad) which is switching to an unstable development snapshot for dev.

Sorry about the noise all, my bad.

@bvarner-ebi it is important that we do not use axiom annotations so freely; for synonyms, we consistently use oio:hasDbXref (even with URLs). If the URL is not evidence that the term can be used as a synonym, do not add it at all. 

@anitacaron the time has come to not listen to my technical advice too much anymore; switching to dev for QC was a wrong call by me, sorry about that! For the future, the only suggestion I do make is that, PRs should separate concerns: as @gouttegd suggests, we should at all times keep PRs to logical units that are easily reviewable, and have meaningful title like ("Using unstable development snapshot for QC (temp fix)" so that the white knights of software development have a chance to intervene when they see it.

### DIFF
--- a/.github/workflows/qc.yml
+++ b/.github/workflows/qc.yml
@@ -19,7 +19,7 @@ jobs:
   ontology_qc:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
-    container: obolibrary/odkfull:dev
+    container: obolibrary/odkfull:1.4.1
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
--- a/.github/workflows/qc.yml
+++ b/.github/workflows/qc.yml
@@ -19,7 +19,7 @@ jobs:
   ontology_qc:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
-    container: obolibrary/odkfull:1.4.1
+    container: obolibrary/odkfull:v1.4.1
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:


gouttegd (approved):

bvarner-ebi:
> we consistently use oio:hasDbXref (even with URLs)

@matentzn, is this true for all annotations or just on synonyms?

Should [this edit to the documentation](https://github.com/obophenotype/cell-ontology/pull/2109/files) be updated to state **use oio:hasDbXref (NOT seeALSO) for all cross-references that are URLs**?
I will keep the caveat that URLs should be avoided when a more stable reference can be used.

gouttegd:
@bvarner-ebi : Basically, any annotation that ends up being written like this in the -edit.obo file:

```
{http://www.w3.org/2000/01/rdf-schema#seeAlso="https://meshb.nlm.nih.gov/record/ui?ui=D009619"}
```

(that is, with a full URL instead of just a keyword on the left side of the `=` sign) should be avoided for now.

It’s a current limitation of the OBO format.

I think this applies to all _axiom_ annotations. Annotations on _terms_ are not affected.

matentzn:
> Should [this edit to the documentation](https://github.com/obophenotype/cell-ontology/pull/2109/files) be updated to state use oio:hasDbXref (NOT seeALSO) for all cross-references that are URLs?

This restriction only applies for _axiom annotations_, but yes, please only use "oboInOwl:hasDbXref" on synonyms. Don't worry about them being URLs - it is not a disaster. 

The rules are a bit undefined for this: Usually, we use _only_ "oboInOwl:hasDbXref" on synonyms. For everything else, you can use "oboInOwl:source" for the seeAlso kind of use cases to denote that this peace of information has been sourced from some website. But, I am just the messenger. This is just some historical stuff - we should initiate a separate process to make that clearer.

In the meantime, @anitacaron @bvarner-ebi can you please approve this?

bvarner-ebi:
Thank you for the context, @gouttegd and @matentzn.

Would either of you mind updating [the UBERON documentation](https://github.com/obophenotype/uberon/blob/master/docs/uberon-editor-sop.md), and if necessary, [the CL documentation ](https://github.com/obophenotype/cell-ontology/blob/master/docs/editing_guidelines.md) to reflect the current technical standards? Or, provide instructions verbatim here and I will edit the documentation.

gouttegd:
> Don't worry about them being URLs - it is not a disaster.

Well, it’s only a “disaster” (in that it breaks our workflows) if the URL happens to contains a `#` character (as is the case for rdfs:seeAlso). Because then the annotation IRI ultimately becomes `http://www.geneontology.org/formats/oboInOwl#http://www.w3.org/2000/01/rdf-schema#seeAlso`, and the fact that it contains _two_ `#` characters is what breaks the RDF parser.

But even annotation properties without a `#` in their IRI are problematic. This one, for example:

```
{http://purl.obolibrary.org/obo/IAO_0000232="A secretion from the glands of the male reproductive system in some aquatic animals that contains sperm and is sprayed onto the eggs."
```

It yields an annotation IRI like this: `http://www.geneontology.org/formats/oboInOwl#http://purl.obolibrary.org/obo/IAO_0000232`. It does not _break_ anything, but it is still incorrect.

bvarner-ebi (approved):
Approve with the following caveat:
The cross-reference for the text definition would also ideally include [the URL](https://meshb.nlm.nih.gov/record/ui?ui=D009619.). The URL directs to the page used to inspire the Uberon text definition, whereas MESH:D009619 points to a general webpage for the concept.

matentzn:
> Don't worry about them being URLs - it is not a disaster.

Dont worry about **the value of the oboInOwl:hasDbXref** being a URL, I meant. A URL as the property _is_ a disaster!

@bvarner-ebi 

For now, just make a PR saying this:

### Rules for axiom annotations:

- When providing evidence on synonyms and definitions, only use "oboInOwl:hasDbXref"
- For everything else (not synonyms!), you can use "oboInOwl:source" to denote that this assertion has been sourced from some website

anitacaron (approved):

bvarner-ebi:
> For now, just make a PR saying this:
> 
> ### Rules for axiom annotations:
> * When providing evidence on synonyms and definitions, only use "oboInOwl:hasDbXref"
> * For everything else (not synonyms!), you can use "oboInOwl:source" to denote that this assertion has been sourced from some website

Thank you. Is this applicable to both Uberon and CL?

matentzn:
> Thank you. Is this applicable to both Uberon and CL?

Lets say, yes.

PR #3048, opened 2023-08-31
NTR: saphenous vein smooth muscle tissue

### LINKED ISSUE #2993: NTR: saphenous vein smooth muscle tissue
**Preferred term label:** saphenous vein smooth muscle tissue

**Definition (free text, please give PubMed ID)** A portion of smooth muscle tissue that is part of a saphenous vein.
https://orcid.org/0000-0002-0819-0473

smooth muscle tissue UBERON_0001135
'part of' some saphenous vein UBERON_0007318

https://orcid.org/0000-0002-0819-0473

paulacalipho:
fixes #2993


src/ontology/uberon-edit.obo
```diff
@@ -224207,6 +224207,15 @@ relationship: dc-contributor https://orcid.org/0000-0002-0819-0473
 relationship: located_in UBERON:0002371 ! bone marrow
 property_value: dcterms-date "2023-06-26T13:48:30Z" xsd:dateTime
 
+[Term]
+id: UBERON:8480068
+name: saphenous vein smooth muscle tissue
+def: "A portion of smooth muscle tissue that is part of a saphenous vein." [https://orcid.org/0000-0002-0819-0473]
```
bvarner-ebi, line 224213: @paulacalipho, per line 103 of https://github.com/obophenotype/uberon/edit/master/docs/uberon-editor-sop.md, ORCIDs should be entered in the IRI Editor window instead of the Literal window in Protégé. Can you enter your ORCID the same way you did for the contributor annotation?

bvarner-ebi (changes requested):
Please see inline comment.
Otherwise looks OK.

src/ontology/uberon-edit.obo
bvarner-ebi: The annotation property should still be database_cross_reference (not editor) and the ORCID would be entered in the IRI Editor screen.

bvarner-ebi: @paulacalipho, I checked and it looks like the values in database_cross_reference move from IRI to the literal after saving and reopening, so it does not seem to make a difference. It may be because of the OBO format.

Regardless, thank you for the edits, and I will approve.

bvarner-ebi (approved):

paulacalipho:
@bvarner-ebi Many thanks

Issue #3043, opened 2023-08-29
[NTR] neuroendocrine system


aleixpuigb:
**Preferred term label:**
neuroendocrine system

**Synonyms**
NES (related, abbreviation)
PMID:15481802

**Definition (free text, please give PubMed ID)**
A regulatory system of the body that integrates the nervous system and the endocrine system. This system consists of a set of cells organized in single organs and diffuse elements, sharing co-production of amine hormone/transmitters, peptide hormone/transmitters and specific markers of neural determination.
PMID:15481802
[Wikipedia:endocrinology](https://en.wikipedia.org/wiki/Neuroendocrinology)

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
overlaps some 'endocrine system'
overlaps some 'nervous system'
'part of' some '[anatomical system](https://www.ebi.ac.uk/ols4/ontologies/uberon/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FUBERON_0000467?lang=en)'

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0001-6677-8489

**Comment**
In vertebrates, the hypothalamic-pituitary-target organ axis (H-P axis), the autonomic nervous system (ANS) and the diffuse neuroendocrine or APUD system contribute to the neuroendocrine system.
PMID:15481802

PR #3050, opened 2023-08-31
3049 Classify 'endocrine system' as a 'non-connected functional system'.

### LINKED ISSUE #3049: 'endocrine system' has no explicit parent
[endocrine system](http://purl.obolibrary.org/obo/UBERON_0000949) is “dangling”: it has no explicit parent and is therefore ultimately classified (after reasoning) as a [material entity](http://purl.obolibrary.org/obo/BFO_0000040).

This is because the removal of its former logical definition in #3039 not only removed the '[capable of](http://purl.obolibrary.org/obo/RO_0002215) some [endocrine process](http://purl.obolibrary.org/obo/GO_0050886)' (as was the intention), but also had the side-effect of removing [non-connected functional system](http://purl.obolibrary.org/obo/UBERON_0015203) as the only parent of [endocrine system](http://purl.obolibrary.org/obo/UBERON_0000949) (which was probably _not_ intended).

[endocrine system](http://purl.obolibrary.org/obo/UBERON_0000949) should be explicitly made a subclass of [non-connected functional system](http://purl.obolibrary.org/obo/UBERON_0015203).

gouttegd:
'endocrine system' is currently a dangling term with no explicit parent, so we (re-)classify it as a 'non-connected functional system' (as it used to be before the removal of its former logical definition).

closes #3049


aleixpuigb (approved):
Thank you for spotting this, @gouttegd.

src/ontology/uberon-edit.obo
```diff
@@ -8749,6 +8749,7 @@ xref: VHOG:0000098
 xref: Wikipedia:Endocrine_system
 xref: XAO:0000158
 xref: ZFA:0001158
+is_a: UBERON:0015203 ! non-connected functional system
```
bvarner-ebi, line 8752: In #3043, the 'neuroendocrine system' is classified as an 'anatomical system', which has a text def asserting it to be a "connected anatomical structure".

The definition of the PATO class [disconnected](http://purl.obolibrary.org/obo/PATO_0010001) is "A structural quality inhering in the bearer by virtue of the bearer consisting of multiple structures lacking any physical connection to each other."

I'm not sure if the above means at a single instant in time. If not, one may argue that physically distant structures are connected over time via the release and binding of hormones, neurotransmitters, etc.

Either way, 'endocrine system' and neuroendocrine system' should be modelled similarly.

gouttegd: I agree that however “endocrine system” is classified, “neuroendocrine system” should be classified similarly.

FWIW, I think interpreting “connected” as “connected over space OR time” is a bit of a stretch, especially since there is nothing in the definition of all the related terms that could suggest such a meaning. For example, [anatomical structure](http://purl.obolibrary.org/obo/UBERON_0000061) (which in Uberon is the root class for all things connected – including [anatomical system](http://purl.obolibrary.org/obo/UBERON_0000467)) is defined as “material anatomical entity that is a single connected structure with inherent 3D shape” – I don’t see how this leaves room for involving time.

In any case, the connected vs disconnected issue has been a regular source of unsatisfiability (especially when merging Uberon with other ontologies, because some other ontologies not only have a distinction between connected and disconnected but also _enforce_ that distinction through disjointness axioms, something that Uberon does not), so any change to the meaning of “(dis)connected” should be discussed carefully in a dedicated ticket.

This PR merely restores “endocrine system” as a “non-connected functional system”, as it was until last week.

bvarner-ebi (approved):

PR #3052, opened 2023-09-04
3051 Fix 'insect pharynx' bogus ID and add ID format check

### LINKED ISSUE #3051: "insect pharynx" has bogus ID
The recently added term '[insect pharynx](http://purl.obolibrary.org/obo/UBERON_60005380)' has an ID with 8-digits (UBERON:60005380) instead of the Uberon standard of 7 digits.

The term needs to be obsoleted for technical reason and replaced with a new term with the correct ID of UBERON:6005380.

Need a QC check to flag similar bogus IDs in the future.

gouttegd:
This PR fixes the bogus ID of 'insect pharynx' (8-digits ID instead of the standard 7-digits format) by obsoleting the original term and replacing it with a new term with a correctly minted ID.

It also adds a new QC check to help prevent any occurrence of similar bogus IDs in the future.

closes #3051

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -42,7 +42,7 @@ REPORT_FAIL_ON =            ERROR
 REPORT_LABEL =              
 REPORT_PROFILE_OPTS =       --profile $(ROBOT_PROFILE)
 OBO_FORMAT_OPTIONS =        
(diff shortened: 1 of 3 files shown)


matentzn (approved):
Looks great!!

anitacaron (approved):

PR #3060, opened 2023-09-07
Fixes #2417 make neck general

### LINKED ISSUE #2417: Make Neck more generalizable
Currently Fish cannot have a neck because of taxon constraints 

@cmungall commented (https://github.com/obophenotype/uberon/issues/2367#issuecomment-1096815312): 
"I think the best solution would be to generalize neck in a way that keeps
the fish people happy. Fish may lack a neck proper but they have an
analogous region of the craniocervucal region with the same skeletal
elements"

For now, we are making a taxon GCI (https://github.com/obophenotype/uberon/pull/2415) but a better solution is to make neck more generalizable like the above

I'm assigning myself for now so that it doesn't got unassigned, but I would like someone who is more knowledgeable about fish stuff to take over this ticket (not sure who, but if anyone knows, please do assign them/let me know)

aleixpuigb:
Fixes #2417 make neck general


src/ontology/uberon-edit.obo
```diff
@@ -9394,7 +9394,6 @@ xref: Wikipedia:Neck
 is_a: UBERON:0000475 ! organism subdivision
 relationship: anteriorly_connected_to UBERON:0000033 ! head
 relationship: dubious_for_taxon NCBITaxon:32443 {source="ISBN:9780674021839"} ! Teleostei
-relationship: only_in_taxon NCBITaxon:8287 ! Sarcopterygii
 relationship: part_of UBERON:0000153 ! anterior region of body
 relationship: part_of UBERON:0007811 ! craniocervical region
```
bvarner-ebi, line 9398: Given [this comment](https://github.com/obophenotype/uberon/issues/2417#issuecomment-1169217139), should `relationship: part_of UBERON:0007811 ! craniocervical region` be removed?

bvarner-ebi (approved):

PR #3071, opened 2023-09-18
Import ORCIDIO

anitacaron:
Documentation about [ORCIDIO](https://github.com/cthoyt/orcidio)


.gitignore
```diff
@@ -215,4 +215,5 @@ src/ontology/subsets/*.json
 src/ontology/subsets/*.obo
 src/ontology/subsets/*.owl
 src/ontology/subsets/*.tsv
-src/ontology/imports/*_import.owl
\ No newline at end of file
+src/ontology/imports/*_import.owl
+src/ontology/imports/orcidio_terms.txt
```
gouttegd: I don’t think that file should be excluded from the repository. The ODK import system is assuming that for a given *IMPORT*, the *$(IMPORT)_terms.txt* file already exists (since it is manually maintained), so there’s no rule to create it. So it should be committed even if it is empty.

gouttegd (changes requested):
Looks good, except for the exclusion of the `src/ontology/imports/orcidio_terms.txt`.

.gitignore
gouttegd: Arf, I was too slow in submitting my review, sorry about that!

gouttegd (approved):
Looks all good to me.

PR #3085, opened 2023-10-03
Serialisation protégé 5.6.3

aleixpuigb:
Serialisation protégé 5.6.3


anitacaron:
This is blocked by #3087

src/ontology/uberon-edit.obo
```diff
@@ -10379,7 +10379,7 @@ intersection_of: composed_primarily_of CHEBI:17029 ! chitin
 relationship: develops_from UBERON:0007376 ! outer epithelium
 relationship: only_in_taxon NCBITaxon:6656 ! Arthropoda
 relationship: part_of UBERON:0002416 ! integumental system
-property_value: editor_note "TODO - split from exoskeleton? Also: See above sf item to see if this belongs in GO" xsd:string
```
anitacaron, line 10382: The QC is failing because of this unexpected change. I don't know what could cause this. When I open Uberon in Protégé, I see this "new" annotation.

There are only a few cases:
Screenshot 2023-10-13 at 10 01 55

Screenshot 2023-10-13 at 10 01 42

anitacaron: @matentzn @gouttegd @balhoff

matentzn: Maybe: editors note were previously not included when converting an obo file to owl format. Solution:

- add "editors note" as permitted property
- remove editors note with robot remove during preprocessing

gouttegd: I’d like to be sure we understand what’s happening before hiding the problem under the rug…

We have 910 `editor_note` annotations on classes in Uberon (not counting the annotations on axioms).

They are correctly preserved in the -edit file when switching the serialisation format to the one used in OWL API 4.5.26 (whether the switch is done with Protégé or with `robot convert`). That is, all instances of `property_note: editor_note "..."` in `uberon-edit.obo` are preserved.

But:

When preprocessing the -edit file to produce `tmp/uberon-preprocessed.owl` (in functional syntax), only 890 of those annotations are correctly rendered as `http://purl.obolibrary.org/obo/IAO_0000116`, as in:

```
AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000116> <http://purl.obolibrary.org/obo/UBERON_0000021> "Mammary glands […] here.")
```

but 25 are rendered as `http://www.geneontology.org/formats/oboInOwl#0000116`, as in:

```
AnnotationAssertion(<http://www.geneontology.org/formats/oboInOwl#0000116> <http://purl.obolibrary.org/obo/UBERON_0001001> "TODO - split from exoskeleton? Also: See above sf item to see if this belongs in GO")
```

This does not cause issues to save the file in functional syntax, but does prevent saving the file in RDF/XML, because the XML serialiser tries to write that as `<oboInOwl:0000116>`, and IIRC XML names cannot start with a number.

Now what I’d like to know is _why_ those 25 annotations are rendered incorrectly???

gouttegd: It becomes interesting… We also have one `IAO:0000116` annotation on an axiom:

```
[Term]
id: UBERON:0003674
[…]
relationship: has_part UBERON:0006844 {IAO:0000116="cardinality not allowed with transitivity in OWL", count="1"}
```

That annotation is _also_ rendered incorrectly as `http://www.geneontology.org/formats/oboInOwl#0000116`. And it seems to be the cause of the entire issue: removing that axiom annotation makes all the other `editor_note` annotations on classes to be correctly rendered as `http://purl.obolibrary.org/obo/IAO_0000116`.

gouttegd: We should be able to use `IAO:0000116` in axiom annotations, as we actually have several other instances where this does not cause any problem, e.g.:

```
relationship: part_of UBERON:0001016 {IAO:0000116="we do not commit to CNS or PNS to allow for flexibility with this class"} ! nervous system
[…]
is_a: UBERON:0012112 {IAO:0000116="consider changing to derivation relationship", source="FMA"} ! ingested food
[…]
relationship: part_of UBERON:0000030 {IAO:0000116="check this"} ! lamina propria
```

All those annotations are correctly rendered as `http://purl.obolibrary.org/obo/IAO_0000116`. So I am unsure what the problem is with

```
relationship: has_part UBERON:0006844 {IAO:0000116="cardinality not allowed with transitivity in OWL", count="1"}
```

balhoff:
I'm not sure what's going wrong. But that annotation that's causing the problem has two annotations:
- `IAO:0000116="cardinality not allowed with transitivity in OWL"`
- `count="1"`

`count` is not declared as an annotation property in the edit file. Could you give it a `Typeref` and see if that helps?

src/ontology/uberon-edit.obo
balhoff: I'm not sure what's going wrong. But that annotation that's causing the problem has two annotations:
- `IAO:0000116="cardinality not allowed with transitivity in OWL"`
- `count="1"`

`count` is not declared as an annotation property in the edit file. Could you give it a `Typeref` and see if that helps?

gouttegd: No, it doesn’t change anything. And:
* the `IAO:0000116` annotation on UBERON:0003674 is incorrectly translated as `http://www.geneontology.org/formats/oboInOwl#0000116` even when the `count=1` annotation is absent;
* elsewhere in the ontology, we also have this:

```
relationship: has_part UBERON:0006844 {IAO:0000116="cardinality not allowed with transitivity in OWL", UBPROP:0000008="some premolars may have 3 cusps", min_count="2"} ! cusp of tooth
```

where the `IAO:0000116` annotation is correctly translated as `http://purl.obolibrary.org/obo/IAO_0000116` despite the fact that `min_count` is also not declared (and therefore translated under `oboInOwl`).

anitacaron:
We have discussed in the Tech meeting a workaround while we wait for the new releases of each tool.
In all cases, as an annotation or an annotation for a logical axiom, we'll change the annotation to `comment` instead of IAO:0000116.

matentzn (commented):
I dont have any larger concerns here (especially since the PR is reasonably sized) but I do think we should in the midterm distinction between comments that are meant for the public, and comments that are meant for the curators of the ontology. So I would suggest to add "EDITOR NOTE:commentcoommentcomment" so we can clearly distinguish those, and have an easier time later to align editors notes with OMO.

src/ontology/uberon-edit.obo
```diff
@@ -10362,6 +10362,7 @@ property_value: taxon_notes "Most vertebrates have some form of duct to transfer
 id: UBERON:0001001
 name: chitin-based cuticle
 def: "A cuticular covering that is composed primarily of chitin. The main structural component of arthropod cuticle is a polysaccharide, chitin, composed of N-acetylglucosamine units, together with proteins and lipids[WP]." [https://orcid.org/0000-0002-6601-2165, https://sourceforge.net/tracker/index.php?func=detail&aid=3348965&group_id=36855&atid=440764, Wikipedia:Arthropod_exoskeleton, Wikipedia:Cuticle#Invertebrate_zoology]
+comment: "TODO - split from exoskeleton? Also: See above sf item to see if this belongs in GO" xsd:string
```
gouttegd, line 10365: In OBO format the `comment` tag expects a **unquoted** string (contrary to `def`) without a type qualifier. Using quotes followed by a `xsd:string` type qualifier results in the parser interpreting the value of the comment to be **literally**

```
"TODO - split from exoskeleton? Also: See above sf item to see if this belongs in GO" xsd:string
```

as can be seen when converting the file to functional syntax:

```
AnnotationAssertion(rdfs:comment <http://purl.obolibrary.org/obo/UBERON_0001001> "\"TODO - split from exoskeleton? Also: See above sf item to see if this belongs in GO\" xsd:string")
```

gouttegd (changes requested):
There are several instances of ill-formatted `comment:` tags (see below for an example). The syntax of a `comment:` tag is

```
comment: this is a comment
```

and NOT

```
comment: "this is a comment" xsd:string
```

gouttegd:
And what about, for example:

```
property_value: editor_note "The anatomy, pathology and medical community considers the anus to be part of the large intestine, and supports the use of 'anorectum' to include rectum, anal canal and anus. Although a few resources seem to only include the distal part of the rectum in 'anorectum', this UBERON grouping term is meant to refer more broadly to the whole of rectum." xsd:string
```

That’s an editor note, but it sounds like something I would like to know if I was a user of Uberon.

```
property_value: editor_note "Modified from the original AAO source" xsd:string
```

Likewise. Why should the fact that the term comes a given source _but_ has been modified from that source should be a curator-only information?

```
[Term]
id: UBERON:2005260
name: fenestrated capillary
[…]
is_a: UBERON:0001982 {source="FMA"} ! capillary
property_value: editor_note "the ZFA class from which this was derived is not explicitly classified as a capillary - need to verify for taxon histology differences" xsd:string
```

So the editor classified the term as a capillary _but_ there is a lingering doubt about this classification. Sure, it’s something for the _editors_ to check, but that’s again something I’d definitely like to know as a mere user.

```
property_value: editor_note "A quick perusal of definitions for fascia dentata suggests that it is sometimes used to refer to the entire dentate gyrus, in which case it should be a synonym, and sometimes to a specific cytoarchitectural field of dentate gyrus. More research is needed. In the meanwhile, I have not assigned it as a part of dentate gyrus, assigning instead the cytoarchitectural layers. (Maryann Martone)" xsd:string
```

Yep, again something I’d like to know, thank you very much.

And as soon as you ask curators to decide, when they want to add a (comment|note), whether it is more intended to other curators or to the “public”, you’re bound to have inconsistencies, because curators won’t have the same idea of what is more useful for each group of people. As I said, needless complication. Let’s just have comments and let people figure out if the comment is useful for them or not.

matentzn:
You are arguing with "its hard to distinguish editors notes from comments" - You are not saying, correct me if I am wrong, that there are not any comments that are not meant for outside consumers.

Since an ontology is essentially a database like object, there should be a way to add annotations that are not meant for downstream consumption - I am certainly not wedded to "editors note", but there should be a way to say: "strip this annotation out before running a release". Would you like to direct the discussion more towards an axiom annotation that can be used on any annotation assertion?

gouttegd:
> You are not saying, correct me if I am wrong, that there are not any comments that are not meant for outside consumers.

Indeed, I did not say that – but I should have, and I am saying it now.

I don’t see why some comments should not be meant for “outside consumers“. We are doing everything in the open. Nothing should be “for insiders only“.

In fact for a free (as in free beer) project like Uberon I believe the distinction between “insiders“ and “outsiders“ is pointless, similarly to the distinction between “developers“ and “users“ in free software. Any “outsider” can become a contributor at anytime, even if only for one contribution (isn’t that what some people call “drive-by curation”?).

> there should be a way to add annotations that are not meant for downstream consumption

I disagree.

Having two annotations to represent “comments meant for outsiders/downstream users” and “comments meant for curators” only results in the users having to always check both types of annotations, because users can never be sure that a “curator-only comment” will not contain an information that is actually useful to them (as is the case for all the examples I have given). It’s a needless complication for both editors and users and a hindrance to drive-by curation.

It would be even worse if, as you seem to be suggesting, the “curator-only comments“ were stripped from the release products, because then users would be left completely ignorant of the existence of comments that may very well be useful to them, unless they are savvy enough to check the ontology source files instead of just using the released artefacts.

matentzn:
I never really thought this all the way through for comments, but there are certainly a lot of annotations in various editors' files that do not belong in a release (we have annotations that allow terms to violate a QC check for example, or we have annotations in one format that need to be transformed to another before release). I guess we should stop discussing this here, seems more like an OMO issue tracker issue. And we will lose all our nice argument after this is merged.

I am fine, if @anitacaron agrees with you, to leave it as is (without adding additional strings to indicate editors comments) in this case.

src/ontology/uberon-edit.obo
anitacaron: Good, I found other examples not part of the `editor notes` change.

gouttegd (approved):
No issues as far as I can tell.

No objection to adding a marker inside comments, though as I have stated I believe this is completely unnecessary.

matentzn (dismissed):
If you load - saved this with Protege LGTM

anitacaron:
Yes, I ran `normalize_src` every time.

gouttegd (changes requested):
There are still some non-oboInOwl annotations that could cause the same bug to be triggered again. One of them is highlighted below.

The other two that I found after a quick look are line 113231:

```
synonym: "milt" NARROW [PMID:31287811, PMID:35203131] {IAO:0000232="A secretion from the glands of the male reproductive system in some aquatic animals that contains sperm and is sprayed onto the eggs."}
```

and line 118703:

```
relationship: has_part UBERON:0006844 {UBPROP:0000008="some premolars may have 3 cusps", min_count="2", comment="cardinality not allowed with transitivity in OWL"} ! cusp of tooth
```

src/ontology/uberon-edit.obo
```diff
@@ -223935,7 +223935,7 @@ property_value: dcterms-date "2022-10-31T07:56:47Z" xsd:dateTime
 id: UBERON:8480041
 name: testicular sheath
 def: "Bilayer sheath that covers both the insect male mature gonad and a portion of genital disc-derived structures. It is composed of an outer pigment cell layer, an inner muscle cell layer, and their respective basal laminae." [PMID:9553045]
-synonym: "testis sheath" EXACT [] {http://purl.obolibrary.org/obo/IAO_0000232="https://www.fortunejournals.com/ijpaes/admin/php/uploads/798_pdf.pdf"}
+synonym: "testis sheath" EXACT [] {IAO:0000232="https://www.fortunejournals.com/ijpaes/admin/php/uploads/798_pdf.pdf"}
```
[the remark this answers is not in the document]
anitacaron: Ok, I'll change it to `comment`

gouttegd (approved):

src/ontology/uberon-edit.obo
```diff
@@ -118700,7 +118700,7 @@ xref: SCTID:304572005
 xref: UMLS:C1704302 {source="ncithesaurus:Bicuspid_Tooth"}
 xref: Wikipedia:Premolar
 is_a: UBERON:0013164 ! molariform tooth
-relationship: has_part UBERON:0006844 {editor_note="cardinality not allowed with transitivity in OWL", min_count="2", taxon_notes="some premolars may have 3 cusps"} ! cusp of tooth
+relationship: has_part UBERON:0006844 {min_count="2", comment="TAXON NOTE some premolars may have 3 cusps", comment="cardinality not allowed with transitivity in OWL"} ! cusp of tooth
```
gouttegd, line 118703: I have a lingering doubt about whether we are allowed to use more than one comment here. The OBO specification clearly states (section 4.4, [Tag Cardinality Constraints](http://owlcollab.github.io/oboformat/doc/obo-syntax.html#cardinality)) that we can only have one comment *in a Term frame*, but it says nothing about what’s permitted in trailing qualifier blocks.

But the OBO parser does not complain about the presence of two comments and the OBO writer is able to roundtrip the file without any hiccup, so I am going to assume this is perfectly fine.

gouttegd (approved):

src/ontology/uberon-edit.obo
anitacaron: Yes, I agree.

PR #3087, opened 2023-10-04
Update ODK image to v1.4.3 in QC

### LINKED ISSUE #3084: CI failure, maybe due to overlapping synonyms
In my recent PRs #3082 and #3083, there appear to be some QA check failures occurring. From how they look, it's because different terms are sharing exact synonyms:

```csv
entity,property,value
http://purl.obolibrary.org/obo/UBERON_0002633,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,nV
http://purl.obolibrary.org/obo/UBERON_2000438,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,phy
http://purl.obolibrary.org/obo/UBERON_2005093,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,NV
http://purl.obolibrary.org/obo/UBERON_8440028,http://www.geneontology.org/formats/oboInOwl#hasExactSynonym,PHY
```

This wasn't an issue in the last merge, so I'm guessing it has to do with updates to odk/robot. Can someone take a look at these and solve them?

anitacaron:
Fixes #3084


anitacaron:
QC failed with the message:

```
Exception in thread "elk-reasoner-thread-2" java.lang.OutOfMemoryError: Java heap space
make: *** [uberon.Makefile:128: uberon.owl] Error 1
```

I increased to 8GB the ENV `ROBOT_JAVA_ARGS` in the QC for testing.

cthoyt:
Looks like 8G wasn't enough

anitacaron:
We found out that the command `robot materialize` which is part of the rule that generates `uberon.owl` was causing the peak memory around 20G (on my macos machine).

We decided to remove this command when running QC because it will not bring unsats issues.

matentzn (approved):
Few changes, fantastic work!

src/ontology/uberon.Makefile
```diff
@@ -124,11 +124,15 @@ $(OWLSRC): $(SRC) $(COMPONENTSDIR)/disjoint_union_over.ofn $(REPORTDIR)/$(SRC)-g
 # For the temporary injection of property chains, see
 # <https://github.com/obophenotype/uberon/issues/2381>
 DEVELOPS_FROM_CHAIN=$(COMPONENTSDIR)/develops-from-chains.owl
+ifeq ($(GH_ACTION),true)
+MATERIALIZE=
+else
+MATERIALIZE = materialize -T $(CONFIGDIR)/basic_properties.txt -r elk
+endif
```
gouttegd, line 131: Could we add a small comment to give the rationale for this? No need for any extensive explanation, but something like:

```
# The 'materialize' step is too resource-intensive for the GitHub Actions runners.
# See <https://github.com/obophenotype/uberon/pull/3087/>
```

gouttegd: Oups, too late! ^^'

matentzn: just do it after the fact, i will approve, no need to run qc

PR #3092, opened 2023-10-13
Import OMO terms

### LINKED ISSUE #3058: Add OMO import to UBERON
Context: https://github.com/obophenotype/cell-ontology/issues/2045

Depends on importing OMO (@anitacaron).

Once OMO is imported, edit all plural term annotation properties to [OMO:0003004](http://purl.obolibrary.org/obo/OMO_0003004) 'plural form'.

anitacaron:
Fixes #3058

Updated `imports/merged_import.owl` with OMO:0003004, OMO:0003000 and OMO:0003011.


matentzn:
Ehm.. a bit too much.. the diff.

gouttegd:
The diff is huge because _many_ terms are removed from the import module (especially a bunch of GO terms). Not sure why.

It may have been a glitch of some sort, because I just tried running `refresh-imports` again and all those terms are back.

src/ontology/imports/omo_terms.txt
```diff
@@ -0,0 +1,3 @@
+OMO:0003004
+OMO:0003000
+
```
bvarner-ebi, line 3: Can OMO:0003011 be added?

anitacaron: Yes, I'll add it.

anitacaron:
@matentzn @gouttegd could you review, please?

gouttegd (changes requested):
I see no problems _in the final state_ of the PR. However I advise against merging it as it is.

In commit 2a10fb6 a lot of terms were removed from the `merged_import.owl` module; then in the more recent 3240ea7 commit, all those terms were added back. I am not sure why this happened, but there must have been a glitch in the import pipeline when the files that were committed in 2a10fb6 were generated.

So now the final state of `merged_import.owl` is good, but having two “opposite” commits (one that removes ~100K lines for nothing, and then a second that adds those ~100K lines back) would needlessly pollute the history and I advise against that.

Either:
* resubmit a clean PR without the “removal-then-reinsertion” dance, or
* at the very least make a “squash” merge.

gouttegd (approved):

anitacaron:
Yes, `squash and merge` is the default option when merging a PR.

anitacaron:
Thanks, Damien.

anitacaron:
I'm not sure.

matentzn:
This is configured at GitHub project level

gouttegd:
But then how comes that it does not apply to me? When I merge a PR on Uberon, for me the default option is always a “normal” merge, _not_ a “squash” one.

PR #3097, opened 2023-10-24
Make base for GO import

### LINKED ISSUE #3057: Change GO import to make base instead of use base
Having a close look into `go-base.owl` file we import and it comes with a long list of RO properties, all with their own annotations.

image

anitacaron:
~The artefact `go-base.owl` is in reality a base plus artefact (base + the inferred axioms). This reduces the size of the `merged_import.owl` by half.~

Fixes #3057

### DIFF
--- a/src/ontology/uberon-odk.yaml
+++ b/src/ontology/uberon-odk.yaml
@@ -44,7 +44,7 @@ import_group:
     - id: cl
       use_base: TRUE
     - id: go
-      use_base: TRUE
+      make_base: TRUE
       base_iris:
         - http://purl.obolibrary.org/obo/GO_
         - http://purl.obolibrary.org/obo/GOREL_
--- a/src/ontology/uberon-odk.yaml
+++ b/src/ontology/uberon-odk.yaml
@@ -45,6 +45,7 @@ import_group:
       use_base: TRUE
     - id: go
       make_base: TRUE
+      mirror_from: http://purl.obolibrary.org/obo/go/go-base.owl
       base_iris:
         - http://purl.obolibrary.org/obo/GO_
         - http://purl.obolibrary.org/obo/GOREL_
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -427,8 +427,10 @@ mirror-cl: | $(TMPDIR)
 .PHONY: mirror-go
 .PRECIOUS: $(MIRRORDIR)/go.owl
 mirror-go: | $(TMPDIR)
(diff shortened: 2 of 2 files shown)


src/ontology/uberon-odk.yaml
```diff
@@ -44,7 +44,7 @@ import_group:
     - id: cl
       use_base: TRUE
     - id: go
-      use_base: TRUE
+      make_base: TRUE
```
matentzn, line 47:
```suggestion
      make_base: TRUE
      mirror_from: http://purl.obolibrary.org/obo/go/go-base.owl
```

matentzn: unfortunately, you have not been properly introduced to the GO ontology structure. What is sold on http://purl.obolibrary.org/obo/go.owl is a drastically reduced version of GO which is lacking a lot of the rich links to other ontologies. So you need to add _mirror_from_ as well here. http://purl.obolibrary.org/obo/go/go-base.owl has _more_, not _less_ then http://purl.obolibrary.org/obo/go.owl.. Gr.. Sorry. 

In any case, hopefully, you will still get _some_reduction, just not as much as you are hoping :)

matentzn (changes requested):

matentzn (commented):
Before I confirm, lets wait and see how the import module looks like - QC should run over the correct version..

src/ontology/uberon-odk.yaml
anitacaron: but we don't need the link to other ontologies when importing GO, right?

matentzn: They drive inference! So those axioms will contribute to reasoning of terms involving GO terms (especially taxon stuff, but also many other things). So yes, conceptually at least we need them. Practically, no one knows.

anitacaron: yeah, I didn't see any difference for Uberon when using the `go.owl`, but that's fine.

src/ontology/imports/merged_import.owl
```diff
@@ -12250,35 +12253,26 @@ AnnotationAssertion(rdfs:isDefinedBy <http://purl.obolibrary.org/obo/IAO_0000232
 AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/IAO_0000232> "curator note"@en)
 AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/IAO_0000232> "curator note")
 
-# Annotation Property: <http://purl.obolibrary.org/obo/IAO_0000233> (term tracker item)
```
matentzn, line 12253: Seems a very risky change.

matentzn (approved):

Issue #3096, opened 2023-10-24
[NTR] pulmonary venule


aleixpuigb:
**Preferred term label:**
pulmonary venule

**Synonyms**

**Definition (free text, please give PubMed ID)**
A venule that is part of a lung and drains oxygenated blood from pulmonary capillaries.

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
venule
'part of' some lung
drains some 'pulmonary capillary'
overlaps some 'pulmonary interlobular septum'
Inferred: [lung blood vessel](https://www.ebi.ac.uk/ols4/ontologies/cl/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FUBERON_0003512?lang=en)

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0001-6677-8489

Additional axiom
'pulmonary interlobular septum' SubclassOf overlaps some 'pulmonary venule'

PR #3098, opened 2023-10-24
Addresses #3096 NTR pulmonary venule

aleixpuigb:
Addresses #3096 NTR pulmonary venule


src/ontology/uberon-edit.obo
```diff
@@ -224613,6 +224613,17 @@ relationship: overlaps UBERON:0018227 ! pulmonary lymphatic vessel
 relationship: surrounds UBERON:0010369 ! secondary pulmonary lobule
 property_value: dcterms-date "2023-10-23T08:52:12Z" xsd:dateTime
 
+[Term]
+id: UBERON:8600024
+name: pulmonary venule
+def: "A venule that is part of a lung and drains oxygenated blood from pulmonary capillaries. A pulmonary venule merges into a pulmonary vein branch that run in the pulmonary interlobular septa." [PMID:22283705, PMID:30247823]
```
bvarner-ebi, line 224619:
```suggestion
def: "A venule that is part of a lung and drains oxygenated blood from pulmonary capillaries. A pulmonary venule merges into a pulmonary vein branch that runs in the pulmonary interlobular septa." [PMID:22283705, PMID:30247823]
```

This sounds like a "pulmonary vein branch", not a pulmonary venule, passes through a pulmonary interlobular septum. Is that aligned with the comment [here](https://github.com/obophenotype/uberon/pull/3095#discussion_r1369754009)?

bvarner-ebi (changes requested):
See inline comment.
Also, with the current modelling, this class does not seem to be related to 'pulmonary venous system'. Consider if it should be.

src/ontology/uberon-edit.obo
aleixpuigb: You are correct, I have to edit interlobular septa accordingly.

aleixpuigb: A new ticket #3115 has been opened to correct this.

aleixpuigb:
> See inline comment. Also, with the current modelling, this class does not seem to be related to 'pulmonary venous system'. Consider if it should be.

Thank you for your suggestion, I have added 'part of' some 'pulmonary venous system'.

bvarner-ebi (approved):
No breaking changes found.
Defer content review to SME. After a cursory online review, there does not seem to be consensus on pulmonary venous vasculature terminology.

PR #3099, opened 2023-10-26
Fix composite unsats

gouttegd:
This PR fixes unsatisfiable classes recently found in composite-metazoan by:

1. removing a bogus axiom (likely coming from an old version of Uberon) in the local CTENO mirror;
2. removing a bogus axiom in the local EHDAA2 mirror;
3. removing the bogus mappings between Uberon’s `serous sac` and EMAPA:16060 and MA:0000005.

See individual commit messages for more details about each issue.

### DIFF
--- a/src/ontology/imports/local-cteno.owl
+++ b/src/ontology/imports/local-cteno.owl
@@ -12857,12 +12857,6 @@
     
 
 
-    <!-- http://purl.obolibrary.org/obo/UBERON_0006846 -->
-
-    <owl:Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0006846">
-        <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/UBERON_0003102"/>
-        <rdfs:label rdf:datatype="http://www.w3.org/2001/XMLSchema#string">surface groove</rdfs:label>
-    </owl:Class>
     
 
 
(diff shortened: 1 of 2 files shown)


matentzn (approved):
Wow how long did _that_ small looking PR take you! That looks like a massive detective hunt..

anitacaron (approved):

gouttegd:
* The CTENO issue: around 1h;
* The EHDAA2 issue: around 10min;
* The EMAPA/MA issue: around 6.022⨉10²³ hours, or so it seemed.

matentzn:
Amazing.. 😅

gouttegd:
While I am here, note to my future self and to anyone attempting to debug unsatisfiability issues in composite-metazoan (or any other composite-* product):

1. Don’t.
2. But if you must, then: avoid using `composite-metazoan.owl` itself or its direct precursor `$(TMPDIR)/unreasoned-composite-metazoan.owl` to chase down the cause(s) of an unsatisfiability issue. This is because the `--merge-species-ontology` and `--merge-equivalence-sets` step that are at the core of the composite pipeline may materialise axioms that were not explicitly asserted in any of the source ontologies and by doing that, it may make it impossible for any reasoner to explain to you the real root cause for an erroneous inferred assertion. Use instead an intermediate file that is just upstream of that step, `$(TMPDIR)/merged-composite-metazoan.owl`.

PR #3102, opened 2023-10-27
Enforce taxon constraints

gouttegd:
This PR is intended to supersede #2928. It applies cleanly to the master branch with a lean history.

It ensures that taxon constraints are enforced during the QC pipeline by:

1. importing the version of the NCBITaxon slim that contains disjointness axioms between taxon siblings;
2. expanding the [present in taxon](http://purl.obolibrary.org/obo/RO_0002175) macro during QC;
3. merging the full NCBITaxon slim (all taxa) during QC;
4. fixing all known existing violations of the taxon constraints, save for two that have an external cause and will need to be fixed upstream.


gouttegd:
The two remaining taxon constraint violations are caused by:

* [DMBA:16271](http://purl.obolibrary.org/obo/DMBA_16271) being mapped to [lateral tuberal nucleus](http://purl.obolibrary.org/obo/UBERON_0000435): the Uberon term is explicitly primate-specific while the DMBA term is for rodent. This comes from the (externally maintained) DMBA bridge and needs to be fixed [there](https://github.com/obophenotype/ABA_Uberon/issues/36) (or rather in [their new repository](https://github.com/brain-bican/developing_mouse_brain_atlas_ontology)).
* [tuberal nucleus (sensu Rodentia)](http://purl.obolibrary.org/obo/UBERON_8440041) asserted to be [present in taxon](http://purl.obolibrary.org/obo/RO_0002175) some [Homo sapiens](http://purl.obolibrary.org/obo/NCBITaxon_9606), even though the term is explicitly rodent-specific. This comes from the (externally maintained) HRA subset and needs to be [there](https://github.com/hubmapconsortium/ccf-validation-tools/issues/293).

Funnily enough, those two issues are inverted mirrors of each other. The mouse-specific resource is making a bogus assertion by mapping to the primate-specific term while the human-specific resource is making a bogus assertion on the rodent-specific term.

gouttegd:
I fixed the problem that we were only importing `taxslim-disjoint-over-in-taxon.owl` which (as pointed by @anitacaron in today’s tech call) contains _only_ the disjointness axioms between taxa and nothing else (no labels, definitions, subclass relationships, etc.). Now we import both `taxslim.owl` (for all the infos about taxa) and `taxslim-disjoint-over-in-taxon.owl` (for the disjointness axioms).

As discussed in the same call, we should consider changing the way we import the Allen bridges and the HRA subset (which are currently forcefully refreshed at every build) so that bogus assertions in those external resources do not end up causing all our QC runs to fail for reasons that are not ours to fix.

cmungall:
Thanks for all this!

Originally the SOT for all bridges was in uberon. Divesting ourselves of
this role is good social and community practice. Yet it leaves us open to
multi ontology incoherencies. I’m surprised this hasn’t come up for CL
before given we divested bridges much earlier. However the nature and
granularity of both linkages and TCs at the cell type level is likely
different

I guess the best strategy is cautious refresh and retain the right to
override? Of course, this is axiom injection, but not in the core ontology

gouttegd:
> I guess the best strategy is cautious refresh and retain the right to
override?

I think so. And by “cautious refresh”, I mean something like “no externally maintained component should ever be _automatically_ updated as part of either the release process or the QC process.” They should only be updated upon manual, explicit action from an editor.

No update at release time because during a release is not the time to fix any sudden problem caused by an external source. And no update at QC time because the point of the automated QC checks is to catch _internal_ problems (it’s to prevent Uberon editors from inadvertently introducing errors in a PR), they should not be caused to fail for something that happened elsewhere.

This is already what we do for imports: they are only refreshed when an editor explicitly runs `make refresh-imports`, _not_ at release time or QC time. We should generalise that for other externally maintained components, not just imports.

src/ontology/uberon-odk.yaml
```diff
@@ -64,6 +64,8 @@ import_group:
     - id: bfo
     - id: ncbitaxon
       mirror_from: http://purl.obolibrary.org/obo/ncbitaxon/subsets/taxslim.owl
+    - id: ncbitaxondisjoints
+      mirror_from: http://purl.obolibrary.org/obo/ncbitaxon/subsets/taxslim-disjoint-over-in-taxon.owl
```
matentzn, line 68: Are we certain we want these in our merged import?

https://github.com/obophenotype/uberon/blob/dc3aa6a28753639eb2b3efe9e433c02ede6c628e/src/ontology/Makefile#L534C1-L537C1

src/ontology/uberon.Makefile
```diff
@@ -636,18 +637,41 @@ $(TMPDIR)/external-disjoints.owl: components/external-disjoints.obo
...
+# be invoked by Make it would not result in the mirror being downloaded
+# if it is not already available.
+# So instead, we make the checks depend on another file, and here we
+# 1. force the download of the mirror if it is not already there;
+# 2. create a link to that mirror.
+$(TMPDIR)/taxslim-disjoint-over-in-taxon.owl:
+	if [ ! -f $(MIRRORDIR)/ncbitaxondisjoints.owl ]; then \
+		$(MAKE) $(MIRRORDIR)/ncbitaxondisjoints.owl MIR=true IMP=true ; \
+	fi && \
+	ln -f -s ../$(MIRRORDIR)/ncbitaxondisjoints.owl $@
```
matentzn, line 654: cool idea tp save disc space. Hopefully this works on windows systems!

matentzn (commented):
I have not reviewed uberon-edit.obo (not the right expertise, but FWIW lgtm).

My main concern is this line in the `Makefile`:

https://github.com/obophenotype/uberon/blob/dc3aa6a28753639eb2b3efe9e433c02ede6c628e/src/ontology/Makefile#L534C1-L537C1

Are we certain we want to bump merged.owl with the entire ncbitaxondisjoints.owl file? Wont that make memory consumption go wild?

src/ontology/uberon-odk.yaml
gouttegd: Well, we do want the disjointness axioms in the released artefacts, so we have to import them somewhere.

On my machine, `refresh-imports` went from ~9 GB of peak memory consumption without importing `taxslim-disjoint-over-in-taxon.owl` to ~10 GB when importing it, for no noticeable difference in runtime (~4 min). Is that wild?

matentzn: Nope, sounds ok to me! Thanks for checking!

src/ontology/uberon.Makefile
gouttegd: Didn’t test, but the Unix emulation layer in Windows (upon which the Windows version of Docker is based, I believe) would do a very poor job if it couldn’t handle something as basic and common as a symbolic link.

matentzn (approved):
This is an approval MINUS changes to uberon-edit.obo.

- [ ] Ensure that @anitacaron reviews these

gouttegd:
FYI, *all* the changes in uberon-edit have been explained beforehand over the past month as part of the discussion in Anita’s previous PR (#2928; see all comments starting from [this one](https://github.com/obophenotype/uberon/pull/2928#issuecomment-1726376929)). There has been no outright objection to them.

There are a couple of cases where @cmungall [suggested an approach](https://github.com/obophenotype/uberon/pull/2928#issuecomment-1739344064) that I have, for now, not followed:

> This is an example we see a bit where we have a choice between 1. Losing a
useful inference in many species for the sake of one; 2. Making an awkward
class for the one species and leaving the others alone
>
> We often do 1 when we should do 2. In this case the element would be
curiously floating with no part parent

I opted for the first option in those cases, because it keeps the PR simple. I’d rather have option 2 discussed in separate, dedicated issues and then implemented in separate, dedicated PRs, which we can do _after_ this one has been merged.

matentzn (approved):
I assume you vetted the action provider a bit, else fantastic

gouttegd:
@matentzn Nothing fishy in the [comment-pr](https://github.com/NejcZdovc/comment-pr/blob/main/index.js). :) That’s the same action that has already been used by @shawntanzk’s “diff-on-demand” (aka “gogoeditdiff”) action.

gouttegd:
Given that the last blocker for this PR are the two TX violations caused by errors in externally maintained resources, if by next week there has been no sign of progress on the corresponding upstream tickets ([here](https://github.com/obophenotype/ABA_Uberon/issues/36) and [here](https://github.com/hubmapconsortium/ccf-validation-tools/issues/293)), I vote for forcibly removing the offending axioms in the local copies of these resources (`src/ontology/bridge/uberon-bridge-to-dmba.owl` and `src/ontology/components/hra_subset.owl`) and merge.

matentzn:
> Given that the last blocker for this PR are the two TC violations caused by errors in externally maintained resources, if by next week thttps://github.com/obophenotype/ABA_Uberon/issues/36 has been no sign of progress on the corresponding upstream tickets (https://github.com/hubmapconsortium/ccf-validation-tools/issues/293 and here), I vote for forcibly removing the offending axioms in the local copies of these resources (src/ontology/bridge/uberon-bridge-to-dmba.owl and src/ontology/components/hra_subset.owl) and merge.

I agree!

gouttegd:
I’ve removed the last two offending assertions from our local copy of the externally maintained `uberon-bridge-to-dmba` and `hra_subset` resources as discussed above. This is now ready.

anitacaron:
Sorry, I'll remove the offending axiom now and need to update today's HRA subset for the last time. It won't have any other update before next year.

PR #3103, opened 2023-10-27
2023-10-27 Release

anitacaron:


src/ontology/mappings/fbbt-mappings.sssom.tsv
```diff
@@ -2,447 +2,445 @@
...
-FBbt:00110746	presumptive prothoracic metatarsus	semapv:crossSpeciesExactMatch	UBERON:6110746	semapv:ManualMappingCuration
-FBbt:00110811	presumptive arista	semapv:crossSpeciesExactMatch	UBERON:6110811	semapv:ManualMappingCuration
-FBbt:10000000	anatomical entity	semapv:crossSpeciesExactMatch	UBERON:0001062	semapv:ManualMappingCuration
+#mapping_set_id: "http://purl.obolibrary.org/obo/fbbt/fbbt-mappings.sssom.tsv"
+#mapping_set_description: "Mappings between the Drosophila Anatomy Ontology and foreign ontologies."
+#creator_id:
+#  - "https://orcid.org/0000-0002-6095-8718"
+#license: "https://creativecommons.org/licenses/by/4.0/"
+#subject_source: "http://purl.obolibrary.org/obo/fbbt.owl"
+#mapping_date: "2023-08-09"
```
anitacaron, line 11: @gouttegd do you confirm these changes, please?

gouttegd: Yes, everything is normal here. :)

The entire file has changed because we’ve added `object_source` to all rows.

anitacaron: Did you notice no more "semapv" in the curie_mapie?

gouttegd: As an aside, I’m starting to wonder whether it really makes sense to have this file committed in the repository, given that it’s a 100% exact copy of the file published by FBbt…

The only benefit of having this file here is that we don’t need network access to fetch it at release time, but given that we already have several network dependencies for a release, I am not entirely convince this is worth it.

Anyway, that’s not something to be changed right now while we are on the verge of releasing. Just something to note for later.

anitacaron: smart this SSSOM haha

aleixpuigb (approved):

Issue #3088, opened 2023-10-10
switch branching_part_of use in the vasculature to connecting branch of


dosumis:
branching_part_of is a subproperty of part_of, so use of it leads to distantly connected parts of the vasculature appearing to be parts of each other. the replacement is a subroperty of connected_to, so does not cause this problem

TBC: change the name of connecting_branch_of in RO to branches_from?

aleixpuigb:
should this change be exclusive of Uberon, or request the change in RO?

aleixpuigb:
This is the [list of terms](https://docs.google.com/spreadsheets/d/1V6aqnzEqESWXaHDyoThE_Sv4LyOxlhliAyTexIWljMs/edit?usp=sharing). The column connected to indicates that the object property needs to be changed. The column Other indicates if AS does not have the object property or if it has multiple and one needs to be removed.

dosumis:
I guess this table just has the subject terms?  I think it would be easier to automate this if we have a table of triples (subject , relation, object) that need to be changed.  It should be easy to do this via UberGraph queries.  

subject_label | subject_id | relation | object_label | object_id |
--|--|--|--|--

This can easily be turned into a robot template or used as source for a SPARQL update to remove the old relations.

dosumis:
@anitacaron we need a query like this

```sparql
SELECT * WHERE {
  ?subject_vessel rdfs:subClassOf vessel: .
  ?object_vessel rdfs:subClassOf vessel: .
  ?subject_vessel branching_part_of: ?object_vessel .
  ?subject_vessel rdfs:label ?s_label .
  ?subject_vessel rdfs:label ?o_label .
}
```

But where the first 2 clauses query the redundant graph and the second two query the non-redundant graph.

aleixpuigb:
As there are some grouping classes of branches 'branch of ...' , making `?subject_vessel branching_part_of: ?object_vessel .` non-redundant misses all the subclasses, which also have the object property 'branching part of'. However, as the grouping class already has this relation and it should be inherited, I would favour to delete them (manually?).

dosumis:
The safest way to identify what's already in the ontology would be to have this clause run on the ontology graph:

  ?subject_vessel branching_part_of: ?object_vessel .

as this will identify OWL axioms you need to target.

@anitacaron  should be able to help you write SPARQL that deals with blank nodes.

aleixpuigb:
[...]

PR #3106, opened 2023-11-01
Addresses #3088 change vascular object property

aleixpuigb:
Addresses #3088 change vascular object property


aleixpuigb:
#gogoeditdiff

dosumis (approved):
Looks good to me.  Task is not to review accuracy of relationships but to switch existing ones - so I've just reviewed on that basis.

PR #3109, opened 2023-11-02
[Test] Taxon constraint check with added feedback

gouttegd:
This is a variation of the PR #3102 with the addition that, if the QC check fails because of some violations of the taxon constraints, a comment should be automatically posted to the PR with the reasoner’s explanations for the violations.

Work in progress, for testing purposes only.


.github/workflows/qc.yml
```diff
@@ -27,10 +27,35 @@ jobs:
...
+        continue-on-error: true
         env:
           DEFAULT_BRANCH: master
           ROBOT_JAVA_ARGS: -Xmx6G
         run: cd src/ontology && mkdir -p tmp reports mirror && make ROBOT_ENV='ROBOT_JAVA_ARGS=-Xmx6G' GH_ACTION=true BRI=false MIR=false IMP=false test > TESTLOG.log
+
+      - name: Reason over taxon constraints
+        id: explaintc
+        continue-on-error: true
+        if: steps.check.outcome == 'failure'
```
anitacaron, line 40: This check could go on another yml file.

gouttegd: Doing it in another workflow would mean that the taxon constraint check should _also_ be performed, again, in that workflow. That’d be a waste of resources since the check has already been done here as part of the main QC workflow.

anitacaron: ok

gouttegd:
Closing here. This was a test and it worked, so now I’ve added a clean commit on top on #3102.

PR #3110, opened 2023-11-02
[Test] Taxon constraint check with added feedback

gouttegd:
This is yet another variation on #3102. Like #3109, its purpose is to add a GitHub Action that checks whether taxon constraints are violated by a PR, and post the reasoner’s explanations as a comment to the PR if that’s the case.

Contrary to #3109, in which running `robot explain` and posting the result was done as part of the main QC workflow, here this is done as part of a distinct workflow.


gouttegd:
So the triggering of a second workflow upon failure of a first don’t seem to work as expected…

anitacaron:
Thank you for the try. That's fine to be inside the QC workflow.

PR #3113, opened 2023-11-07
Replace type definition for latin term synonyms

cthoyt:


cmungall (approved):

cmungall:
DIff is too big but I reviewed individual commits and they seem fine. Merge once the conflict is resolved.

(it seem conflicts are likely for this kind of change, maybe we can batch these and do them all with a perl one-liner?)

cthoyt:
@anitacaron I rebased on the main branch and added the right import. This should be ready to merge

anitacaron (approved):

PR #3121, opened 2023-11-17
Preliminary check for ontobot workflow

hrshdhgd:
The workflow first does a preliminary check if `Hey ontobot` exists in the issue body. Only if the phrase exists will the rest of the workflow run.


hrshdhgd:
@anitacaron , translating the same update for the ontobot workflow from cell-ontology to uberon.

anitacaron (changes requested):
Could you please update it to use the new plugin, the same as in CL?  https://github.com/obophenotype/cell-ontology/blob/master/.github/workflows/ontobot.yml

anitacaron:
In the CL repo, for the same workflow, you downloaded robot plugin, this part is not needed here?

https://github.com/obophenotype/cell-ontology/blob/f76d14e7e30cca21c477d91d15e2c9a46743dcd0/.github/workflows/ontobot.yml#L74-L78

hrshdhgd:
No. That's only if the resource is an owl file. Here the resource is an obo file so the workflow is up-to-date.

hrshdhgd:
Exactly!

anitacaron (approved):

Issue #3056, opened 2023-09-06
Bridging axioms when 2 xrefs point to the same foreign term


gouttegd:
We have quite a few cases where two different Uberon terms have a cross-reference to the same foreign term, as in this example:

```
[Term]
id: UBERON:0000977
name: pleura
xref: EMAPA:16775

...
[Term]
id: UBERON:0003390
name: mesothelium of pleural cavity
xref: EMAPA:16775
```

I am unsure about what to do with such cross-references when generating the bridges.

Under the current system (using @cmungall ’s `make-bridge-ontologies-from-xrefs.pl` script), this will result initially in two frames in the bridge file:

```
[Term]
id: EMAPA:16775
intersection_of: UBERON:0000977 ! pleura
intersection_of: part_of NCBITaxon:10090


[Term]
id: EMAPA:16775
intersection_of: UBERON:0003390 ! mesothelium of pleural cavity
intersection_of: part_of NCBITaxon:10090
```

Upon converting the OBO bridge file to OWL, the two frames will be merged, resulting in a single equivalence axiom between EMAPA:16775 and the intersection of UBERON:0000977, UBERON:0003390, and the existential restriction on part_of NCBITaxon:10090:

```
EquivalentClasses(EMAPA:16775 ObjectIntersectionOf(UBERON:0000977 UBERON:0003390 ObjectSomeValuesFrom(BFO:0000050 NCBITaxon:10090)))
```

I am not sure this behaviour is correct or expected.

Currently, my SSSOM-based bridge generation process, in the same situation, would generate the two following equivalence axioms instead:

```
EquivalentClasses(EMAPA:16775 ObjectIntersectionOf(UBERON:0000977 ObjectSomeValuesFrom(BFO:0000050 NCBITaxon:10090)))
EquivalentClasses(EMAPA:16775 ObjectIntersectionOf(UBERON:0003390 ObjectSomeValuesFrom(BFO:0000050 NCBITaxon:10090)))
```

But I am not convinced this is the correct thing to do either.

From the cases I have seen, I am inclined to think that most if not all cases of "2 Uberon terms mapped to the same foreign term" are actually bogus, most likely the result of one editor adding a cross-reference on a Uberon term to a foreign term without realising that another Uberon term was already mapped to the same foreign term.

To generate the bridges, I am considering either:
* ignoring such cases entirely: do not generate any bridging axioms to a foreign term if there are more than strictly one cross-reference to it;
[...]

PR #3122, opened 2023-11-18
Small fixes to the new bridge pipeline

gouttegd:
This PR amends the recently introduced SSSOM-based bridge pipeline to use the latest version (0.6.1) of the SSSOM plugin for ROBOT. This allows to:

* make sure the mapping sets extracted from CL and ZFA have an ID and a license tag, as required by the SSSOM specification (the license is the same as the one used by the ontology those sets are derived from);
* (much more importantly) make sure the the CL-to-ZFA bridge is complete.

### DIFF
--- a/src/ontology/bridge/cl-bridge-to-zfa.owl
+++ b/src/ontology/bridge/cl-bridge-to-zfa.owl
@@ -9,7 +9,7 @@
      xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
      xmlns:terms="http://purl.org/dc/terms/">
     <owl:Ontology rdf:about="http://purl.obolibrary.org/obo/uberon/bridge/cl-bridge-to-zfa.owl">
-        <owl:versionIRI rdf:resource="http://purl.obolibrary.org/obo/uberon/releases/2023-11-13/bridge/cl-bridge-to-zfa.owl"/>
+        <owl:versionIRI rdf:resource="http://purl.obolibrary.org/obo/uberon/releases/2023-11-18/bridge/cl-bridge-to-zfa.owl"/>
         <terms:contributor>Melissa Haendel</terms:contributor>
         <terms:contributor>Veri Van Slyke</terms:contributor>
         <terms:contributor>Yvonne Bradford</terms:contributor>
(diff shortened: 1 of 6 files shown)


gouttegd:
<details>
<summary>For the technically inclined, this (collapsed) comment explains why the CL-to-ZFA bridge is currently not complete (TL;DR: because of a bug in the SSSOM plugin, fixed in 0.6.1).</summary>

The SSSOM/T ruleset that generates the bridging axioms starts with three generic rules:

```
subject==UBERON:* || subject==CL:* -> invert();
```

The first one takes any mapping that has a UBERON or CL term as the subject and inverts it, so that the UBERON or CL term becomes the object. This is necessary because we collect the mappings from several sources (from UBERON itself, from CL, from FBbt, and from ZFA) and the “orientation” of the mappings may be different from one source to the next. This rules ensure all the mappings are always in the same orientation, with the UBERON or CL terms on the object side.

```
!(object==UBERON:* || object==CL:*) -> stop();
```

The second rule takes any mapping whose object is *not* a UBERON or CL term and excludes it from any further processing (in effect, it’s as if the mapping was simply discarded). This is necessary because, again, we collect mappings from different sources and a particular mapping set may contain more than just mappings to UBERON or CL. For example, the ZFA mapping set contains mappings between ZFA and CL (that we want) but also mappings between ZFA and TAO (that we don’t care about). This rule eliminates the mappings with anything else than UBERON or CL, so that we don’t have to worry about their presence in the rest of the process.

```
!cardinality==*:1 -> stop();
```

Finally, the third rule excludes any mapping that does not have a cardinality of `1:1` (one foreign subject mapped to one UBERON/CL object) or `n:1` (several foreign subjects mapped to one UBERON/CL object). That is, it eliminates `1:n` mappings (one foreign subject mapped to several UBERON/CL objects) and `n:n` mappings. This is necessary because such mappings (which have been found in the wild, see #3056) are almost certainly bogus: it’s perfectly normal for a UBERON term to be mapped with more than one foreign term (in fact that’s expected: a UBERON term will be mapped to one term from FBbt, to one term from ZFA, to one term from WBbt, etc.), but it’s not expected that a single foreign term (say, a FBbt term) should be mapped to more than one UBERON term. If that happens, it’s almost certainly a mapping error, and the rule above takes care of eliminating that case.

So far so good, but the problem is that, up to version 0.6.0 (included), the SSSOM plugin for ROBOT was calculating the cardinality of the mappings only once, just after collating all the input mapping sets into a single set and before applying any rule. To understand how it is a problem, consider the following mappings:

```
ZFA:0009005  semapv:crossSpeciesExactMatch  CL:0000017
ZFA:0009005  semapv:crossSpeciesExactMatch  TAO:0009005
```

When the cardinality is calculated, these mappings are identified as `1:n` mappings: one subject (ZFA:0009005) mapped to many objects (CL:0000017, TAO:009005).

When the second rule above is applied, the mapping with TAO is excluded, since it is a mapping with an object that is neither a UBERON term nor a CL term. But the remaining mapping with CL:0000017 is still considered to be a `1:n` mapping! So when we reach the third rule, that mapping gets also excluded, since it does not have the cardinality we expect.

And that is how many of the ZFA-CL mappings were ignored in the first version of the new pipeline! Sorry about that. (Though it was still better than the old pipeline, which ignored *all* the ZFA-CL mappings.)

The fix is simply to ensure that, when we need to filter on cardinality (when a SSSOM/T rule has a `cardinality==...` filter), the cardinality of each mapping is re-calculated if any mapping has been excluded by a previous rule. That’s what version 0.6.1 of the SSSOM plugin does.

</details>

matentzn (approved):
Looks great, can't see anything suspicious!

gouttegd:
Of course, all the suspicious stuff is hidden away in the code of the SSSOM plugin! :smiling_imp:

PR #3127, opened 2023-11-23
Use official PURLs for the FMA bridge.

gouttegd:
When extracting cross-references to FMA, make sure the resulting SSSOM file use the official PURLs for FMA classes, instead of OBO PURLs (which had been used in the past for the now obsolete OBO version of FMA). This is turn results in the `uberon-bridge-to-fma.owl` bridge file also using those PURLs.


gouttegd:
See issue #3056, especially [this comment](https://github.com/obophenotype/uberon/issues/3056#issuecomment-1822616493), for context.

matentzn (approved):
Cool, assuming the final fma is not followed by a separator.

gouttegd:
> assuming the final fma is not followed by a separator.

No, FMA class IDs are of the form `http://purl.org/sig/ont/fma/fmaXXXXXXX`, _without_ a separator (`#`, `_`, `/` or otherwise) before the numerical part.

Note that _properties_ in FMA do not follow that form, and are instead of the form `http://purl.org/sig/ont/fma/AAAAAAAAA` (where `AAAAAAAAA` is a human-readable ID, as in `http://purl.org/sig/ont/fma/adjacent_to`). But we are only mapping to FMA _classes_, so it shouldn’t be a concern.

Issue #2129, opened 2021-10-25
There are multiple terms that have "." as definition


shawntanzk:
There are multiple terms that have "." as definition - this seems super weird, it is triggering duplicate_definition. Will investigate. Examples include:

First 5 violations:
ERROR	duplicate_definition	UBERON:0000089	IAO:0000115	.
ERROR	duplicate_definition	UBERON:0000219	IAO:0000115	.
ERROR	duplicate_definition	UBERON:0000444	IAO:0000115	.
ERROR	duplicate_definition	UBERON:0001430	IAO:0000115	.
ERROR	duplicate_definition	UBERON:0001431	IAO:0000115	.

Note: this does have xrefs attached to it, so it seems it might have been accidentally removed or something?

Btws, found this when running `sh run.sh make reports/uberon-edit.obo-obo-report.tsv`

shawntanzk:
- [ ] @gouttegd to look into this in makefile to see where this should come from
- [ ] @shawntanzk will follow up if we can't figure this out

gouttegd:
The “.” definitions are not injected from anywhere in the pipeline, they come straight from the -edit file.

Contrary to what we have in FBbt, in Uberon it does not seem that there is anything in place in the pipeline to deal with “dot-definitions“.

For reference, what happens in FBbt is that terms with “dot-definitions” are extracted (using [this SPARQL query](https://github.com/FlyBase/drosophila-anatomy-developmental-ontology/blob/master/src/sparql/dot-definitions.sparql)), and processed by [a custom Java program](https://github.com/monarch-ebi-dev/eqwriter) (apparently specifically written for FlyBase) to generate textual definitions from the logical definitions.

Not sure if we want to apply the FlyBase approach in Uberon (of note, some Uberon terms with a “dot-definition” don’t have a logical definition, so that approach would not work anyway for them)… For 131 terms, I am inclined to say the FlyBase approach is overkill and that it would be easier and faster to have the curators manually add a proper definitions to all those terms…

shawntanzk:
@cmungall @dosumis might need help with this

cmungall:
I always thought the dot-definitions in FBbt were placeholders for
literature or some other pointer. Anyway that's what they are in uberon.
Not sure what I can add beyond what I said earlier. If this is priority
then we can just translate to seeAlso or similar

matentzn:
@cmungall the question is, what exactly are they placeholders for? Where is the pipeline that fills them in with actual definitions?

[...]

PR #3128, opened 2023-11-27
Replace "dot definitions" for some insect-related terms.

gouttegd:
Several terms about insect anatomical structures have so-called “dot definitions”. We replace by rudimentary definitions drawn from the corresponding terms in FBbt.

This is part of the work to eliminate all such “dot definitions” in Uberon (#2129).


anitacaron (approved):

anitacaron:
Thanks, @gouttegd!

Issue #2588, opened 2022-08-08
Fix composite-metazoan pipeline 


shawntanzk:
1) Look at https://github.com/obophenotype/uberon/issues/2584 -> how did CL relationship go missing? (CL variant that is imported in this process that is probably not the right one? or CL is not imported?) 
2) https://github.com/obophenotype/uberon/issues/1466
3) ?

gouttegd:
From the Uberon call of October 24th, 2022: The whole point of the `composite-*` products is precisely to have the kind of “unfolded” structure that `owltools --merge-species-ontology` generates, so this should not be changed.

-> Keep the `owltools`-dependent pipeline for now, and start thinking whether the `--merge-species-ontology` feature should be ported/re-implemented elsewhere.

aschroed:
Is there anywhere that a recent version of composite-metazoan.owl is available for download and use?

anitacaron:
Hi @aschroed, the `composite-metazoan.owl` is now available in the [release assets](https://github.com/obophenotype/uberon/releases).

gouttegd:
I propose to use this ticket to centralise all discussions about the overhauling of the `composite-metazoan` (hereafter CM) pipeline.

Here’s a list of all the issues affecting this product. I strongly suspect most if not all are directly caused by the complexity of the CM pipeline (and the lack of love shown to that pipeline over the last few years), and therefore most if not all of them could be fixed by streamlining the pipeline (one can always hope, at least).

- [ ] #1466
- [ ] #1813
- [ ] #1952
- [ ] #1957
- [ ] #1958
- [ ] #1959
- [ ] #2032
- [ ] #2584
- [ ] #2782

Regarding #1952, the unsats may be caused by the actual contents of the merged ontologies and as such may not be fixed by a new pipeline, _but_ a cleaner, faster pipeline may make testing for unsats easier and help detecting and fixing those in the future.

gouttegd:
Here’s a quick overview of the current CM pipeline.

## Step 1: $TMPDIR/merged-composite-metazoan.owl

This first step is merely a big merge of a bunch of files:

* `$TMPDIR/ext-weak.owl` (mostly Uberon itself, minus some disjointness axioms and equivalent-to-nothing axioms);
* all the cross-ontology bridges (e.g. `uberon-to-fbbt-bridge.owl`, etc.);
* `imports/local-*.owl` (where `*` = `fbbt`, `wbbt`, `zfa`, etc.)
[...]

PR #2614, opened 2022-08-16
Created general_subset, to replace CARO ontology

balhoff:
Related to #2349

I created a `general_subset` containing terms that must be applicable outside of Metazoa. I'm open to suggestions for an alternative name for the subset. For all these terms, I added present_in_taxon annotations:

- relationship: present_in_taxon NCBITaxon:33090 ! Viridiplantae
- relationship: present_in_taxon NCBITaxon:33208 ! Metazoa
- relationship: present_in_taxon NCBITaxon:4751 ! Fungi

Hopefully these will help catch unwanted taxonomic specificity connected to these terms.

This PR also updates the taxon import to include Fungi.

PR #3129, opened 2023-11-30
New composite pipeline

gouttegd:
This is an overhaul of the pipeline that creates the `composite-*` products, most importantly the `composite-metazoan` product.

* Any use of `owltools` is replaced by `robot`, when necessary using a [Uberon-specific plugin](https://github.com/gouttegd/uberon-robot-plugin). This allows the pipeline to work with some OWL constructs introduced in CL and FBbt over the past two years, that OWLTools was not able to handle.
* The `src/ontology/bridge/collected-*.owl` files, which were used to define the contents of each composite product (through import declarations) are not used anymore, and are removed. Instead, Each product is solely defined by pre-requisites listed in `uberon.Makefile`.
* Some composite products are retired (see [this comment](https://github.com/obophenotype/uberon/issues/2588#issuecomment-1832706448) in #2588 for which products are retired and why).
* The immediate precursor to `composite-metazoan`, `collected-metazoan`, is turned into an official release product.
* All the foreign ontologies that end up in one or several of the `composite` products are treated as “local imports” and are committed to the repository. This was already the case for most of them, this PR generalises this behaviour for _all_ of them.


src/ontology/uberon.Makefile
```diff
@@ -24,6 +24,12 @@ all: uberon-qc
 	echo "make $@ succeeded..."
 
 
+# Enable second expansion of prerequisites (GNU Make only!).
+# We need this to expand the COLLECTED_*_SOURCES variables in the
+# composite pipeline.
+.SECONDEXPANSION:
```
matentzn, line 30: Are we sure that this does not at all affect anything in the normal, first expansion?

src/ontology/uberon.Makefile
```diff
@@ -49,8 +55,7 @@ update_dynamic_catalog:
 	@echo "WARNING: files are expected to be available. Do not do this if you dont"
 	@echo "WARNING: know what you are doing."
 	$(SCRIPTSDIR)/make-catalog.pl uberon.owl mirror/ncbitaxon.owl imports/*_import.owl \
-		mirror/ro.owl imports/local-*owl $(BRIDGEDIR)/*owl $(TMPDIR)/allen-*.obo \
-		$(TMPDIR)/developmental-stage-ontologies/src/ssso-merged.obo > $@.tmp && \
```
matentzn, line 53: What is the story behind removing `ssso-merged.obo`?

src/ontology/uberon.Makefile
```diff
@@ -121,6 +126,11 @@ $(TMPDIR)/plugins/sssom.jar:
 	mkdir -p $(TMPDIR)/plugins
 	curl -L -o $@ https://github.com/gouttegd/sssom-java/releases/download/sssom-java-0.6.1/sssom-robot-plugin-0.6.1.jar
 
+# Ditto for the specific Uberon plugin
+$(TMPDIR)/plugins/uberon.jar:
+	mkdir -p $(TMPDIR)/plugins
+	curl -L -o $@ https://github.com/gouttegd/uberon-robot-plugin/releases/download/uberon-robot-plugin-0.1.0/uberon.jar
```
matentzn, line 132: Is there a _really_ good reason for splitting all the gouttegd-robot-extensions into multiple jars?

src/ontology/uberon.Makefile
```diff
@@ -209,163 +219,213 @@ mirror-ehdaa2: | $(TMPDIR)
...
 # The following imports are used in Uberon-specific pipelines. They are
 # *not* updated as part of the refresh-imports pipeline. In fact, they
 # are normally never updated automatically, since Uberon-specific
 # pipelines are typically run with IMP set to false.
 
-# Map legacy OBO-format object properties to their BFO/RO equivalent
-# FIXME: this may no longer be needed for some of these imports
-# (https://github.com/obophenotype/uberon/issues/3017)
+# By default, we would like to simply use the mirrored ontology as it
+# is, but to avoid injecting random crap, we make a pseudo-base.
```
matentzn, line 331:
```suggestion
# is, but to avoid injecting axioms that do not belong to the ontology and are potentially conflicting, we make a pseudo-base.
```

😆

src/ontology/uberon.Makefile
```diff
@@ -209,163 +219,213 @@ mirror-ehdaa2: | $(TMPDIR)
...
+		        --axioms external --preserve-structure false --trim false \
+		 convert -f ofn -o $@
+
+# We need a special case for FBbt/dv because of the mixed-case prefix
+imports/local-fb%.owl: mirror/fb%.owl
+	$(ROBOT) remove -i $< --base-iri $(URIBASE)/FB \
+		        --axioms external --preserve-structure false --trim false \
+		 convert -f ofn -o $@
+
+# Ditto for WBbt/ls
```
matentzn, line 345:
```suggestion
# We need a special case for WBbt/ls because of the mixed-case prefix
```

src/ontology/uberon.Makefile
```diff
@@ -209,163 +219,213 @@ mirror-ehdaa2: | $(TMPDIR)
...
+imports/local-wb%.owl: mirror/wb%.owl
+	$(ROBOT) remove -i $< --base-iri $(URIBASE)/WB \
+		        --axioms external --preserve-structure false --trim false \
+		 convert -f ofn -o $@
+
+# Ditto for MmusDv
```
matentzn, line 351:
```suggestion
# We need a special case for MmusDv because of the mixed-case prefix
```

src/ontology/uberon.Makefile
```diff
@@ -209,163 +219,213 @@ mirror-ehdaa2: | $(TMPDIR)
...
+imports/local-mmusdv.owl: mirror/mmusdv.owl
+	$(ROBOT) remove -i $< --base-iri $(URIBASE)/MmusDv_ \
+		        --axioms external --preserve-structure false --trim false \
+		 convert -f ofn -o $@
+
+# Ditto for HsapDv
```
matentzn, line 357:
```suggestion
# We need a special case for HsapDv because of the mixed-case prefix
```

src/ontology/uberon.Makefile
```diff
@@ -209,163 +219,213 @@ mirror-ehdaa2: | $(TMPDIR)
...
+		 remove --base-iri $(URIBASE)/XAO_ --axioms external \
+		        --preserve-structure false --trim false \
+		 convert -f ofn -o $@
+
+# And for those ones, we also remove the imports. They were last
+# refreshed in 2016, we don't want to bring in old issues from old
+# versions of Uberon, CL, RO, etc.
+# ----------------------------------------
+imports/local-poro.owl: mirror/poro.owl
+	$(ROBOT) remove -i $< --select imports --trim false \
```
matentzn, line 413: Sidenote: if the presence of imports is a general issue, it would have been safe to add this clause to all `imports/local-%` rules (for futureproofing). I know this is overly cautious, so I am happy for the comment to be ignored.

src/ontology/uberon.Makefile
```diff
@@ -1025,147 +1095,144 @@ subsets/%-view.owl: uberon.owl contexts/context-%.owl tmp/simple-slim-seed.txt
...
+
+COLLECTED_worm_SOURCES =             $(IMPORTDIR)/local-wbbt.owl \
+				     $(IMPORTDIR)/local-wbls.owl \
+				     $(BRIDGEDIR)/uberon-bridge-to-wbbt.owl \
+				     $(BRIDGEDIR)/uberon-bridge-to-wbls.owl \
+				     $(BRIDGEDIR)/cl-bridge-to-wbbt.owl
+
+COLLECTED_zebrafish_SOURCES =        $(IMPORTDIR)/local-zfa.owl \
+				     $(BRIDGEDIR)/uberon-bridge-to-zfa.owl \
+				     $(BRIDGEDIR)/cl-bridge-to-zfa.owl
```
matentzn, line 1118: I am noticing here that there is no specific zfs related import, which is probably fine, but I wanted it said; For now, zfa comes with zfs built in, but this may change in the next year.

src/ontology/uberon.Makefile
```diff
@@ -1025,147 +1095,144 @@ subsets/%-view.owl: uberon.owl contexts/context-%.owl tmp/simple-slim-seed.txt
...
-# Likewise.
-composite-vertebrate.owl: $(TMPDIR)/unreasoned-composite-vertebrate.owl
-	$(ROBOT) reason -i $< -r ELK --equivalent-classes-allowed all \
+# Step 1: Create a "collected" ontology, which is simply a merge of
+# Uberon, CL, the CARO bridges, and the components listed in the
+# COLLECTED_*_SOURCES variables above.
+.PRECIOUS: $(TMPDIR)/collected-%.owl
+$(TMPDIR)/collected-%.owl: $(BRIDGEDIR)/collected-%-hdr.owl uberon.owl $(IMPORTDIR)/local-cl.owl \
+		 $(BRIDGEDIR)/uberon-bridge-to-caro.owl \
+		 $(BRIDGEDIR)/cl-bridge-to-caro.owl \
```
matentzn, line 1190: Given that there has been an effort to get rid of caro in favour of a thin anatomy top layer in uberon, https://github.com/obophenotype/uberon/pull/2614, can you judge if removing this will cause issues?

src/ontology/bridge/collected-adult-mammal.owl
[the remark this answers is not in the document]
gouttegd: Not really, because `collected-X-hdr.owl` is _not_ merely the new name of `collected-X.owl`.

`collected-X.owl` was the old way of describing what goes into the `X` product: it contained import declarations for all the components that made up that product. Which was nice, except that it led to needless confusion as I tried to explain [in this comment](https://github.com/obophenotype/uberon/issues/2588#issuecomment-1830459080) (for example, when you look at the import declarations in `collected-metazoan.owl`, you find that it imports `ssso-merged.obo`, but then you needed to look up in the catalog to find out that what was really imported was `ssso-merged-uberon.obo`).

The new `collected-X-hdr.owl` file is _not_ used at all to define the contents of the X product – only the Makefile does that.

src/ontology/catalog-v001.xml
```diff
@@ -1,60 +1,58 @@
...
-      <uri name="http://purl.obolibrary.org/obo/uberon/bridge/collected-embryonic-mammal.owl" uri="bridge/collected-embryonic-mammal.owl"/>
-      <uri name="http://purl.obolibrary.org/obo/mba.owl" uri="tmp/allen-mba.obo"/>
-      <uri name="http://purl.obolibrary.org/obo/pba.owl" uri="tmp/allen-pba.obo"/>
-      <uri name="http://purl.obolibrary.org/obo/hba.owl" uri="tmp/allen-hba.obo"/>
-      <uri name="http://purl.obolibrary.org/obo/dmba.owl" uri="tmp/allen-dmba.obo"/>
-      <uri name="http://purl.obolibrary.org/obo/dhba.owl" uri="tmp/allen-dhba.obo"/>
+      <uri name="http://purl.obolibrary.org/obo/ceph.owl" uri="imports/local-ceph.owl"/>
+      <uri name="http://purl.obolibrary.org/obo/cl.owl" uri="imports/local-cl.owl"/>
+      <uri name="http://purl.obolibrary.org/obo/cteno.owl" uri="imports/local-cteno.owl"/>
+      <uri name="http://purl.obolibrary.org/obo/dhba.owl" uri="imports/local-allen-dhba.obo"/>
```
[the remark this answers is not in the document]
gouttegd: It’s OWL files for everything _except_ for the “Allen” ontologies, which are in OBO because they are custom-generated by [a script](https://github.com/obophenotype/uberon/blob/master/src/scripts/allen-json2obo.pl) that produces OBO files from whatever JSON we get from the Allen API. I have not changed anything about that.

src/ontology/imports/map-mouse-stages.tsv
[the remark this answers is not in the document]
gouttegd: Yes, `robot rename --mappings` is not designed to accept SSSOM files. Nice idea for a new feature in ROBOT, though.

src/ontology/uberon.Makefile
gouttegd: Yes. All it does is allowing the use of this:

```Make
$(TMPDIR)/collected-%.owl: $$(COLLECTED_$$*_SOURCES)
```

That is, to define the prerequisites of a pattern rule using a variable whose name is itself dependent on the pattern.

src/ontology/uberon.Makefile
gouttegd: Yes. The SSSOM plugin is a generic plugin to use SSSOM files in a ROBOT pipeline. There’s nothing specific to Uberon in it.

By constrast, the Uberon plugin, as its name implies, is _very_ specific to Uberon. I doubt any other project than Uberon will ever need a `merge-species` command.

I invert the question: give me a _really_ good reason why those plugins should be merged into a single one?

src/ontology/uberon.Makefile
gouttegd: I don’t see any difference, the two sentences seem completely equivalent to me. :D

src/ontology/uberon.Makefile
gouttegd: Overall, and for now, the composition of the various collected/composite products has been left mostly unchanged. I did not include ZFS here because the original `composite-zebrafish` did not include it.

If/when we need to explicitly include ZFS, it would be very easy to do so. Just add a `$(IMPORTDIR)/local-zfs.owl` import and update this variable to include that file as well as the (existing but currently unused) `$(BRIDGEDIR)/uberon-bridge-to-zfs`.

src/ontology/uberon.Makefile
matentzn: I somehow lost touch of this part of the development - so we don't need the ssso in the process at all anymore?

src/ontology/uberon.Makefile
matentzn: I know, my statement was about the unlikely possibility that someone starts adding imports to their ontology; adding this statement just makes sure that everything runs; not a biggiue

src/ontology/uberon.Makefile
matentzn: I just wanted you to be aware, makes sense!

src/ontology/uberon.Makefile
matentzn: Ok, just wanted you to be aware of this, since you are, I will resolve comment!

src/ontology/uberon.Makefile
gouttegd: I’d rather not add code and complexity to deal with hypothetical problems. _If_ we observe that another ontology is publishing a release product with import declarations, _then_ we’ll add code to deal with that.

src/ontology/uberon.Makefile
gouttegd: You must be joking… Building Uberon requires a ~2.5 GB Docker image, and you’re concerned about the size of two Jar files of ~2,600 KB (the SSSOM plugin) and ~23 KB (the Uberon plugin)?

In any case, merging the two plugins would not change anything about their size, because each plugin only contains its own code and nothing else – all the common code, such as the OWL API, is provided by ROBOT. So yeah, we could merge the two plugins, and then instead of having two files of 2,600 KB and 23 KB, we would have only one of 2,623 KB…

src/ontology/uberon.Makefile
gouttegd: We still include the SSSO in `composite-metazoan`, but instead of doing the weird things that we used to do (cloning the entire repository in `$(TMPDIR)` and using it from there), we treat it the same way we treat all the other foreign ontologies:

* mirror it into `$(MIRRORDIR)`;
* create a “local import” module in `$(IMPORTDIR)/local-ssso.owl`, which is committed to the repository;
* use `$(IMPORTDIR)/local-ssso.owl` whenever we need it that ontology (such as when building `composite-metazoan`).

The only special thing about SSSO compared to the other foreign ontologies is that for the mirroring step, we get the ontology directly from its GitHub source instead of going through `purl.obolibrary.org`, because the SSSO does _not_ have a PURL configuration. (By the way it should have one IMO, but that’s another issue.)

src/ontology/uberon.Makefile
matentzn: I remember in the past some strange issues with updating Variables like using the += operator and similar, but I guess we don't use such patterns.

src/ontology/uberon.Makefile
matentzn: I know you dont mean that in an _absolute_ sense, but lets roll with it given the low probability of the issue happening.

gouttegd:
Note that there are still things that can be improved, including:

* The “merge species” step leaves behind a bunch of class declaration axioms for the taxon-specific classes that have been “merged“ into their taxon-neutral equivalent. That is, when (for example) `EMAPA:17443` is replaced by `UBERON:00005695 and (BFO:0000050 some NCBITaxon:10090)`, we still have a class declaration axiom for `EMAPA:17443`:

```xml
<!-- http://purl.obolibrary.org/obo/EMAPA_17443 -->
<owl:Class rdf:about="http://purl.obolibrary.org/obo/EMAPA_17443">
    <rdfs:subClassOf rdf:resource="http://www.w3.org/2002/07/owl#Thing"/>
</owl:Class>
```

* General class axioms are not translated. If we have, say, a GCA that says

```
FBbt:X and (BFO:0000050 some FBbt:Y) SubClassOf BFO:0000050 some FBbt:Z
```

and upon merging `FBbt:X` is merged into `UBERON:X and (BFO:0000050 some NCBITaxon:7227)`, the GCA will _still_ refer to `FBbt:X`.

* We have a bunch of FBgn pseudo-terms from the FlyBase ontologies, which in my opinion don’t belong to `composite-metazoan`.

Those problems are not new – they already existed in previous versions of `composite-metazoan`. They should be easier to fix now, but I’d rather do that as a separate PR (possibly several separate PRs). My aim for this PR was strictly to replace the existing pipeline by one that is easier to maintain, but that is otherwise equivalent (including in its bugs! :P ).

anitacaron:
Just to let you know, I've read all the comments. I'll create an issue to remove CARO from the pipeline.

Issue #3133, opened 2023-12-04
Revisit how the Allen ontologies are imported into `composite-metazoan`


gouttegd:
Among the components that make up `composite-metazoan` are the 5 “ontologies“ derived from the Allen brain atlases: HBA/DHBA (human), MBA/DMBA (mouse), and PBA (primate).

They are included into `composite-metazoan` by a very specific process in which we download some JSON representation from the `brain-map.org` API server and transform the result into OBO files:

```Makefile
$(TMPDIR)/allen-mba.json: | $(TMPDIR)
        wget https://api.brain-map.org/api/v2/structure_graph_download/1.json -O $@

$(IMPORTDIR)/local-allen-%.obo: $(TMPDIR)/allen-%.json $(SCRIPTSDIR)/allen-json2obo.pl
        $(SCRIPTSDIR)/allen-json2obo.pl $< > $@
```

There are at least two problems with this approach:

a) The first is that the `allen-json2obo.pl` script is hardcoding some axioms when creating the OBO files from the JSON structure:

```perl
print "is_a: UBERON:0002616 ! regional part of brain\n";
```

This line is executed unconditionally and therefore forcibly makes any term from the Allen ontologies a subclass of UBERON:0002616. As it was recently found out (#2903), this classification is actually wrong, at least for _some_ Allen terms.

b) The second is that the Allen ontologies are switching (or have already switched) to using PURLs that do not depend on `purl.obolibrary.org` (`https://purl.brain-bican.org/ontology/`). Because of that, it’s a bad idea to use the OBO format to import those ontologies, since support for “non-OBO” PURLs in the OBO format is still [under development](https://github.com/owlcs/owlapi/pull/1102). When we merge those OBO files to create `composite-metazoan`, the IDs they contain are forcibly interpreted as being OBO PURLs.

Both those problems could easily be solved if we (Uberon) stopped generating the Allen ontologies ourselves from whatever JSON is provided by the Allen institute. It seems the Allen folks are now providing “ready-to-use“ ontologies (e.g. https://github.com/brain-bican/human_brain_atlas_ontology for HBA, https://github.com/brain-bican/developing_mouse_brain_atlas_ontology for DMBA, etc.), we could/should use them instead.

[...]

PR #3134, opened 2023-12-05
Fix the Allen ontologies used in composite-metazoan

gouttegd:
This PR:

* changes the way we include the Allen ontologies into the `composite-*` products: instead of generating the ontologies locally by processing JSON data obtained from `brain-map.org`, we use the pre-built ontologies made available on https://github.com/brain-bican, as proposed in #3133;
* updates the DHBA, HBA, and PBA bridges to make sure they are using the correct IRIs (e.g. `https://purl.brain-bican.org/ontology/dhbao/DHBA_XXX` instead of `http://purl.obolibrary.org/obo/DHBA_XXX`).


matentzn (approved):
LGTM. There dies another PERL script a painful death!

anitacaron:
I think it's important to link this ancient draft PR #1406 that is also changing the Allen ontologies and tag people from there @tgbugs @dosumis @cmungall.

dosumis:
> I think it's important to link this ancient draft PR https://github.com/obophenotype/uberon/pull/1406 that is also changing the Allen ontologies and tag people from there @tgbugs @dosumis @cmungall.

For reference, the apparent problems noted on that PR with unstable IDs are resolved by ensuring that the correct mappings between allen 'ontologies' and atlases.  I think we can close.

PR #3135, opened 2023-12-05
Use features from latest Uberon plugin to generate composite-metazoan.

### LINKED ISSUE #3132: Remaining minor issues with the `composite-` products
The new `composite-` pipeline has inherited some of the defects of the old pipeline that it replaced. Hopefully in the new pipeline it will be easier to fix those.

## Useless class declarations
The “merge species” step leaves behind a bunch of class declaration axioms for the taxon-specific classes that have been “merged“ into their taxon-neutral equivalent. That is, when (for example) `EMAPA:17443` is replaced by `UBERON:00005695 and (BFO:0000050 some NCBITaxon:10090)`, we still have a class declaration axiom for `EMAPA:17443`:

```xml
<!-- http://purl.obolibrary.org/obo/EMAPA_17443 -->
<owl:Class rdf:about="http://purl.obolibrary.org/obo/EMAPA_17443">
    <rdfs:subClassOf rdf:resource="http://www.w3.org/2002/07/owl#Thing"/>
</owl:Class>
```

## Untranslated GCAs
The “merge species“ step does not translate general class axioms. If we have, say, a GCA that says

```
FBbt:X and (BFO:0000050 some FBbt:Y) SubClassOf BFO:0000050 some FBbt:Z
```

and upon merging `FBbt:X` is merged into `UBERON:X and (BFO:0000050 some NCBITaxon:7227)`, the GCA will still refer to `FBbt:X`.

## Polluting FBgn terms

We have a bunch of FBgn pseudo-terms from the FlyBase ontologies, which in my opinion don’t belong to composite-metazoan.

gouttegd:
This PR updates the Uberon ROBOT plugin to [version 0.2.0](https://github.com/gouttegd/uberon-robot-plugin/releases/tag/uberon-robot-plugin-0.2.0). This versions includes the following features to the `uberon:merge-species` command:

* `-d`: drop declaration axioms for classes that have been merged;
* `-g`: try to translate general class axioms (GCAs) that refer to a class that has been merged;
* `-x`: translate more class expressions than what was possible with OWLTools’ `--merge-species-ontology` command.

We use those new features when creating the `composite-*` products.

closes #3132

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -1204,7 +1204,7 @@ collected-metazoan.owl: $(TMPDIR)/collected-metazoan.owl
 # The pipeline starts by removing all the disjointness axioms, because
 # many external ontologies do not adhere to all Uberon constraints.
 TAXON_GCI_RELS = RO:0002202 RO:0002496 RO:0002497 BFO:0000051
-MERGESPECIES_OPTS = -d -x -g
+MERGESPECIES_OPTS = --remove-declarations --extended-translation --translate-gcas
 .PRECIOUS: $(TMPDIR)/composite-%.owl
 $(TMPDIR)/composite-%.owl: $(TMPDIR)/collected-%.owl $(TMPDIR)/plugins/uberon.jar
 	$(ROBOT) remove -i $< --axioms "DisjointClasses DisjointUnion" \
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -129,7 +129,7 @@ $(TMPDIR)/plugins/sssom.jar:
 # Ditto for the specific Uberon plugin
 $(TMPDIR)/plugins/uberon.jar:
 	mkdir -p $(TMPDIR)/plugins
-	curl -L -o $@ https://github.com/gouttegd/uberon-robot-plugin/releases/download/uberon-robot-plugin-0.1.0/uberon.jar
+	curl -L -o $@ https://github.com/gouttegd/uberon-robot-plugin/releases/download/uberon-robot-plugin-0.2.0/uberon.jar
 
 
 # ----------------------------------------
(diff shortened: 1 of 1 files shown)


gouttegd:
To better understand what those new options are doing:

Let’s assume that `FBbt:X` is mapped to `UBERON:X`. This means that, when creating `composite-metazoan` (or any composite-* product that includes FBbt, such as `composite-drosophila`), any axiom that refers to `FBbt:X` (i.e., has `FBbt:X` in its signature) is “translated” so that `FBbt:X` becomes the class expression `UBERON:X and (part_of some NCBITaxon:7227)`.

a) Until now, even after all axioms referring to `FBbt:X` have been so translated, the class declaration axiom for `FBbt:X` would remain in the resulting merged ontology, even though it would not be referred to from anywhere. This means we would end with a lot of “dangling” classes just under `owl:Nothing`. Option `-d` allows to drop those declaration axioms, resulting in a slightly cleaner product.

b) Until now, general class axioms (GCAs) that refer to `FBbt:X` would _not_ be translated. So if we have, for example, the following GCA:

```
FBbt:X and (part_of some FBbt:Y) SubClassOf: part_of some FBbt:Z
```

then ever after `FBbt:X` has been merged, that axiom would remain as it is, referring to a class that may still exist (if the declaration has not been dropped), but which has been stripped of any meaning (since it has been merged into `UBERON:X`).

The `-g` option allows to translate the general class axioms as any other axiom, so that the axiom above becomes:

```
(UBERON:X and (part_of some NCBITaxon:7227)) and (part_of some FBbt:Y) SubClassOf: part_of some FBbt:Z
```

c) Until now, only _ObjectSomeValuesFrom_ class expressions are translated. If an axiom is referring to a merged class in any other type of expression (e.g., an mix of _ObjectIntersectionOf_ and _ObjectSomeValuesFrom_, as in the way we define endocrine cells in CL: `cell and (capable_of some ('hormone secretion' and (has_target_end_location some 'circulatory system')))`), that axiom is simply dropped from the merged ontology.

Option `-x` allows to translate more class expressions than just _ObjectSomeValuesFrom_, reducing the loss of useful axioms in the composite ontology.

matentzn (dismissed):
You and chris are the same man.. I really do not like this use of short CLI parameters, I think huge expressive readable `--` parameters are much better for maintainability.

gouttegd:
> You and chris are the same man

Well, I did spend hours studying the pipelines that he initially designed and code that he initially wrote, so what do you expect? :D

I agree, though – short options are nice when using a tool on the command line, but in a Makefile there’s no reason to avoid long options.

matentzn (approved):
Wonderful.

anitacaron (approved):

PR #3141, opened 2023-12-06
Automatically import Biomappings cross-references

### LINKED ISSUE #3078: Automatically import Uberon-related mappings from Biomappings
Rather than manually adding Biomappings-derived cross-references directly to the -edit file (as in #2950), it should be possible to directly import relevant mappings from the set published by Biomappings in the SSSOM format ([TSV file here](https://w3id.org/biopragmatics/biomappings/sssom/biomappings.sssom.tsv), [metadata here](https://w3id.org/biopragmatics/biomappings/sssom/biomappings.sssom.yml)).

We are already doing that for the FBbt-provided mapping set and #3061 will (among other things) generalize the process to make it easy to re-use any externally provided mapping set (2b5808a5aa1c98cc642f6868f87f4e103f72b56a), so once #3061 is merged adding Biomappings as a source should be pretty straightforward.

gouttegd:
This PR adds [Biomappings](https://biopragmatics.github.io/biomappings/) as an “external SSSOM provider”, allowing to automatically import Uberon-related mappings from Biomappings into Uberon as cross-references.

Roughly, we fetch the SSSOM mapping set exported from Biomappings, filter it to only keep the mappings pertaining to Uberon (strictly speaking this is not necessary, but we do it because of the sheer size of the Biomappings set: we commit the fetched set to the repository as for all the external resources, and we don’t want to commit a ~10MB file which mostly contains things unrelated to Uberon), and then use it along with all the other mapping sets to generate the `components/mappings.owl` component, which contains a user-friendly representation of all the mappings as cross-references.

closes #3078

### DIFF
--- a/src/ontology/components/mappings.owl
+++ b/src/ontology/components/mappings.owl
@@ -1,1078 +1,5958 @@
 <?xml version="1.0"?>
-<rdf:RDF xmlns="http://www.w3.org/2002/07/owl#"
-  xml:base="http://www.w3.org/2002/07/owl"
-  xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-  xmlns:oboInOwl="http://www.geneontology.org/formats/oboInOwl#">
-  <Ontology rdf:about="http://purl.obolibrary.org/obo/uberon/components/mappings.owl"/>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0000468">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000001</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000002">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000002</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0000914">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000003</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000004">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000004</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000005">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000005</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000006">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000006</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000007">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000007</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000008">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000008</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000009">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000009</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000011">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000011</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000014">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000014</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000015">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000015</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000016">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000016</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000017">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000017</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000018">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000018</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000019">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000019</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000020">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000020</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000021">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000021</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000029">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000029</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000030">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000030</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0000920">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000038</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0003125">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000042</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000046">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000046</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0000922">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000052</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0009571">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000093</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0010302">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000095</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000096">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000096</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000097">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000097</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000104">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000104</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0000923">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000110</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0000924">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000111</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000112">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000112</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000119">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000119</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0000931">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000123</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0000925">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000125</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0000926">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000126</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000128">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000128</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000130">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000130</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0000927">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000136</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000137">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000137</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000154">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000154</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_0008816">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000155</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000157">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000157</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000158">
-    <oboInOwl:hasDbXref rdf:datatype="http://www.w3.org/2001/XMLSchema#string">FBbt:00000158</oboInOwl:hasDbXref>
-  </Class>
-  <Class rdf:about="http://purl.obolibrary.org/obo/UBERON_6000160">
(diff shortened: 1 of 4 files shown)


gouttegd:
@cthoyt FYI.

src/ontology/uberon-edit.obo
```diff
@@ -786,7 +786,6 @@ synonym: "Komplexauge" RELATED [BTO:0001921]
 synonym: "zusammengesetztes Auge" RELATED [BTO:0001921]
 xref: BTO:0001921
 xref: HAO:0000217
-xref: IDOMAL:0002421 {http://purl.org/dc/terms/contributor="https://orcid.org/0000-0003-4423-4370"}
```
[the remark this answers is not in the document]
anitacaron: They're on the sssom tsv file now.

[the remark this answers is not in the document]
gouttegd: Yep, so in the end they end up in the `components/mappings.owl` component, so they don’t need to be in the `-edit` file anymore.

src/ontology/uberon.Makefile
```diff
@@ -1282,6 +1282,24 @@ ifeq ($(strip $(IMP)),true)
...
+# metadata in a separate file. We assemble a single YAML+TSV file and
+# we also filter out anything that has nothing to do with Uberon: the
+# set is quite large and we commit it to the repository, so we want to
+# keep it to the minimum stuff we actually need.
+# (We do the filtering with sed because "sssom dosql" is WAY too slow
+# for what it does.)
+mappings/biomappings-mappings.sssom.tsv: $(TMPDIR)/biomappings.sssom.tsv \
+					 $(TMPDIR)/biomappings.sssom.yml
+	sed 's/^/#/' $(TMPDIR)/biomappings.sssom.yml > $@
+	sed -n '1p; /UBERON:/p' $(TMPDIR)/biomappings.sssom.tsv >> $@
```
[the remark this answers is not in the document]
gouttegd: No, `sssom-py` is already included in the ODK anyway, so that is not the problem.

The problem is that `sssom-py` is just excruciatingly slow, sorry. Just assembling a single YAML+TSV file (which is necessary because `sssom dosql` cannot use a mapping set with non-embedded metadata) takes (on my machine) 20 seconds:

```sh
$ time sssom parse -I tsv biomappings.sssom.tsv -m biomappings.sssom.yml -o combined-biomappings.sssom.tsv
21.57s user 0.25s system 107% cpu 20.334 total
```

Then filtering out everything that is not relevant for Uberon takes 30 more seconds:

```sh
$ time sssom dosql -Q "SELECT * FROM df1 WHERE (subject_id LIKE 'UBERON:%' OR object_id LIKE 'UBERON:%')" -o filtered.sssom.tsv combined.sssom.tsv
28.24s user 0.68s system 108% cpu 27.409 total
```

`sed` does that in less than one second.

In the future, when the next release of the ODK (which includes SSSOM-Java’s `sssom-cli`), we will be able to do everything in a single step which will also take less than one second:

```sh
time sssom-cli --prefix=UBERON=http://purl.obolibrary.org/obo/UBERON_ -R "(subject==UBERON:* || object==UBERON:*) -> include()" -i biomappings.sssom.tsv -o filtered.sssom.tsv
0.79s user 0.11s system 281% cpu 0.818 total
```

src/ontology/uberon-edit.obo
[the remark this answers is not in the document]
matentzn: Thanks!

src/ontology/uberon.Makefile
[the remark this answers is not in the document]
matentzn: Which version of sssom-py?

src/ontology/uberon.Makefile
```diff
@@ -1429,9 +1447,8 @@ $(COMPONENTSDIR)/hra_depiction_3d_images.owl: $(TMPDIR)/hra_depiction_3d_images.
 $(COMPONENTSDIR)/mappings.owl: $(SRC) $(EXTERNAL_SSSOM_SETS) $(TMPDIR)/plugins/sssom.jar
 	$(ROBOT) sssom:inject -i $< \
 		              $(foreach set, $(EXTERNAL_SSSOM_SETS), --sssom $(set)) \
-		              --invert --only-subject-in UBERON \
-		              --check-subject --drop-duplicate-objects \
-		              --hasdbxref --no-merge --bridge-file $@ \
+		              --ruleset $(SCRIPTSDIR)/mappings-to-xrefs.rules \
+		              --no-merge --bridge-file $@ \
```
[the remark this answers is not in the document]
gouttegd: You can have a look at the `components/mappings.owl` file to see the results of this command. :)

src/ontology/uberon.Makefile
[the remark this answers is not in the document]
gouttegd: 0.3.41.

Latest 0.4.0 is better but still decidedly not impressive:

```sh
$ time sssom parse -I tsv biomappings.sssom.tsv -m biomappings.sssom.yml -o combined.sssom.tsv
/Users/dpg44/Library/Python/3.11/lib/python/site-packages/sssom/util.py:164: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  non_double_cols.replace(np.nan, "", inplace=True)
14.53s user 0.30s system 117% cpu 13.239 total
```

~15s to catenate the metadata and the TSV… meh.


```sh
$ time sssom dosql -Q "SELECT * FROM df1 WHERE (subject_id LIKE 'UBERON:%' OR object_id LIKE 'UBERON:%')" -o filtered.sssom.tsv combined.sssom.tsv
/Users/dpg44/Library/Python/3.11/lib/python/site-packages/sssom/util.py:164: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  non_double_cols.replace(np.nan, "", inplace=True)
17.10s user 0.33s system 127% cpu 19.835 total
```

~20s for the filtering step.

Overall a ~2⨉ performance increase. I applaud the `sssom-py` developers for achieving that with Python, but when you start from such abysmal performances to begin with, even a 2⨉ increase is not great. :(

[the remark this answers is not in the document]
matentzn: Fair enough, although for a pipeline that runs 1 hour 15 seconds is not really that much of a deal in my humble view :P No worries, ignore me.

matentzn (dismissed):
All questions addressed!

src/ontology/uberon.Makefile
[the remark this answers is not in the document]
matentzn: Yes but you are trading an opportunity for standardising a workflow using a hack; there is a cost in that too; for example, there is a risk that the two pieces "don't fit together" (curie map is outdated etc), and you would never know. Its also less maintainable because there are 100 ways of doing this, and everyone could be doing it slightly differently creating cognitive overhead. I am not saying your judgement is wrong, I am just saying it that I would not consider it right to say that "runtime" is the only factor, and there are no tradeoffs.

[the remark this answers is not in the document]
gouttegd: 1. You can't complain that people are using Unix-level hacks to manipulate SSSOM file when you specifically designed the format to he amenable to this kind of hacks (quoting the SSSOM spec: “Additionally, it is a simple unix-level or language-level operation to filter these as a pre-processing in a robust fashion.”)

2. It's a _temporary_ hack until the next ODK version is available. As I said above, the next ODK will natively include my `sssom-cli` tool, which will allow to perform the desired operations (catenating the metadata and the mappings themselves, and filtering the non-Uberon mappings) in a completely “non-hacky” way in 1/20th of the time needed by `sssom-py`. In fact I _could_ have opted to manually download `sssom-cli` in this PR to use it straigh away (that would still have been faster than using `sssom-py`, even factoring on the downloading time...), but I decided it was not worth it.

src/ontology/components/mappings.owl
```diff
@@ -28,6 +29,24 @@
     
 
 
+    <!-- https://w3id.org/sssom/author_id -->
+
+    <owl:AnnotationProperty rdf:about="https://w3id.org/sssom/author_id"/>
```
cthoyt, line 34: why is this its own property? shouldn't this be dc:creator or dc:contributor?

src/ontology/components/mappings.owl
```diff
@@ -39,1071 +58,5898 @@
...
         <oboInOwl:hasDbXref>FBbt:00005098</oboInOwl:hasDbXref>
     </rdf:Description>
+    <owl:Axiom>
+        <owl:annotatedSource rdf:resource="http://purl.obolibrary.org/obo/UBERON_0000010"/>
+        <owl:annotatedProperty rdf:resource="http://www.geneontology.org/formats/oboInOwl#hasDbXref"/>
+        <owl:annotatedTarget>FBbt:00005098</owl:annotatedTarget>
+        <sssom:mapping_justification rdf:resource="https://w3id.org/semapv/vocab/ManualMappingCuration"/>
+    </owl:Axiom>
+    <rdf:Description rdf:about="http://purl.obolibrary.org/obo/UBERON_0000016">
+        <oboInOwl:hasDbXref>https://meshb.nlm.nih.gov/record/ui?ui=D007515</oboInOwl:hasDbXref>
```
cthoyt, line 77: this doesn't seem right

src/ontology/components/mappings.owl
gouttegd: Because that’s what the [SSSOM spec mandates](https://mapping-commons.github.io/sssom/spec/#predicate-is-an-annotation-property):

> All mapping level metadata (sssomMetadata) gets converted into OWLAnnotation objects which are materialised as axiom annotations on the mapping annotation assertion [...]
>
> ```
> AnnotationAssertion(sssomMetadata P, S, O)
> ```
>
> Where sssomMetadata is a sequence of OWL Annotations objects like:
>
> ```
> Annotation(Q1,V1) Annotation(Q2,V2) ... Annotation(Qn,Vn)
> ```
>
> where Qi is a SSSOM metadata element and Vi is an annotation value.
> [...]
> Example:
>
> ```
> AnnotationAssertion(Annotation(sssom:creator_id <https://orcid.org/0000-0002-7356-1779>) Annotation(sssom:mapping_justification semapv:LexicalMatching) skos:exactMatch <http://purl.obolibrary.org/obo/HP_0009894> <http://purl.obolibrary.org/obo/MP_0000019>)
> ```

cthoyt: oh boy...

src/ontology/components/mappings.owl
gouttegd: It’s not written as a CURIE because ROBOT is not aware of the `https://meshb.nlm.nih.gov/record/ui?=` URL prefix, so it doesn’t know how to condense  such URLs.

We need to add `--add-prefix "MESH: https://meshb.nlm.nih.gov/record/ui?ui=` to the ROBOT call if we want such URLs to be written as CURIEs.

src/ontology/components/mappings.owl
matentzn: Some of this have mappings to standard properties, like creator_id: https://github.com/mapping-commons/sssom/blob/master/src/sssom_schema/schema/sssom_schema.yaml#L317

Feel free to request more such mappings to standard properties. But "author" is, for example, more fine-grained than dcterms:contributor so unlikely to be a suitable match.

src/ontology/components/mappings.owl
matentzn: This is the only non-curie xref, I would strongly suggest to add the prefix. 

But I am a bit surprised what ROBOT has to do with this, as hasDbXref are not of type "IRI" but of type string; I assume your plugin has to compress? In this case, shouldn't it use the prefixmap supplied by the sssom file (`biomappings.sssom.yml`)?

gouttegd: > shouldn't it use the prefixmap supplied by the sssom file (biomappings.sssom.yml)?

It can’t. The prefix map is _not_ part of the SSSOM data model (maybe it should, but it is not). Once the SSSOM file has been read, the code downstream of the SSSOM reader has no access to whatever the original prefix map was.

matentzn: Hmmm. This does not seem ideal. The fact that the `curie_map` is not mentioned in the sssom data model needs to be remedied then, as sssom-py for example just assumes that it is part for the "format" without being part of the "model" - I guess this prevents it from being "passed downstream" according to your thinking.

So internally, how does sssom-java keep track of a `subject_id`? just a URI?

gouttegd: > So internally, how does sssom-java keep track of a subject_id? just a URI?

Yes. The prefix map provided in the SSSOM file is used internally by the reader to expand all CURIEs on the fly. Once the reader is done, the client only gets full-length IRIs. This is the only sane way to deal with CURIEs. Expand them as soon as possible (just after reading) and condense them as late as possible (just before writing).

And in this particular situation here, where we generate the `mappings.owl` component from the mappings, I see no reason why the original prefix map should be used, even if the model gave us access to it. Doing otherwise would only open a can of worms, for example: we are merging several mapping sets here (the one from Biomappings is only one among others), they all come with their own prefix maps – how would we solve any conflict between two prefix maps? Should the first entry take precedence? The last entry? Why?

We don’t have to worry about that because all “input” prefix maps are discarded after use. If a prefix map has to be used for writing, it has to be an explicitly specified one, at the discretion of the user.

gouttegd: >  sssom-py for example just assumes that it is part for the "format" without being part of the "model"

Yes, that’s exactly the case. The _format_ has a CURIE map, but the _mapping set object_ does not. The CURIE map is only an internal detail of the format. Once the reader has done its job of reading the file, it gives you a _mapping set object_, with all the fields defined in the spec – and nothing more, so the map is _not_ part of it.

balhoff:
Where does this prefix definition come from?

```
mesh: https://meshb.nlm.nih.gov/record/ui?ui=
```

MeSH has a well-defined IRI namespace: `mesh: <http://id.nlm.nih.gov/mesh/>`

gouttegd:
> Where does this prefix definition come from?

From the [Biomappings SSSOM mapping set](https://raw.githubusercontent.com/biopragmatics/biomappings/master/docs/_data/sssom/biomappings.sssom.yml):

```
#  mesh: https://meshb.nlm.nih.gov/record/ui?ui=
```

src/ontology/components/mappings.owl
matentzn: > Doing otherwise would only open a can of worms, for example: we are merging several mapping sets here (the one from Biomappings is only one among others), they all come with their own prefix maps – how would we solve any conflict between two prefix maps? Should the first entry take precedence? The last entry? Why?

We (in the sense of @cthoyt) have started developing some very well documented processes for prefix standardisation, so its not entirely wild west, but I do agree that there are some kind of conflicts that are hard to overcome. For an ontology project I guess you could argue that mapping sets go in, get expanded, but then get contracted again based on an internally managed prefixmap like https://bioregistry.io/context/obo, this is what we do in sssom-py. Its not enough to use the prefixmaps to go to URI syntax, because different sources will declare different URI expansions for the same exact prefix. 

I still maintain that we should have a prefix standardisation process in place which (roughly) does this:

1. expand CURIEs to URIs according to provided prefixmap (what you do)
2. _always_ compress URIs to CURIEs when writing xrefs, and failing if contraction fails due to missing prefix
3. using https://bioregistry.io/context/obo as the standard map for compression, and add any missing (I doubt there will be many) URIs prefixes in there (its a super fast turnaround). A compromise here would be to keep a small "internal" extension map in the repo to quickly add unsupported uri prefixes to that, and chain this internal prefixmap with the obo one.

gouttegd: > For an ontology project I guess you could argue that mapping sets go in, get expanded, but then get contracted again based on an internally managed prefixmap like https://bioregistry.io/context/obo, this is what we do in sssom-py

“Get contracted again” when? Just before writing? Just after reading?

> Its not enough to use the prefixmaps to go to URI syntax

What the heck is the prefix map good for then?

> because different sources will declare different URI expansions for the same exact prefix.

If one source expands `mesh:D123456` to `https://meshb.nlm.nih.gov/record/ui?ui=D123456` and another expands `mesh:D123456` to `http://id.nlm.nih.gov/mesh/D123456`, then they are not referring to the same entities. Maybe they _wanted_ to refer to the same entities, but that’s their problem to fix, it’s out of scope for SSSOM. As I said: garbage in, garbage out.

> 1. expand CURIEs to URIs according to provided prefixmap (what you do)

OK.

> 2. always compress URIs to CURIEs when writing xrefs, and failing if contraction fails due to missing prefix

OK. `sssom-inject` does not fail when a prefix is missing (it writes the uncontracted IRI instead), but that can be configurable.

> 3. using https://bioregistry.io/context/obo as the standard map for compression, and add any missing

OK, so that’s what I said above: any “input” prefix map is discarded, and if contraction is wanted upon writing then an “output” prefix map has to be explicitly provided (which can be done with prefix declarations in SSSOM/T). From the implementation’s point of view it doesn’t matter whether that map is the one from `https://bioregistry.io/context/obo` or any other.

matentzn: > If one source expands mesh:D123456 to https://meshb.nlm.nih.gov/record/ui?ui=D123456 and another expands mesh:D123456 to http://id.nlm.nih.gov/mesh/D123456, then they are not referring to the same entities

I will respond to the rest tomorrow, but this sentence here ⬆️ is at least debatable. We had long years of discussions on this, and the main problem is that there are quite a few resources that do not have a standardised URI expansion  which led to the wild west out there. The main point of bioregistry (from our POV) is to group synonymous expansions into the same "identifier spaces", so they can be standardised. So in fact, `https://meshb.nlm.nih.gov/record/ui?ui=D123456` and `http://id.nlm.nih.gov/mesh/D123456` are the exact same thing. Or rather, to be very precise, insisting on them being _not_ the same thing because they are not the same URI is impractical for most, of not all, use cases. There is huge disagreement on a lot of this, but well, I will just push the practical view over the "formally correct" one.

gouttegd: > quite a few resources that do not have a standardised URI expansion which led to the wild west out there. The main point of bioregistry (from our POV) is to group synonymous expansions into the same "identifier spaces", so they can be standardised

That’s the main point of bioregistry, OK. Is that the main point, or even _a_ point, of SSSOM? Is it up to a SSSOM implementation to fix that mess?

Before you answer “yes”, please go read again the SSSOM specification: there is no mention of that problem _anywhere_. I defy anyone to read the spec and to conclude that they must be prepared to deal with the fact that maybe, a subject ID of `https://meshb.nlm.nih.gov/record/ui?ui=D123456` in one mapping could refer to the same entity as `http://id.nlm.nih.gov/mesh/D123456` in another mapping, and that therefore you may need to “reconcile” IDs before doing anything.

If you want(ed) all implementations to implement reconciliation, you should have made that clear, sorry. I wrote SSSOM-Java without that particular problem in mind, and changing that now would basically mean having to rewrite the entire thing from scratch. Not going to happen. Reconciliation of messy IDs is forever going to be out of scope for SSSOM-Java.

Reconciliation can always be performed upstream of SSSOM-Java if/when needed.

gouttegd: > insisting on them being not the same thing because they are not the same URI is impractical for most, of not all, use cases. 

It is practical enough for all the use cases I care about. For example, it was practical enough to rebuild the bridge pipeline in Uberon, which is the primary reason why I wrote SSSOM-Java in the first place.

That it is impractical for the use-case of creating cross-references from the Biomappings set is the least of my concerns.

gouttegd: By the way:

> but then get contracted again based on an internally managed prefixmap like https://bioregistry.io/context/obo, **this is what we do in sssom-py**

Is that really what you do? Because parsing the Biomappings set with `sssom-py` (last version, 0.4.2) still yields `mesh` CURIEs that expands to `https://meshb.nlm.nih.gov/record/ui?ui=`, despite the OBO context supposedly associating MESH identifiers to the `http://id.nlm.nih.gov/mesh/` prefix.

matentzn: > That’s the main point of bioregistry, OK. Is that the main point, or even a point, of SSSOM? Is it up to a SSSOM implementation to fix that mess?
Before you answer “yes”, please go read again the SSSOM specification: there is no mention of that problem anywhere. I defy anyone to read the spec and to conclude that they must be prepared to deal with the fact that maybe, a subject ID of https://meshb.nlm.nih.gov/record/ui?ui=D123456 in one mapping could refer to the same entity as http://id.nlm.nih.gov/mesh/D123456 in another mapping, and that therefore you may need to “reconcile” IDs before doing anything.

Sorry I didn't mean to say this "should be solved by SSSOM" - it is, as you say, out of scope! The argument here is  on the level of the ETL pipeline (Uberon), and here I think it is justifiable to say that we reconcile prefixes before republishing them for our users.

> If you want(ed) all implementations to implement reconciliation, you should have made that clear, sorry. I wrote SSSOM-Java without that particular problem in mind, and changing that now would basically mean having to rewrite the entire thing from scratch. Not going to happen. Reconciliation of messy IDs is forever going to be out of scope for SSSOM-Java.

No, reconciliation should be a special purpose convenience method, not a part of the spec! I never meant to suggest that.

> It is practical enough for all the use cases I care about. For example, it was practical enough to rebuild the bridge pipeline in Uberon, which is the primary reason why I wrote SSSOM-Java in the first place.

In my understanding of the problem the bridge file generation is a bit unrelated to the question of whether "two URIs can refer to the same entity". This PR here is about redistributing xrefs, not the bridge pipeline, right? So the argument is purely about standardising whatever we redistribute, and again, yes, this happens downstream (or ideally upstream, but certainly offstream) of SSSOM.

> ..parsing the Biomappings set with sssom-py (last version, 0.4.2) still yields mesh CURIEs that expands to https://meshb.nlm.nih.gov/record/ui?ui=, despite the OBO context supposedly associating MESH identifiers to the http://id.nlm.nih.gov/mesh/ prefix.

This is a mistake in the prefixmap, I will fix it. (The prefix map is unfortunately an eternal work in progress).

matentzn:
> MeSH has a well-defined IRI namespace: mesh: <http://id.nlm.nih.gov/mesh/>

@balhoff in sssom-py at least this has already been fixed: https://github.com/mapping-commons/sssom-py/blob/f4256047f75f8d0a286b7255f76ab884d16e049d/src/sssom/obo.epm.json#L16980

Hopefully this will trickle down soon!

src/ontology/components/mappings.owl
gouttegd: > Sorry I didn't mean to say this "should be solved by SSSOM" - it is, as you say, out of scope! […] reconciliation should be a special purpose convenience method

So in the scope of SSSOM and unless you’re running in a special mode in which you explicitly want to mangle identifiers, if two identical CURIEs expand to different full-length identifiers (because of a different prefix map), the identifiers are _not_ the same and should not be treated as if they were the same. Good.

>  I think it is justifiable to say that we reconcile prefixes before republishing them for our users

That’s debatable (who are we to say which URL prefix is to be used? for example, the Allen institute now uses `https://purl.brain-bican.org/ontology/xxxo/XXX_` as its prefix for its brain ontologies, why would we say that “no no no, we know better than the folks at the Allen, the correct prefix is `http://purl.obolibrary.org/obo/XXX_`”?), but okay, if it’s an explicit move and not something done silently by the implementation without any user opt-in. I can implement  such a “reconciliation” as an option to `sssom-cli`.


> In my understanding of the problem the bridge file generation is a bit unrelated to the question of whether "two URIs can refer to the same entity".

That question is _very_ relevant for the generation of the bridges. The bridges are OWL files, and in OWL (as in RDF), only full-length identifiers matter, and they are compared as opaque strings without any notion of “URL prefix“. Maybe outside of OWL/RDF you can pretend that `https://meshb.nlm.nih.gov/record/ui?ui=D123456` in one file and `http://id.nlm.nih.gov/mesh/D123456` in another file are actually the same thing (because magic), but `<owl:Class rdf:about="https://meshb.nlm.nih.gov/record/ui?ui=D123456">` and `<owl:Class rdf:about="http://id.nlm.nih.gov/mesh/D123456">` are _never_ going to be considered the same, unless you have an explicit equivalence axiom.

So it is very important to be sure we are dealing with the correct URL prefix when we generate the bridges, if you don’t want to generate a bridge that connect to a different island than the one intended. Right now, we can be sure of that because we only generate the bridges from two kinds of mappings:

*) mappings that we (Uberon editors) maintain ourselves in Uberon: we maintain them, so we can check that we are using the correct prefixes;
*) mappings from CL, FBbt, and ZFA: we don’t have control over them (well, I control the FBbt ones but let’s assume I don’t) but they are mappings between OBO resources only (CL-to-Uberon, FBbt-to-Uberon, ZFA-to-Uberon), so it’s all `http://purl.obolibrary.org/obo` and there is little risk of anything fishy happening with the identifiers (apart from an always possible silly mistake).

> This PR here is about redistributing xrefs, not the bridge pipeline, right?

This PR, yes. This discussion, no. For distributing xrefs, this entire discussion is actually irrelevant, since the xrefs are pure CURIEs whose interpretation is entirely left to the user, there is no associated prefix name declaration. Which makes those cross-references much less useful than they could be, but well, that’s what they have always been – we are generating and distributing them only for backwards compatibility for the sake of people using tools that know nothing else than `oboInOwl:hasDbXref`.

The only real problem in this PR was the fact that the `https://meshb.nlm.nih.gov/record/ui?ui=` was not declared, which prevented the `sssom-inject` command from shortening `https://meshb.nlm.nih.gov/record/ui?ui=D123456` into `MESH:D123456`. This is now fixed and an option has been added to raise an error if an undeclared prefix sneaks in again.

Everything else is a more general discussion that could (should?) have happened elsewhere.

gouttegd:
> in sssom-py at least this has already been fixed: https://github.com/mapping-commons/sssom-py/blob/f4256047f75f8d0a286b7255f76ab884d16e049d/src/sssom/obo.epm.json#L16980
>
> Hopefully this will trickle down soon!

So is this file supposed to be the reference extended prefix map to be used by the OBO world? The one you would want to use to “reconcile” prefixes in Uberon before redistributing mappings?

matentzn (approved):
Looks great to me!

matentzn:
> So is this file supposed to be the reference extended prefix map to be used by the OBO world? The one you would want to use to “reconcile” prefixes in Uberon before redistributing mappings?

In my dreamworld, yes, but this is _not at all_ agreed upon by anyone. Just started somewhere, and once there are more stakeholders, there would need to be some more governance. For now, just a vision.

gouttegd:
It’s not great that it is hidden deep in an obscure repository (especially a repository that, at face value, appears to have nothing to do with OBO)…

I was about to suggest at least two ways of making it more easily available ((1) provide it in the ODK, e.g. in `/resources/obo.epm.json` where all ODK tools could access it instantly; (2) publish it under a PURL like `http://purl.obolibrary.org/obo/epm.json`), but I suppose this would require more adhesion to the idea?

balhoff:
> > MeSH has a well-defined IRI namespace: mesh: http://id.nlm.nih.gov/mesh/
> 
> @balhoff in sssom-py at least this has already been fixed: https://github.com/mapping-commons/sssom-py/blob/f4256047f75f8d0a286b7255f76ab884d16e049d/src/sssom/obo.epm.json#L16980
> 
> Hopefully this will trickle down soon!

Thanks, sorry for jumping in and stirring things up. 😅

PR #3146, opened 2023-12-08
Fix `refresh-external-resources`.

gouttegd:
This PR removes a reference to the old `allen_all` target, which no longer exists now that we changed the way we import the Allen ontologies. Instead, we make sure the Allen ontologies are explicitly listed in `LOCAL_IMPORTS`, which is used to generate the pre-requisites of the `all_local_imports` target. This makes the `refresh-external-resources` command work again.


matentzn (approved):
Lgtm

PR #3152, opened 2023-12-14
Filter and reconcile the Biomappings set using sssom-cli

gouttegd:
This PR is a follow-up to #3141. It particularly addresses [this comment](https://github.com/obophenotype/uberon/pull/3141#discussion_r1419489287) by using the `sssom-cli` tool rather than a `sed`-based hack to filter out non-Uberon mappings.

It also “reconciles” IRI prefixes in the Biomappings set against the OBO-specific extended prefix map, thereby removing the need for a manual fix of the `https://meshb.nlm.nih.gov/record/ui?ui=` MESH IRIs.


gouttegd:
(Cancelling the CI check because the PR needs the `sssom-cli` tool, which is not yet available in the ODK – that’s why it’s a draft PR, and will remain so until the next version of the ODK is released.)

anitacaron:
Doesn't it need to update components/mappings.owl?

gouttegd:
The PR only changes the way our copy of the Biomappings set is generated, not the _contents_ of the set. So the `mappings.owl` component is unchanged after this PR.

anitacaron:
I see many changes in biomappings-mappings.sssom.tsv

gouttegd:
They are format changes only. `sssom-cli` always write a set in canonical form (that’s one of the reasons I use it here), by:

* removing any unused prefix declaration in the prefix map;
* writing all set-level metadata fields and all columns in the order recommended by the SSSOM spec;
* sorting all mappings in a deterministic order.

In addition, the rules used here also ensure that all mappings are always oriented with the Uberon term as the subject (whereas the original Biomappings set contains a mix of mappings where the Uberon term is sometimes the subject, sometimes the object).

But the mappings themselves are unchanged. They are just written more predictably.

anitacaron (approved):

PR #3151, opened 2023-12-14
multicellular organism disjoint from organism substance #2421

ddooley:
Addresses Issue #2421 .  This is in external disjoints file though the two terms are native to Uberon though.

Issue #2421, opened 2022-04-15
multicellular organism and organism substance should be disjoint


ddooley:
It came up in discussion with @cmungall  that currently there is no disjunction between multicellular organism and organism substance.  Since NCBITaxon terms correspond to multicellular organisms, someone can define a 'human heart' as an axiom like:

'human milk': equivalentTo 'milk and 'Homo Sapiens'

and not get a logical contradiction.

**Suggested revision of logical definition**

Can we make multicellular organism and organism substance disjoint to trigger unsatisfiability?

shawntanzk:
added to uberon agenda, will assign you

bvarner-ebi:
@ddooley, can you advise what is the ultimate goal that would like to solve? By making disjoint as you mention above, there is concern this would cause downstream issues. There are concerns with scope related to food (FoodOn) vs anatomy terms (Uberon).

ddooley:
I think it's a basic quality control check - a distinction between a whole, and a proper part.  We need to know which references point to whole organisms, and which do not. One should not be confused with the other, so this would trip an error if that happened in UBERON or someone's reuse of it.

However, do things get complicated if, as Chris told me, a reference to an UBERON part, such as heart, logically entails connection to some whole organism in an abstract (dead or alive) sense? I really need an answer to that question - in FoodOn we need to be able to talk about a separated (severed / butchered) heart, one that does not entail a whole connected organism.

cmungall:
Let's keep this issues separate

This issue is about the request to make substance and organism disjoint. This is completely reasonable and straightforward. I am surprised we don't have this already, actually.

I don't think there are any concerns. I can't think of any exceptions. there are no organisms made of gas, liquid or other substances created by the body. even an ingested animal ceases at some point to become a multicellular organism and is broken down into substances.

Note that X disjoint with Y simply means there is nothing that instantiates X and Y. There are no further entailments to do with connectedness or parthood.

We already have other issues for the butchered heart use case, but that is completely separate. I think we should grant this request, close this issue, and continue discussing butchered hearts on the other tickets

ddooley:
So I should do an UBERON pull request with the disjoint axiom?

ddooley:
I realize now I don't know how to curate UBERON. Is it the root /uberon.obo file one would edit for disjoints?  I tried to edit and save via Protege, but it reported a "multiple def tags not allowed" issue which suggests Uberon team is using some other editor, or text edit?

anitacaron:
[...]

PR #3153, opened 2023-12-15
Add permission to post comment in forked PR from GH Action 

anitacaron:
More details on https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target and https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

### DIFF
--- a/.github/workflows/qc.yml
+++ b/.github/workflows/qc.yml
@@ -9,6 +9,8 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  pull_request_target:
+    branches: [ master ]
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
@@ -20,7 +22,9 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
     container: obolibrary/odkfull:v1.4.3
-
+    permissions:
+      contents: read
+      pull-requests: write
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
--- a/.github/workflows/qc.yml
+++ b/.github/workflows/qc.yml
@@ -9,8 +9,6 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
-  pull_request_target:
-    branches: [ master ]
 
   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:


anitacaron:
Related to #2421 and failing QC https://github.com/obophenotype/uberon/pull/3151/checks

.github/workflows/qc.yml
```diff
@@ -9,6 +9,8 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  pull_request_target:
+    branches: [ master ]
```
matentzn, line 13: Given the huge list of `activity types` listed in:

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

Isnt it a bit risky to not write them out? Like: editing the PR text? adding a label?

Maybe add something like:

```
types: [assigned, opened, synchronize, reopened]
```

.github/workflows/qc.yml
```diff
@@ -20,7 +22,9 @@ jobs:
     # The type of runner that the job will run on
     runs-on: ubuntu-latest
     container: obolibrary/odkfull:v1.4.3
-
+    permissions:
+      contents: read
+      pull-requests: write
```
matentzn, line 27: what does this do?

.github/workflows/qc.yml
```diff
@@ -9,6 +9,8 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  pull_request_target:
```
gouttegd, line 12: From my understanding of [the documentation on the `pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target), this is not suitable:

> This event runs in the context of the base of the pull request, rather than in the context of the merge commit

This means that we check out the repository to run the test suite, the commit that will be checked out is the **base** from which the foreign branch was forked, _not_ the **tip** of the foreign branch. So the test suite will be triggered, but it will only test contents that were already present in the repository, **without** the changes that the PR is proposing to introduce!

.github/workflows/qc.yml
anitacaron: give permission to forked prs to write on it via a gh action

.github/workflows/qc.yml
anitacaron: Yes, but there's this sentence here.

> This event allows your workflow to do things like label or comment on pull requests from forks.

gouttegd: But what would be the point of allowing to comment if the test suite runs against the base of the PR? It wouldn’t test anything useful – none of the changes from the PR would be tested.

anitacaron: Ok, only need to give permission then.

.github/workflows/qc.yml
anitacaron: No need to add this in the end.

gouttegd (approved):

matentzn (approved):

PR #3154, opened 2023-12-18
Use SSSOM to correct local imports when needed.

gouttegd:
This is a small follow-up to PR #3129 and more precisely [this comment](https://github.com/obophenotype/uberon/pull/3129#discussion_r1411832383).

When a local import needs some corrections (for use of old-style properties, and in the specific case of EMAPA, to force it to use MmusDv terms), we do that using a SSSOM mapping set as the source of truth for the corrections, rather than the ad-hoc TSV format used by the standard rename command of ROBOT.


cmungall (approved):
What's wrong with doing this in perl or long convoluted owltools commands?

j/k, this is how it should always have worked, thank for moving this workflow one important step closer to sanity!

src/ontology/mappings/import-corrections.sssom.tsv
```diff
@@ -0,0 +1,50 @@
...
+#  EMAPA: "http://purl.obolibrary.org/obo/emapa#"
+#  IAO: "http://purl.obolibrary.org/obo/IAO_"
+#  MA: "http://purl.obolibrary.org/obo/ma#"
+#  MmusDv: "http://purl.obolibrary.org/obo/MmusDv_"
+#  RO: "http://purl.obolibrary.org/obo/RO_"
+#  TS: "http://purl.obolibrary.org/obo/TS_"
+#  XAO: "http://purl.obolibrary.org/obo/xao#"
+#mapping_set_id: "http://purl.obolibrary.org/obo/uberon/import-corrections.sssom.tsv"
+#mapping_set_description: "Terms and properties to replace in imported ontologies."
+#license: "http://creativecommons.org/licenses/by/3.0/"
```
matentzn, line 12: My assumption is that this is "our" work, so I would suggest to use whenever possible CC-0. CC-BY 3 is a convoluted license, even BY-4 is better.

src/ontology/uberon.Makefile
```diff
@@ -424,25 +424,20 @@ imports/local-ssso.owl: mirror/ssso.owl
 # external entities we also need to replace some old-style properties.
 # ----------------------------------------
 # EMAPA also needs translation between TS-style stages to MmusDv stages
-imports/local-emapa.owl: mirror/emapa.owl $(IMPORTDIR)/map-properties.tsv
-	$(ROBOT) rename -i $< --mappings $(IMPORTDIR)/map-properties.tsv \
-		        --allow-missing-entities true --allow-duplicates true \
-		 rename --mappings $(IMPORTDIR)/map-mouse-stages.tsv \
-		        --allow-missing-entities true \
+imports/local-emapa.owl: mirror/emapa.owl mappings/import-corrections.sssom.tsv $(TMPDIR)/plugins/sssom.jar
+	$(ROBOT) sssom:rename -i $< --sssom mappings/import-corrections.sssom.tsv \
```
matentzn, line 428: nice...

matentzn (approved):
I made a small suggestion, but because I don't really mind if you heed it or not, I will approve this excellent looking PR!

src/ontology/mappings/import-corrections.sssom.tsv
gouttegd: I simply used the license of Uberon itself, which happens to be CC-BY 3.0:

```
property_value: dcterms-license http://creativecommons.org/licenses/by/3.0/
```

I have no strong opinion on that particular license versus CC-0, CC-BY-4.0, or even the WTFPL, but I don’t think it is a good idea to have different licenses for the different parts of a same project. (Except maybe for documentation, which may be under a different license than actual code.)

PR #3172, opened 2024-01-12
Add present_in_taxon statements 

balhoff:
Add present_in_taxon statements for male and female external genitalia to ensure they work in both mammals and insects. These terms are used in the AISM ontology; this will ensure that the usage remains valid as the ontology changes.

cc @JCGiron @sergeitarasov


JCGiron:
I would also edit the definition to make it clear that "including the penis and scrotum" is just an example and not the whole extent of the usage of this term.

balhoff:
Thanks @JCGiron, I made a small tweak to the definition.

cmungall:
This looks good. You should feel free to propose a more significantly
altered definition that is less biased to mammals (I am not a fan of the
GO-style definitions that say "an example of this is found in <narrow
taxon>" when the true applicability is much broader).

@balhoff I have some ideas about more powerful present-in axioms that would
entail some of these more specific statements

anitacaron:
#gogoeditdiff

anitacaron:
Hi @balhoff, let me know if you need someone to approve this.

balhoff:
@anitacaron yes, please!

src/ontology/uberon-edit.obo
```diff
@@ -80988,6 +80988,8 @@ intersection_of: part_of UBERON:0000079 ! male reproductive system
 disjoint_from: UBERON:0005056 ! external female genitalia
 disjoint_from: UBERON:0009196 ! indifferent external genitalia
 relationship: contributes_to_morphology_of UBERON:0000079 ! male reproductive system
+relationship: present_in_taxon NCBITaxon:50557 ! Insecta
+relationship: present_in_taxon NCBITaxon:9606 ! Homo sapiens
```
aleixpuigb, line 80992: > Add present_in_taxon statements for male and female external genitalia to ensure they work in both mammals and insects.

Wouldn't it be more accurate to use [Mammalia](http://purl.obolibrary.org/obo/NCBITaxon_40674) instead of Homo sapiens?

src/ontology/uberon-edit.obo
```diff
@@ -80971,7 +80971,7 @@ relationship: adjacent_to UBERON:0002336 ! corpus callosum
 [Term]
 id: UBERON:0004053
 name: external male genitalia
-def: "The external masculine genital organs, including the penis and scrotum." [MP:0009199]
+def: "The external masculine genital organs. In mammals, this includes the penis and scrotum." [MP:0009199]
```
aleixpuigb, line 80974: I would advice to remove the second sentence, since there are several mammals that lack penis and/or scrotum. Also, I agree with Chris that a less mammalian biased definition would be more appropriate.

aleixpuigb (changes requested):
Please, see inline suggestions.

src/ontology/uberon-edit.obo
balhoff: For present_in_taxon it's useful to be as specific as possible, as far as QC goes (present in Mammalia is implied). Any number of other species could be added. But if the intent is just readability that would be different.

src/ontology/uberon-edit.obo
balhoff: I removed the second sentence.

aleixpuigb (approved):

PR #3174, opened 2024-01-15
Add LICENSE file.

gouttegd:
This PR adds the full text of the Creative Commons Attribution 3.0 Unported license in a new LICENSE file at the root of the repository.

This does NOT change the license used by Uberon. Uberon IS already using the CC BY 3.0 Unported license right now (as indicated by an annotation in the -edit file). This commit simply makes the license more obvious.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,319 @@
+Creative Commons Legal Code
+
+Attribution 3.0 Unported
+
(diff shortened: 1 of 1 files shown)


cmungall (approved):

gouttegd:
I’ve cancelled the CI check run. Adding a simple text file at the root of the repository _cannot_ _possibly_ break anything …right?

PR #3177, opened 2024-01-16
Allow to optionally keep disjointness axioms when building composite products.

gouttegd:
All `composite-*` products are built by, among other operations, stripping any disjointness axiom in the merged ontology. This is necessary as most `composite-*` products would be impossible to build otherwise (the `merge-species` step requires the ontology to be consistent, and most of the `collected-*` merges are not because of incompatibilities between Uberon and one or several of the taxon-specific ontologies).

This will not change anytime soon, but we should still at least _aim_ to achieve inter-ontology consistency in the (far) future. This has already been achieved at least for the `composite-drosophila` product (Uberon + CL + FBbt), so it _is_ possible.

This PR makes the removal of the disjointness axioms optional, so one can easily try to build a `composite-*` product with the disjointness axioms still in place. Simply run the pipeline with `COMPOSITE_STRIP_DISJOINTS=false`, e.g.

```sh
sh run.sh make composite-mouse.owl COMPOSITE_STRIP_DISJOINTS=false
```

For most products this is likely to fail, but this gives us an easy way to _test_ building with disjointness axioms in place, so that we can start addressing the inconsistency issues.

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -1160,6 +1160,14 @@ COLLECTED_metazoan_SOURCES =         $(COLLECTED_vertebrate_SOURCES) \
 # Composite pipeline proper
 # ----------------------------------------
 
+# Set to false to keep disjointness axioms in collected/composite
+# products. Such axioms are stripped by default because too many
+# taxon-specific ontologies are inconsistent with Uberon.
+COMPOSITE_STRIP_DISJOINTS=true
+ifeq ($(COMPOSITE_STRIP_DISJOINTS),true)
+COMPOSITE_STRIPPING_COMMAND = remove --axioms "DisjointClasses DisjointUnion"
+endif
+
 # Step 1: Create a "collected" ontology, which is simply a merge of
 # Uberon, CL, and the components listed in the COLLECTED_*_SOURCES
 # variables above.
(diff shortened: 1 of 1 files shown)


matentzn (commented):
This PR does not seem to affect the default behaviour (basically changing nothing in practice). 

Just so I understand -

1. Why even try to build with disjoint (even if it is possible)
2. And if its possible, why not always do it for, say, a specific composite (like drospohila)
3. How do you see this used in practice? People checking out uberon and running a composite build with `COMPOSITE_STRIP_DISJOINTS=false`?

gouttegd:
> This PR does not seem to affect the default behaviour (basically changing nothing in practice).

Yes.

> Why even try to build with disjoint (even if it is possible)?

So that we have a chance to detect the unsats that are caused (or rather, revealed) by the disjointness axioms, so that we can have a go at trying to fix them instead of continuing to pretend that everything is fine.

Forever building without those disjointness axioms is the ontological equivalent of sweeping the dust under the rug and pretending the room is clean.

“Your ontology is inconsistent!
— La la la, I can’t hear you! 🙉”

> And if its possible, why not always do it for, say, a specific composite (like drospohila)

The only composite product we routinely build and publish as part of a release is `composite-metazoan`. All the other products are only built upon demand. No point in cluttering the Makefile for that.

> How do you see this used in practice? People checking out uberon and running a composite build with COMPOSITE_STRIP_DISJOINTS=false?

Exactly. Though in effect, I don’t expect that anyone would have to _check out_ Uberon – the only people who would use that are likely to be Uberon editors and would already have a copy of the repo.

This option is merely a facility offered to Uberon editors (at least those of them actually interested in getting rid of unsats) to temporarily disable the unsats, see what happens, and either recoil in horror at the number of unsats they found, or decide that they can try to fix at least some of them.

matentzn (approved):
LGTM, thanks for the explanations @gouttegd !

PR #3184, opened 2024-01-17
Only trigger QC when any file inside `src` folder is changed

### LINKED ISSUE #3112: Only run QC when any files inside `src` was changed
No need to run QC when updating documentation files, for example, README.md, mkdocs.yml, etc.

anitacaron:
Fixes #3112


matentzn (approved):
Agreed!

PR #3185, opened 2024-01-17
Turn `disjoint_union_over` into a proper ODK component

### LINKED ISSUE #3180: `disjoint_over_union.ofn` should be a proper ODK component
The `disjoint_over_union.ofn` component (in `src/ontology/components`) contains several disjointness axioms.

That component is merged into the main ontology as part of the preprocessing step:

```Makefile
$(OWLSRC): $(SRC) $(COMPONENTSDIR)/disjoint_union_over.ofn
	$(ROBOT) merge -i $< \
			-i $(COMPONENTSDIR)/disjoint_union_over.ofn \
			-i issues/contributor.owl \
		expand --no-expand-term http://purl.obolibrary.org/obo/RO_0002175 \
			-o $@
```

From there, its axioms percolate through the entire pipeline down to the final `uberon.owl` release artefact.

**But** they are _not_ present in the `uberon-base.owl` artefact. That’s because that artefact is obtained through a different route (one that is part of the standard ODK workflows) that does not involve the `$(OWLSRC)` preprocessing step:

```Makefile
$(ONT)-base.owl: $(EDIT_PREPROCESSED) $(OTHER_SRC)
	$(ROBOT_RELEASE_IMPORT_MODE_BASE) \
	$(SHARED_ROBOT_COMMANDS) \
	annotate --link-annotation http://purl.org/dc/elements/1.1/type http://purl.obolibrary.org/obo/IAO_8000001 \
		--ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) \
		--output $@.tmp.owl && mv $@.tmp.owl $@
```

The `$(OTHER_SRC)` variable in that rule references the ODK components so that they get merged into the `-base` product, but because `disjoint_over_union.ofn` is _not_ declared as a proper ODK component, it is excluded from this merge operation.

We should treat `disjoint_over_union.ofn` as a true ODK component instead of merging it in an ad-hoc operation as part of the preprocessing step.

gouttegd:
This PR:

* rename `src/ontology/components/disjoint_union_over.ofn` to `src/ontology/components/disjoint_union_over.owl`, both for consistency with other components in OWL functional syntax that use a `.owl` extension and for alignment with the ontology IRI within that file;
* import that file into the `-edit` file, so that its axioms are explicitly visible to editors (e.g. from Protégé) instead of only appearing in the `uberon.owl` release product;
* declare that file as a proper ODK component, so that it is merged in all release products instead of being explicitly merged only in `uberon.owl`.

closes #3180


matentzn (approved):
Looks great!

anitacaron (approved):

PR #3195, opened 2024-01-26
Revert "Remove the NIFSTD bridge."

tgbugs:
This reverts commit ebdc7b157765c359769b9bfda645feb39460ca79.


tgbugs:
This file needs to be maintained at the very least because it is in the import chain for many older versions of external ontologies and removing it breaks the ability to import them in protege. The axioms are also useful for anyone wanting to understand the historical mappings for terms the originated in NIFSTD and are now in Uberon.

gouttegd (changes requested):
> This file needs to be maintained at the very least because it is in the import chain for many older versions of external ontologies and removing it breaks the ability to import them in protege.

The proper way to do that is by adding a redirection to the PURL system so that `http://purl.obolibrary.org/obo/uberon/bridge/uberon-bridge-to-nifstd.{obo,owl}` still resolves to the last version of that file, _not_ by cluttering the Uberon repository.

The axioms are also useful for anyone wanting to understand the historical mappings for terms the originated in NIFSTD and are now in Uberon.

tgbugs:
Adding the redirect is the better solution. Thanks!

PR #3197, opened 2024-02-05
Remove `part_of o develops_from -> develops_from` injection

### LINKED ISSUE #2723: Remove property chain injection

anitacaron:
Fixes #2723

### DIFF
--- a/src/ontology/components/develops-from-chains.owl
+++ b/src/ontology/components/develops-from-chains.owl
@@ -8,6 +8,5 @@ Prefix(rdfs:=<http://www.w3.org/2000/01/rdf-schema#>)
 
 Ontology(<http://purl.obolibrary.org/obo/uberon/components/develops-from-chains.owl>
 
-SubObjectPropertyOf(ObjectPropertyChain(<http://purl.obolibrary.org/obo/BFO_0000050> <http://purl.obolibrary.org/obo/RO_0002202>) <http://purl.obolibrary.org/obo/RO_0002202>)
 SubObjectPropertyOf(ObjectPropertyChain(<http://purl.obolibrary.org/obo/RO_0002202> <http://purl.obolibrary.org/obo/BFO_0000050>) <http://purl.obolibrary.org/obo/RO_0002202>)
 )


anitacaron:
#gogoeditdiff

gouttegd:
The “gogoeditdiff“ action failed to post the diff because it is too large (comments are apparently limited to 64 KB). Here’s the diff: [simple-report.md](https://github.com/obophenotype/uberon/files/14166260/simple-report.md)

The few changes I’ve looked at seem exactly the kind of changes we’d expect upon removal of the chain.

gouttegd (approved):

cmungall:
### upper jaw region `http://purl.obolibrary.org/obo/UBERON_0001709`
#### Removed
- [upper jaw region](http://purl.obolibrary.org/obo/UBERON_0001709) SubClassOf [structure with developmental contribution from neural crest](http://purl.obolibrary.org/obo/UBERON_0010314)

balhoff (changes requested):
Discussion with @cmungall @anitacaron @matentzn:
- add property chain `'has part' o 'has developmental contribution from' -> 'has developmental contribution from'` (edit existing which I think uses 'develops from')
- remove "derived" probe classes (e.g. "ectoderm-derived structure", and other germ layers)

anitacaron:
> add property chain 'has part' o 'has developmental contribution from' -> 'has developmental contribution from' (edit existing which I think uses 'develops from')

[This property chain is in RO](https://www.ebi.ac.uk/ols4/ontologies/ro/properties/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FRO_0002254), so the change needs to be done there. I'll create an issue there.

gouttegd:
@anitacaron I presume you intend to take care of the second action item (“remove "derived" probe classes (e.g. "ectoderm-derived structure", and other germ layers)”) in a separate PR?

If so, looks good to me here, given that the other item needs to happen elsewhere.

gouttegd (approved):
It seems a bit weird that we now have an entire component for just _one_ axiom, but oh well, that’s not the weirdest thing in Uberon!

dosumis:
Discussing this on a call.  We are unsure why this property chain is being removed (apart from that it ultimately should come from RO.  It is doing useful work for inference that we likely can't effectively manage by hand.   ALso note - we seem to have effectively the same axiomatisation via the subproperty of develops_from: develops_from_part_of https://www.ebi.ac.uk/ols4/ontologies/clo/properties/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FRO_0002225 - shouldn't this suffice to give the inference we need, and if it currently does not, why not?

(Note - the other property chain 'develops_from o part_of --> develops_from should be removed - has_developmental_contribution_from is sufficient).

gouttegd:
In the last RO call it was decided to test the various combinations of actions regarding the injected chains (remove both, remove one, remove the other) and have a look at the corresponding diffs to better understand the consequences of each chain and of their removal.

PR #3201, opened 2024-02-07
Remove basic.owl and metazoan-view.owl

### LINKED ISSUE #3013: `basic.owl` is exactly the same product as `uberon-basic.owl`
The custom `basic.owl` artifact is nothing more than the `uberon-basic.owl` artifact with just a different ontology IRI:

```Makefile
# ----------------------------------------
# STEP 5: Create basic subset
# ----------------------------------------

basic.owl:  uberon-basic.owl
	$(ROBOT) merge -i $< annotate --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $@
```

So we have a call to ROBOT that loads the ~48MB of `uberon-basic.owl` just to write them again immediately in a different file whose only difference is that its ontology ID is now `http://purl.obolibrary.org/obo/uberon/basic.owl` instead of `http://purl.obolibrary.org/obo/uberon/uberon-basic.owl`… Is is worth it?

I’d suggest removing this step so that we only create and publish the `uberon-basic.owl` artifact and, just in case there are folks out there who are currently relying on `basic.owl` (maybe because they don’t know it’s the same thing), we add a new redirect rule to Uberon’s PURL configuration so that `http://purl.obolibrary.org/obo/uberon/basic.owl` points to `http://purl.obolibrary.org/obo/uberon/uberon-basic.owl`.

Likewise for `basic.json` which is the same thing as `uberon-basic.json`.

gouttegd:
This PR removes two release products that are completely redundant since they are identical to another release product:

* `basic.owl`: exactly identical to `uberon-basic.owl` except that is has a different ontology IRI;
* `metazoan-view.owl`: exactly identical to `basic.owl` (and therefore to `uberon-basic.owl`).

There’s no need to waste time building those products and waste space hosting them. If some some pipelines out there are dependent on those files, they should be updated to use `uberon-basic.owl` instead. And if these are old pipelines that for some reason can’t be updated, then we should just have one more redirection in Uberon’s PURL configuration.

[Download statistics](https://tooomm.github.io/github-release-stats/?username=obophenotype&repository=uberon) for the Uberon releases suggest the `metazoan-view.owl` file at least is never downloaded by anyone.

closes #3013

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -62,7 +62,7 @@ PATTERN_RELEASE_FILES=      $(PATTERNDIR)/definitions.owl $(PATTERNDIR)/pattern.
 
 FORMATS = $(sort  owl obo json owl)
 FORMATS_INCL_TSV = $(sort $(FORMATS) tsv)
-RELEASE_ARTEFACTS = $(sort $(ONT)-base $(ONT)-full $(ONT)-simple $(ONT)-basic basic collected-metazoan composite-metazoan composite-metazoan-basic composite-vertebrate composite-vertebrate-basic common-anatomy )
+RELEASE_ARTEFACTS = $(sort $(ONT)-base $(ONT)-full $(ONT)-simple $(ONT)-basic collected-metazoan composite-metazoan composite-metazoan-basic composite-vertebrate composite-vertebrate-basic common-anatomy )
 
 # ----------------------------------------
 # Top-level targets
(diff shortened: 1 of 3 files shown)


matentzn (approved):
Good move. 

I made the relevant change to the PURL system: https://github.com/OBOFoundry/purl.obolibrary.org/pull/977

PR #3207, opened 2024-02-08
Remove `consider` tags pointing to Wikipedia pages

### LINKED ISSUE #3206: `consider` tags should not point to a Wikipedia page
We have a handful of cases where an obsoleted term is tagged with a `consider` tag pointing to a Wikipedia page:

```
[Term]
id: UBERON:0000449
name: obsolete decidual cell
consider: Wikipedia:Decidual_cells

[Term]
id: UBERON:0002986
name: obsolete neurofilament
consider: Wikipedia:Neurofilament

[Term]
id: UBERON:0003000
name: obsolete intermediate filament
consider: Wikipedia:Intermediate_filament

[Term]
id: UBERON:0003003
name: obsolete microfilament
consider: Wikipedia:Microfilament

[Term]
id: UBERON:0005894
name: obsolete locomotive weight-bearing appendage
consider: Wikipedia:Leg

[Term]
id: UBERON:0005896
name: obsolete jointed/prehensile appendage
consider: Wikipedia:Limb_(anatomy)

[Term]
id: UBERON:0013466
name: obsolete zygomasseteric system
consider: Wikipedia:Zygomasseteric_system

[Term]
id: UBERON:0015118
name: obsolete neocortical column
consider: Wikipedia:Cortical_column

[Term]
id: UBERON:0035560
name: obsolete fascia dentata
consider: Wikipedia:Fascia_dentata
```

The purpose of the `consider` tag is to “give a term which may be an appropriate substitute for an obsolete term, but needs to be looked at carefully by a human expert before the replacement is done” ([OBO Flat File Format Guide v1.4](http://owlcollab.github.io/oboformat/doc/GO.format.obo-1_4.html)). A Wikipedia page cannot be an “appropriate substitute“ for an ontological term – pages of Wikipedia are not intended to be semantic identifiers (there’s Wikidata for that).

Those tags should be removed.

In addition, this particular tag:

```
consider: Wikipedia:Limb_(anatomy)
```

will cause technical problems with the next version of the OWL API, because of the `(` and `)` characters.

gouttegd:
Wikipedia pages are not suitable substitutes for obsoleted terms. They are not and should not be considered to be identifiers. All the terms carrying those tags already have other `consider` tags pointing to other ontological terms, so those Wikipedia-pointing tags don’t bring anything useful.

closes #3206


src/ontology/uberon-edit.obo
```diff
@@ -64352,7 +64349,6 @@ consider: MESH:A11.284.195.190.750.510
 consider: NCIT:C13255
 consider: NIF_Subcellular:sao2006047981
 consider: UMLS:C0025979 {source="ncithesaurus:Actin_Filament"}
-consider: Wikipedia:Microfilament
```
anitacaron, line 64355: Could these Wikipedia references be in another annotation, like xref, for terms without a definition?

gouttegd: Yes, I suppose they could be xrefs or seeAlso. Is that really necessary, though? We’re talking about _obsolete_ terms after all. The only reason someone is ever going to look at them is they have been using them in their own ontologies or databases and they need to know how to replace them – and for that they will be looking at `consider` tags, not xrefs or anything else.

anitacaron: We never know. If we can keep the information, it's always better.

gouttegd: Fine, if it’s make you happy.

I believe that keeping information for the sake of keeping information utterly ridiculous, though. There’s a difference between “keeping information” and keeping all the crap that we don’t dare clean up. The focus should be on providing useful information, which sometimes means _removing_ stuff.

aleixpuigb (approved):

PR #3222, opened 2024-02-19
Fixes #3221 import RO vessel relationships and refresh imports

### LINKED ISSUE #3221: Import and use relations from RO (blood vessels)
The following terms have been created in RO for Uberon and need to be imported:
- [vessel drains blood from](http://purl.obolibrary.org/obo/RO_0020102)
- [vessel supplies blood to](http://purl.obolibrary.org/obo/RO_0020101)
- [directly supplies and drains](http://purl.obolibrary.org/obo/RO_0020103)

aleixpuigb:
Fixes #3221 to import RO terms for blood vessels and refresh imports


anitacaron:
The properties will be properly imported when they are used in an axiom. So you can use this PR to also add a term. Don't forget to update `imports/merged_import.owl` again.

anitacaron (approved):
Never mind. The `imports/merged_import.owl` was updated now with the RO relationships.

Issue #3223, opened 2024-02-19
The ontobot workflow needs a dedicated user account


gouttegd:
(Note: this issue also applies to other obophenotype repositories where the ontobot workflow is enabled, such as CL.)

We have in this repository a GitHub Action workflow that can be used to automatically create pull requests for simple changes described in a ticket with the KGCL syntax.

Currently, pull requests created with that workflow do not trigger the CI test suite automatically. This is because the workflow runs with the default authentication token `GITHUB_TOKEN`, which explicitly does not allow a workflow to trigger the execution of other workflows.

We need to create a non-default authentication token specifically for the purpose of running the ontobot workflow. While any person with admin rights on the repo can do that, it would be better if this was done from a distinct account that is not actually used by any editor. Pull requests created by the ontobot workflow will appear to have been created by whichever account created the authentication token – that account will therefore be unable to approve or request changes on the automatically created pull requests.

CC @anitacaron : This is related to [your question](https://github.com/obophenotype/uberon/pull/3216#pullrequestreview-1882188197) about why the CI test suite was not triggered on #3216.

balhoff:
@gouttegd I have done this for GO using this account: @ontobot. 

Example: https://github.com/geneontology/go-ontology/pull/27037

Do you want to use this account somehow? How do we manage that? Or else will you just create another bot account?

gouttegd:
@balhoff I had forgotten the existence of that account. Reusing it for Uberon/CL would be nice, rather than creating a new one. Would that be possible? Who controls that account?

balhoff:
It's my account. Just let me know the best way to create a token and give it to you.

gouttegd:
I suppose you have admin access to the Uberon and CL repos? Then if I understand correctly I don’t even need to see the token, you should be able to add it directly as a “repository secret” here: https://github.com/obophenotype/uberon/settings/secrets/actions

Then on our side we would need to grant “collaborator access“ to the `ontobot` account.

@hrshdhgd Is there anything else that we should do that I am not aware of?

hrshdhgd:
Once the token is saved as `GH_TOKEN` with adequate permissions, everything should work as expected flawlessly.

balhoff:
[...]

PR #3225, opened 2024-02-20
Use a dedicated token for the ontobot workflow.

gouttegd:
This commit updates the ontobot workflow (`auto-pr-ontobot`) to make it use a dedicated authentication token (`ONTOBOT_TOKEN`), rather than the default `GITHUB_TOKEN`.

This should ensures that PRs created by the bot can trigger other workflows, notably the CI workflow.

related to #3223

### DIFF
--- a/.github/workflows/auto-pr-ontobot.yml
+++ b/.github/workflows/auto-pr-ontobot.yml
@@ -47,7 +47,7 @@ jobs:
         id: gh-script-issue
         uses: actions/github-script@v6
         with:
-          # github-token: ${{ secrets.GH_TOKEN }}
+          github-token: ${{ secrets.ONTOBOT_TOKEN }}
           script: |
             const issue_number = context.issue.number
             const repo = context.repo.owner+"/"+context.repo.repo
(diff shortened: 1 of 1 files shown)


.github/workflows/auto-pr-ontobot.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         id: gh-script-repo
         uses: actions/github-script@v6
         with:
-          # github-token: ${{ secrets.GH_TOKEN }}
+          github-token: ${{ secrets.ONTOBOT_TOKEN }}
```
hrshdhgd, line 60: This may not be needed either.

hrshdhgd (approved):
The 2 lines that I've written comments on were commented before because I was testing if those authentication lines were needed or no. I think we can get away without using them. The rest looks good.

gouttegd:
Yes, I suspected that the lines you flagged might not be necessary – the standard permissions granted by the default `GITHUB_TOKEN` should be enough for what those steps are doing. Still, I like the consistency of using the same token throughout the workflow.

(There’s a release ongoing, so I’ll wait for the release to be over before merging this.)

hrshdhgd:
Sounds good!

anitacaron (approved):

PR #3218, opened 2024-02-16
Obsolete and merge a handful of redundant terms

gouttegd:
This PR obsoletes a few terms that are redundant with other terms:

* 'optic eminence surface ectoderm' (https://github.com/obophenotype/uberon/issues/3156),
* 'vitelline venous plexus' (https://github.com/obophenotype/uberon/issues/3157),
* 'thymus subunit' (https://github.com/obophenotype/uberon/issues/3158),
* 'gland of anal sac' (https://github.com/obophenotype/uberon/issues/3159),
* 'telencephalic tracts' (https://github.com/obophenotype/uberon/issues/3160),
* 'common crus' (https://github.com/obophenotype/uberon/issues/3161).

It supersedes #3196, which attempted to do the same change but contains a commit that should not have been done.

PR #3226, opened 2024-02-20
2024-02-20 Release

anitacaron:


gouttegd (approved):
For information, about the changes in bridge files:

* The huge diffs regarding the `cl-bridge-to-fma.owl` and `uberon-bridge-to-fma.owl` are the normal consequence of #3127.
* The other meaningful changes are the normal consequence of #3218.

In short, nothing unexpected. All quiet on the bridge front.

aleixpuigb (approved):
There are many new inferred axioms that are correct but I'm not sure from which PR are they from. Since they are correct, I accept the PR.
Examples:
image

PR #3229, opened 2024-02-20
Testing the ontobot workflow

### LINKED ISSUE #3228: Testing the ontobot workflow
** /!\ This is not a real issue. /!\ **

It is only intended to test the ontobot workflow that automatically turns issues into pull requests. All editors can safely ignore it. I am deliberately instructing the bot to make a change that is almost outrageously wrong – obsoleting one the core terms of the ontology – so that no one can inadvertently think of merging the PR that will be created

Hey ontobot, apply:

* obsolete UBERON:0001062

ontobot:
The following commands were executed: </br> * obsolete UBERON:0001062</br></br>Fixes #3228


gouttegd:
Cancelling the CI workflow since all we wanted to know was whether the workflow would be triggered (which was the case). No need to waste CPU time running the entire test suite.

PR #3230, opened 2024-02-20
Remove custom OBO generation rule.

### LINKED ISSUE #3014: Revise MAKEOBO custom rule
The ODK standard method to create a `.obo` artifact from a `.owl` one is this:

```sh
robot convert -i myfile.owl --check false -f obo -o myfile.obo.tmp && grep -v ^owl-axioms: myfile.obo.tmp > myfile.obo && rm tmp.obo
```

In Uberon, we use this custom rule instead:

```Makefile
MAKEOBO=  $(OWLTOOLS) $< --add-obo-shorthand-to-properties  -o -f obo --no-check $@.tmp1 && grep -v ^property_value: $@.tmp1 | perl -npe 's@relationship: dc-@property_value: dc-@' | grep -v ^owl-axioms: > $@.tmp && mv $@.tmp  $@
```

So in addition to removing the `owl-axioms:` tag, this rule:
* adds shorthands to object properties that don’t already have one (an “OBO shorthand” is basically what allows to use, e.g, `part_of` instead of `BFO:0000050` in a class expression);
* removes all `property_value` tags;
* replaces DC annotations incorrectly represented as relationships by proper annotations (e.g. `relationship: dc-creator ...` -> `property_value: dc-creator ...`.

I question whether this custom rule is useful and if so, whether it should be ported to the ODK instead. I am particularly unconvinced by the usefulness of the additions of OBO shorthands.

gouttegd:
In a few places, when generating OBO files from OWL files, we use a custom OWLTOOLS-based command instead of relying of the default OWL-to-OBO rule set forth in the ODK.

There should be no need for such custom rules, so we remove them here.

closes #3014

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -31,15 +31,6 @@ all: uberon-qc
 
 
 # ----------------------------------------
-# COMMANDS
-# ----------------------------------------
-
-# FIXME: Is a custom OBO generation rule needed?
-# https://github.com/obophenotype/uberon/issues/3014
-MAKEOBO=  $(OWLTOOLS) $< --add-obo-shorthand-to-properties  -o -f obo --no-check $@.tmp1 && grep -v ^property_value: $@.tmp1 | perl -npe 's@relationship: dc-@property_value: dc-@' | grep -v ^owl-axioms: > $@.tmp && mv $@.tmp  $@
-
-
-# ----------------------------------------
 # XML CATALOG
 # ----------------------------------------
 
@@ -165,12 +156,6 @@ uberon.owl: $(OWLSRC) $(BRIDGEDIR)/uberon-bridge-to-bfo.owl $(DEVELOPS_FROM_CHAI
 	                       --equivalent-classes-allowed asserted-only \
 	         unmerge -i $(DEVELOPS_FROM_CHAIN) \
 	         annotate -O $(URIBASE)/$@ -V $(RELEASE)/$@ -o $@
-
-# Step 2.1: Generating other formats.
-# also do OE check here
-uberon.obo: uberon.owl
-	$(MAKEOBO)
-.PRECIOUS: uberon.obo
 
 uberon.json.gz: uberon.json
 	gzip -c $< > $@.tmp && mv $@.tmp $@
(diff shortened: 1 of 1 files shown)


src/ontology/uberon.Makefile
```diff
@@ -166,12 +157,6 @@ uberon.owl: $(OWLSRC) $(BRIDGEDIR)/uberon-bridge-to-bfo.owl $(DEVELOPS_FROM_CHAI
 	         unmerge -i $(DEVELOPS_FROM_CHAIN) \
 	         annotate -O $(URIBASE)/$@ -V $(RELEASE)/$@ -o $@
 
-# Step 2.1: Generating other formats.
-# also do OE check here
-uberon.obo: uberon.owl
-	$(MAKEOBO)
-.PRECIOUS: uberon.obo
```
matentzn, line 173: Have you tried this? This is basically switching the uberon.obo release (one of teh most important release files) to ODK default right? Does it still look ok in the diff?

matentzn (commented):
One comment, else YES

src/ontology/uberon.Makefile
gouttegd: Yes, I tried. The diff is (expectedly) quite large, because one of the things the custom `MAKEOBO` rule did was to remove _all_ the `property_value:` tags. With the standard ODK rule, all those tags are still there.

Whether this is a problem or not, I don’t know. But if it is not a problem for all the ontologies using the standard ODK rule, why would that be a problem for Uberon?

gouttegd: I suspect this could be a problem only if a downstream user is parsing the `uberon.obo` file with some weird or application-specific script (maybe a script that doesn’t even know about the existence of the `property_value` tag?). But if so, that’s the kind of things we may very well never know about.

matentzn: Ok, I can get behind that! Thanks for checking.

gouttegd: If someone does need a OBO version of Uberon that is stripped of those tags, I’d suggest to do the same thing we do in FlyBase. The OBO parser we use to load our ontologies into the database does have some special constraints, so we have a special release artefact to produce an OBO file that meet those constraints.

We could do the same here: leave the `uberon.obo` file be the file generated by the standard ODK rule, and have an extra `uberon-uncomplicated-obo-version.obo` where we strip everything that a (too) simple parser might not like. But we should only do that if, and only if, someone asks for it.

matentzn: ok!

PR #3233, opened 2024-02-24
Remove intermediate file `external-disjoints.owl`

gouttegd:
This PR is a minor cleanup of the custom Makefile.

It removes a step that converts the `$(COMPONENTSDIR)/external-disjoints.obo` file into an intermediate `$(TMPDIR)/external-disjoints.owl`.

The `external-disjoints.owl` file is only ever used with ROBOT, and ROBOT is perfectly able to consume directly an OBO file, there’s no need for an explicit conversion to OWL.

It also removes the step that copied both the OBO and OWL versions of that file to the top-level `bridge` directory, which does not make any sense:

* that directory only contains those two files;
* in that directory, these files are “unreachable” by whoever would want to download them from a PURL, because `http://purl.obolibrary.org/obo/uberon/bridge` redirects to the `src/ontology/bridge` directory, _not_ to the top-level `bridge` directory.

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -676,12 +676,6 @@ $(REPORTDIR)/%-dv.txt: %.owl
 # Overall "all-bridge" check
 # ----------------------------------------
 
-# The disjointness axioms are maintained in a OBO file, but the checks
-# below use an OWL version.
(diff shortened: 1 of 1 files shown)


matentzn (approved):
LGTM

PR #3126, opened 2023-11-23
3056 Fix duplicate mappings

gouttegd:
This is a first pass at fixing the duplicated cross-references highlighted in #3056.

PR #3244, opened 2024-03-22
Refresh external resources for release

anitacaron:


gouttegd (approved):
I don’t see anything fishy.

All changes to FlyBase-related stuff are fully expected.

Most of the other changes are the normal and expected consequence of #3126: duplicate mappings were fixed, so mappings that were previously ignored (because of the duplicates) can now find their way to the bridge files.

PR #3248, opened 2024-04-08
Use ODK-managed ROBOT plugins.

gouttegd:
This PR replaces the custom code that was used to download and install ROBOT plugins by ODK-managed code.

The SSSOM plugin for ROBOT is bundled with the ODK, so we no longer need to manually download it.

As for the Uberon-specific plugin, we still need to download it but we can rely on the ODK to do that for us--all we need is to declare the need to use that plugin.

### DIFF
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -10,7 +10,7 @@
 # More information: https://github.com/INCATools/ontology-development-kit/
 
 # Fingerprint of the configuration file when this Makefile was last generated
-CONFIG_HASH=                c8fa86ab8f984cb6c9ea20468b4ae4ff2e446501d9c361b8a3993841a38177eb
+CONFIG_HASH=                67fd0d77c743b5ebe8d373f9296b22a0c6c3ef8d825a11d5639915e37fbe7eb1
 
 
 # ----------------------------------------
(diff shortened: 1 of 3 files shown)


matentzn (approved):
Looks great!

src/ontology/Makefile
```diff
@@ -147,6 +147,9 @@ $(ROBOT_PLUGINS_DIRECTORY)/%.jar:
 
 # Specific rules for supplementary plugins defined in configuration
 
+$(ROBOT_PLUGINS_DIRECTORY)/uberon.jar:
+	curl -L -o $@ https://github.com/gouttegd/uberon-robot-plugin/releases/download/uberon-robot-plugin-0.2.0/uberon.jar
```
gouttegd, line 151: By the way this makes me think that I should probably move that repository `uberon-robot-plugin` to the OBOphenotype organisation, since the plugin is completely Uberon-specific. Would you agree?

matentzn: I have no objections either way! Thank you!

gouttegd: OK, not a priority, but I’ll do that at some point. It won’t change anything in practical terms but it will be “cleaner“.

PR #3253, opened 2024-04-15
Addresses #3252 NTR trophoblast island

aleixpuigb:
Fixes #3252 NTR trophoblast island
Names in Orcid are dissappearing, maybe a problem importing Orcidio

PR #3257, opened 2024-04-17
Fix ORCID individuals import

anitacaron:
Related to #3253 (ORCID names are being removed in diff)

I chose to have a separate import because the merged_import excludes individuals, and I wanted to enable this only for Orcidio. I created a custom rule for `imports/orcidio_import.owl` and imported it in `uberon-edit.obo`.

What do you think?

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -438,13 +438,20 @@ all_local_imports:
 endif
 
 # ----------------------------------------
-# EXTRACT ORCIDs
+# EXTRACT ORCIDs INDIVIDUALS
 # ----------------------------------------
 
 # Extract ORCID from dcterms:contributor annotations
-$(IMPORTDIR)/orcidio_terms_combined.txt: $(SRCMERGED)
+$(IMPORTDIR)/orcidio_terms_combined.txt: $(SRC)
 	$(ROBOT) query -f csv -i $< --query ../sparql/orcids.sparql $@.tmp &&\
 	cat $@.tmp | sort | uniq >  $@
+
+# Create orcidio_import.owl to extract individuals excluded in the merged_import.owl
+$(IMPORTDIR)/orcidio_import.owl: $(MIRRORDIR)/orcidio.owl $(IMPORTDIR)/orcidio_terms_combined.txt
+	if [ $(IMP) = true ]; then $(ROBOT) query -i $< --update ../sparql/preprocess-module.ru \
+    extract -T $(IMPORTDIR)/orcidio_terms_combined.txt --copy-ontology-annotations true --force true --individuals include --method STAR \
+	query --update ../sparql/postprocess-module.ru \
+	$(ANNOTATE_CONVERT_FILE); fi
 
 
 # ----------------------------------------
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -437,6 +437,15 @@ all_local_imports:
 
 endif
 
+# ----------------------------------------
+# EXTRACT ORCIDs
+# ----------------------------------------
+
+# Extract ORCID from dcterms:contributor annotations
+$(IMPORTDIR)/orcidio_terms_combined.txt: $(SRCMERGED)
+	$(ROBOT) query -f csv -i $< --query ../sparql/orcids.sparql $@.tmp &&\
+	cat $@.tmp | sort | uniq >  $@
+
 
 # ----------------------------------------
 # UNRELEASED PRODUCTS
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -448,9 +448,7 @@ $(IMPORTDIR)/orcidio_terms_combined.txt: $(SRC)
 
 # Create orcidio_import.owl to extract individuals excluded in the merged_import.owl
 $(IMPORTDIR)/orcidio_import.owl: $(MIRRORDIR)/orcidio.owl $(IMPORTDIR)/orcidio_terms_combined.txt
-	if [ $(IMP) = true ]; then $(ROBOT) query -i $< --update ../sparql/preprocess-module.ru \
-    extract -T $(IMPORTDIR)/orcidio_terms_combined.txt --copy-ontology-annotations true --force true --individuals include --method STAR \
-	query --update ../sparql/postprocess-module.ru \
+	if [ $(IMP) = true ]; then $(ROBOT) filter -i $< -T $(IMPORTDIR)/orcidio_terms_combined.txt --trim false \
 	$(ANNOTATE_CONVERT_FILE); fi
 
 
--- a/src/sparql/orcids.sparql
+++ b/src/sparql/orcids.sparql
@@ -1,0 +1,7 @@
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+prefix owl: <http://www.w3.org/2002/07/owl#>
+SELECT DISTINCT ?orcid
+WHERE {
+  ?term <http://purl.org/dc/terms/contributor> ?orcid . 
+  FILTER(isIRI(?term))
+}
--- a/.gitignore
+++ b/.gitignore
@@ -219,3 +219,4 @@ src/ontology/subsets/*.obo
 src/ontology/subsets/*.owl
 src/ontology/subsets/*.tsv
 src/ontology/imports/*_import.owl
+!src/ontology/imports/orcidio_import.owl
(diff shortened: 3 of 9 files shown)


src/ontology/uberon.Makefile
```diff
@@ -437,6 +437,22 @@ all_local_imports:
...
+
+# Extract ORCID from dcterms:contributor annotations
+$(IMPORTDIR)/orcidio_terms_combined.txt: $(SRC)
+	$(ROBOT) query -f csv -i $< --query ../sparql/orcids.sparql $@.tmp &&\
+	cat $@.tmp | sort | uniq >  $@
+
+# Create orcidio_import.owl to extract individuals excluded in the merged_import.owl
+$(IMPORTDIR)/orcidio_import.owl: $(MIRRORDIR)/orcidio.owl $(IMPORTDIR)/orcidio_terms_combined.txt
+	if [ $(IMP) = true ]; then $(ROBOT) query -i $< --update ../sparql/preprocess-module.ru \
+    extract -T $(IMPORTDIR)/orcidio_terms_combined.txt --copy-ontology-annotations true --force true --individuals include --method STAR \
```
matentzn, line 452: surprised this does not include _all_ individuals in Orcidio.. Usually, individuals are considered "global" by SLME and therefore included by default (even if they are not in the signature file).

src/sparql/orcids.sparql
```diff
@@ -0,0 +1,7 @@
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+prefix owl: <http://www.w3.org/2002/07/owl#>
+SELECT DISTINCT ?orcid
+WHERE {
+  ?term <http://purl.org/dc/terms/contributor> ?orcid . 
+  FILTER(isIRI(?term))
```
matentzn, line 6: This constraint could miss oio:hasDbXrefs on defs, FYI

matentzn (dismissed):

src/sparql/orcids.sparql
anitacaron: I might be redundant.

src/ontology/uberon.Makefile
gouttegd: Are we sure the observed behaviour is not a bug in either ROBOT or the OWL API then?

My understanding of the documentation for ROBOT’s `extract` command (“When using the SLME method of extraction, all individuals (ABox axioms) and their class types (the TBox axioms they depend on) are included by default.”) is that indeed we should get individuals – regardless of the type of module.

I’m trying to have a look at what the code is actually doing, but I find the logic of the `SyntacticLocalityModuleExtractor` quite hard to follow…

gouttegd: In fact I am not even sure whether ROBOT `extract` is _supposed_ to be usable for extracting individuals.

According to the [folks at Manchester University](http://owl.cs.manchester.ac.uk/research/modularity/), “a locality-based module M is a subset of the axioms in an ontology O, and is extracted from O for a set S of terms (**class or property names**)” (emphasis mine). So did they just forget to mention individuals here? Or is SLME really not designed to be able to extract a module from a set of **individual** names?

gouttegd: Also, is it expected that the extracted module does not contain the class assertions? None of the individuals in the `orcidio_import.owl` module have a type. In the original `orcidio.owl` ontology the individuals are asserted to be instances of `NCBITaxon:9606`.

OK, it may not be a great loss of information (we don’t expect the contributors to be of another species than NCBITaxon:9606, do we? :D ), but it means we end up with all the individuals “dangling”, not attached to any class. It does not look great.

matentzn: > Or is SLME really not designed to be able to extract a module from a set of individual names?

Exactly, from the perspective of SLME, individuals are part of any "sub-domain" (module), regardless of whether they "touch" the list of terms provided. 

IMO the correct module type here is simply "filter", maybe "minimal". Both are build into ODK (you can steal the recipe from there), and boths should have the desired effect.

Note that I think right now the import, since its not officially registered with ODK, need to be manually refreshed, i.e. none of the usual processes will refresh it..

gouttegd: > Note that I think right now the import, since its not officially registered with ODK, need to be manually refreshed, i.e. none of the usual processes will refresh it..

Er, `orcidio` _is_ declared as a proper import:

```yaml
import_group:
  products:
    - id: orcidio
      mirror_from: https://w3id.org/orcidio/orcidio.owl
      base_iris:
        - https://orcid.org
```

In my tests the local `orcidio.owl` mirror was correctly refreshed as needed.

gouttegd: > Exactly, from the perspective of SLME, individuals are part of any "sub-domain" (module), regardless of whether they "touch" the list of terms provided.

If so, then how comes that we get the individuals or not depending on the type of SLME module we use?

Because I confirm Anita’s observations: with `--method BOT`, the module contains _all_ the individuals (which is the normal behaviour, according to both ROBOT’s documentation and what you say); but with `--method STAR`, I only get the individuals listed in the seed. From what you say, this is _not_ normal.

I have not seen any hint anywhere (in ROBOT’s documentation or in the references for SLME) that the STAR method should behave differently than the others regarding the inclusion of individuals. So, I suspect in bug in ROBOT’s (or the OWL API’s) implementation of STAR.

gouttegd: My (rudimentary) understanding of the STAR method, as described [here](https://www.informatik.uni-bremen.de/tdki/users/schneidt/publ/ssz09_insep_report.pdf), is that it consists in chaining a TOP extraction and a BOTTOM extraction, and doing so repeatedly until it no longer makes any difference on the size of the module (meaning the module cannot be reduced any further). I don’t see how this would explain the observed effects on individuals, if it is true that individuals are always considered to be part of any module anyway.

anitacaron: Actually, first I tried the same recipe available in RO. It only gets the ORCIDs in the seed. But it didn't work here.

https://github.com/oborel/obo-relations/blob/f0994da3b621d7284848abb02803d5f7f27a3ac0/src/ontology/Makefile#L422-L428

anitacaron: If I try the same recipe here, it only gets the individual declarations.

matentzn: > Er, orcidio is declared as a proper import:

Oops, ok. Then it will sort of also appear as part of `merged_import.owl` no?

Anyways I didnt mean to distract here, maybe STAR has a bug, maybe not; I think filter is more appropriate.

anitacaron:
Can I please merge this PR and move the discussion to another place?

src/ontology/uberon.Makefile
anitacaron: > Oops, ok. Then it will sort of also appear as part of merged_import.owl no?

No, as I said before, the individuals are excluded in the merged_import.
https://github.com/obophenotype/uberon/blob/911720b9aafb05afdb436ea6e31b83a443d9de87/src/ontology/uberon-odk.yaml#L33

matentzn: Ahhh sorry. Too much comms at the moment. Thanks for the patience and sorry for the duh.

anitacaron:
I can change it to BOT and then get all individuals, which is the expected behaviour.

gouttegd:
But then _all_ the individuals defined in `orcidio` would end up in the `orcidio_import.owl` module, including those who have nothing to do with Uberon.

gouttegd:
This works as I believe it should:

```
robot filter -i mirror/orcidio.owl -T imports/orcidio_terms_combined.txt --trim false  -o imports/orcidio_import.owl
```

src/ontology/uberon.Makefile
```diff
@@ -437,6 +437,20 @@ all_local_imports:
...
+# ----------------------------------------
+
+# Extract ORCID from dcterms:contributor annotations
+$(IMPORTDIR)/orcidio_terms_combined.txt: $(SRC)
+	$(ROBOT) query -f csv -i $< --query ../sparql/orcids.sparql $@.tmp &&\
+	cat $@.tmp | sort | uniq >  $@
+
+# Create orcidio_import.owl to extract individuals excluded in the merged_import.owl
+$(IMPORTDIR)/orcidio_import.owl: $(MIRRORDIR)/orcidio.owl $(IMPORTDIR)/orcidio_terms_combined.txt
+	if [ $(IMP) = true ]; then $(ROBOT) filter -i $< -T $(IMPORTDIR)/orcidio_terms_combined.txt --trim false \
```
gouttegd, line 451: In the future, I would like us to adopt more broadly the practice of using Make conditionals instead of shell conditionals; Make conditionals are more efficient and easier to read.

gouttegd (approved):
Looks good.

You don’t need to do anything about my remark regarding the use of Make conditionals. It’s something we can do later, maybe by doing a single overhauling pass over the custom Makefile to replace _all_ shell conditionals in one step.

PR #3263, opened 2024-04-25
Fix downloading of FBbt mapping set

gouttegd:
This PR amends the code that refreshes the FBbt mapping set to:

* use the new canonical location of the set (`http://purl.obolibrary.org/obo/fbbt.sssom.tsv`, valid since today’s FBbt release, instead of the old `http://purl.obolibrary.org/obo/fbbt-mappings.sssom.tsv`);
* ensure the remote file is always downloaded (unless `IMP=false`) even if the target file already exists (otherwise the rule would always be ignored and the set would never be refreshed – this is a bug inadvertently introduced in #3249).

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -1259,7 +1259,8 @@ ifeq ($(strip $(IMP)),true)
 
 # FBbt mapping set. Nominally a simple mirror, but we need a custom rule
 # because the default, ODK-generated rule would ignore the MIR variable.
-$(MAPPINGDIR)/fbbt.sssom.tsv:
+.PHONY: fbbt-mappings
+$(MAPPINGDIR)/fbbt.sssom.tsv: fbbt-mappings
 	wget "http://purl.obolibrary.org/obo/fbbt/fbbt.sssom.tsv" -O $@
 
 # CL mapping set (extracted from CL cross-references). We need to first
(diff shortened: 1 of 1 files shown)


matentzn (dismissed):

src/ontology/uberon.Makefile
```diff
@@ -1259,8 +1259,9 @@ ifeq ($(strip $(IMP)),true)
 
 # FBbt mapping set. Nominally a simple mirror, but we need a custom rule
 # because the default, ODK-generated rule would ignore the MIR variable.
-$(MAPPINGDIR)/fbbt.sssom.tsv:
-	wget "http://purl.obolibrary.org/obo/fbbt/fbbt-mappings.sssom.tsv" -O $@
+.PHONY: fbbt-mappings
```
matentzn, line 1262: Usually I recomment to use .FORCE which is a widely used hack for this and a bit nicer for documentation purposes (no need to check what fbbt-mappings does) 

https://seanbone.ch/makefile-hacks/#Forcing_a_target_to_rebuild

gouttegd: Adopted, thanks!

matentzn (commented):
Unfortunately this is not a built in but a convention - the fact that .FORCE is .PHONY has to be declared somewhere in the makefile, usually up top. Maybe it is already declared in the main ODK makefile though (I am on phone)

matentzn (approved):

PR #3266, opened 2024-04-29
Use CL-provided mapping set.

gouttegd:
This PR implements the Uberon side of the idea proposed in CL: https://github.com/obophenotype/cell-ontology/issues/2358.

It updates the mappings/bridges pipeline to use the mapping set that will now be provided directly by CL, so that we no longer have to deal with the CL mappings in Uberon.

We no longer need a ZFA mapping set, since that set only contained the CL-to-ZFA mappings and those mappings are now provided by CL itself. The Uberon-to-ZFA mappings are in the uberon-local set, since Uberon xrefs are the source of truth for those mappings.

When we mirror the FBbt set, we no longer need to keep the FBbt-to-CL mappings, since again they are in the CL-provided set.

Lastly, the Uberon "meta" set (the one that ultimately gets published) will now only contain the Uberon-related mappings (Uberon-to-anything) and will no longer bundle the CL mappings. Basically, the idea is that from now on every ontology is only responsible for providing its own mappings:

* FBbt provides the FBbt-to-anything mappings;
* CL provides the CL-to-anything mappings;
* Uberon provides the Uberon-to-anything mappings.

This does not change anything for the *bridges*. Uberon will still provide both the Uberon-to-anything bridges *and* the CL-to-anything bridges, because I don't think duplicating the logic that produces those bridges across both the Uberon and CL repositories would be worth it, especially since it is in Uberon that those bridges are actually needed (to build the collected/composite ontologies).


gouttegd:
CI workflow explicitly cancelled because the new mappings pipeline won’t work until we got a CL release, since it’s only at that point that the new CL mapping set will become available at `http://purl.obolibrary.org/obo/cl/cl.sssom.tsv`. That’s why this a draft PR for now.

@matentzn FYI.

docs/bridges.md
```diff
@@ -203,23 +203,24 @@ The pipeline may be triggered by invoking the Make target
...
 
-1. Production of a SSSOM mapping set from the Uberon and CL
-   cross-references. The Uberon `-edit` file is merged with the mirrored
-   copy of CL (which is downloaded if not already available) and
-   cross-references from both Uberon and CL are extracted and turned
-   into a SSSOM mapping set. This takes care of all the mappings with
-   the foreign ontologies for which Uberon and CL are the source of
-   truth.
+1. Production of a SSSOM mapping set from the Uberon cross-references.
+   Cross-references are extracted directly from the Uberon `-edit` file
```
matentzn, line 207: I am wondering if long term, the choice of using the edit file as the basis for extracting mapping sets intended for release will bite us in the tail. There are a small handful of pipelines I know of which process edit file mappings into something else during the release process, and it is that "something else" that you would want to see in the published mapping set.

matentzn (approved):
LGTM! Nice one.

docs/bridges.md
gouttegd: If/when needed, it would be possible to extract the cross-references from the _preprocessed_ file (`EDIT_PREPROCESSED`) instead of the `-edit` file. That would still be compatible with the scenario where the mapping sets are needed to produce a component needed to produce the final release artefacts.

And where the mapping sets are _not_ needed to produce a component needed to produce the final release artefacts, there shouldn’t be any problem in extracting the cross-references from the final release artefacts.

gouttegd: The only situation where we could paint ourselves in a corner is if we have a mapping set that, for any reason, must be extracted near the end of the pipeline (e.g. because it depends on some custom processes that happen late in the pipeline) _and_ that is also needed to derive a component that must be available early in the pipeline. That would obviously be a problem.

anitacaron (approved):

src/ontology/uberon.Makefile
```diff
@@ -1233,55 +1233,41 @@ $(MAPPINGDIR)/uberon-local.sssom.tsv: $(SRC) | all_robot_plugins
...
-# ZFA mapping set (extracted from ZFA cross-references).
-$(MAPPINGDIR)/zfa.sssom.tsv: $(IMPORTDIR)/local-zfa.owl | all_robot_plugins
-	$(ROBOT) sssom:xref-extract -i $(IMPORTDIR)/local-zfa.owl --mapping-file $@ \
-		                    -v --drop-duplicates \
-		                    --set-id "$(ONTBASE)/mappings/zfa-mappings.sssom.tsv" \
-	> $(REPORTDIR)/zfa-xrefs-extraction.txt
+# CL mapping set. We simply fetch it as it is from CL. It already
+# contains all mappings related to CL (including the FBbt and ZFA ones).
+$(MAPPINGDIR)/cl.sssom.tsv: .FORCE
+	wget "http://purl.obolibrary.org/obo/cl/cl.sssom.tsv" -O $@
```
matentzn, line 1271: Should this be under the MIR false regime?

src/ontology/uberon.Makefile
```diff
@@ -1233,55 +1233,41 @@ $(MAPPINGDIR)/uberon-local.sssom.tsv: $(SRC) | all_robot_plugins
...
-$(MAPPINGDIR)/cl.sssom.tsv: $(SRC) $(IMPORTDIR)/local-cl.owl | all_robot_plugins
-	$(ROBOT) merge -i $(SRC) -i $(IMPORTDIR)/local-cl.owl --collapse-import-closure false \
-		 remove --base-iri http://purl.obolibrary.org/obo/CL_ --axioms external \
-		 sssom:xref-extract --mapping-file $@ -v --drop-duplicates \
-		                    --set-id "$(ONTBASE)/mappings/cl-mappings.sssom.tsv" \
-		                    --prefix 'KUPO:  http://purl.obolibrary.org/obo/KUPO_'  \
-		                    --prefix 'SCTID: http://purl.obolibrary.org/obo/SCTID_' \
-		                    --prefix 'FMA:   http://purl.org/sig/ont/fma/fma' \
-	> $(REPORTDIR)/cl-xrefs-extraction.txt
+	wget "http://purl.obolibrary.org/obo/fbbt/fbbt.sssom.tsv" -O - | \
```
matentzn, line 1262: Should this be under the MIR false regime?

matentzn (approved):
Looks good! Too clarifying comments, else ready for merge!

src/ontology/uberon.Makefile
gouttegd: The entire section is already within a conditional block:

```Make
ifeq ($(strip $(IMP)),true)
    ...
endif
```

It’s dependent on `IMP` rather than `MIR` because we are not _mirroring_ anything here. Contrary to imports, for which we have a two-steps process (first mirroring, and then creating the import module proper), for foreign mappings it’s all done in a single step.

PR #3278, opened 2024-05-16
GH Action to post release diff on release PR

anitacaron:
The same as we have in [CL](https://github.com/obophenotype/cell-ontology/blob/master/.github/workflows/post-release-diff.yaml)

### DIFF
--- a/.github/workflows/post-release-diff.yml
+++ b/.github/workflows/post-release-diff.yml
@@ -1,0 +1,30 @@
+name: Post release diff
+
+on:
+  # Triggers the workflow on pull request events for the master branch
+  pull_request:
+    branches: [ master ]
+    paths:
+      - 'src/ontology/reports/uberon-base-diff.md'
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+
+  post_diff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Prepare release diff comment
+        env:
+            GITHUB_SHA: ${{ github.sha }}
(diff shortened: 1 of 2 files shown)


matentzn (dismissed):
I didnt think about this deeply, but LGTM!

.github/workflows/post-release-diff.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Prepare release diff comment
+        env:
+            GITHUB_SHA: ${{ github.sha }}
+        run: "echo \"[Here's a diff of how this release impacts uberon-base.owl](https://github.com/obophenotype/uberon/blob/${{ env.GITHUB_SHA }}/src/ontology/diffs/uberon-diff.md)\" >comment.md"
+      - name: Post comment
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: NejcZdovc/comment-pr@v1.1.1
```
gouttegd, line 26: Likewise, could you update to `v2`?

gouttegd (dismissed):
All good, but please consider updating the actions we are using (and we should also at some point do that for all our workflows).

.github/workflows/post-release-diff.yml
```diff
@@ -0,0 +1,30 @@
...
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+
+  post_diff:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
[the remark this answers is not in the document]
anitacaron: Yeah, thanks for the reminder!

.github/workflows/post-release-diff.yml
anitacaron: Yes, done.

matentzn (approved):

anitacaron:
I cancelled the QC, which was unnecessary for this PR.

PR #3298, opened 2024-06-14
NEVER MERGE: Test release Uberon (not to be merged!)

matentzn:
If you see this commit in the git history someone made a mistake, it was supposed to be for review purpose only.

This PR is the consequence of running the `make uberon` pipeline, omitting the faulty `DMBA:16271` mapping which causes an inconsisteny. All release files are updated, but there are some interesting changes:

- Lots of prefix changes
- The consider axioms are changed to PURL syntax in OWL serialisation? that seems odd

OLD:

<oboInOwl:consider>EFO:0001015</oboInOwl:consider>

New:

<oboInOwl:consider rdf:resource="http://purl.obolibrary.org/obo/EFO_0001015"/>


matentzn:
> 
@matentzn that change is from a string to an IRI, and that is what we wanted to happen for replaced_by and consider.

So for the EFO and FMA cases, we basically need to add idspace declarations into uberon-edit to make that work?

balhoff:
> So for the EFO and FMA cases, we basically need to add idspace declarations into uberon-edit to make that work?

That sounds right.

src/ontology/uberon-edit.obo
```diff
@@ -222728,7 +222728,6 @@ def: "A hypothalamic region in rodents that is homologous to the lateral tuberal
 comment: While lateral tuberal nucleus is often used interchangeably with tuberal nucleus in rodents, lateral tuberal nucleus is specific to humans and higher primates. In some cases, lateral tuberal nucleus is thought to be part of the tuberal nucleus in rodents, and could possibly be homologous to human tuberal region instead. {xref="PMID:29976824", xref="neuronames:1890", xref="PMID:1362279"}
 synonym: "lateral tuberal nucleus (sensu Rodentia)" BROAD [neuronames:1890]
 synonym: "lateral tuberal nucleus (sensu Rodentia)" EXACT [PMID:29976812]
-xref: DMBA:16271
```
gouttegd, line 222731: Just for the record: the proper way to workaround the issue caused by the problematic axiom in the DMBA bridge (which is something that ultimately must be fixed [upstream](https://github.com/obophenotype/ABA_Uberon/issues/36), the problem is _not_ in Uberon) is not to remove that cross-reference (the mapping here is correct) but to manually remove DMBA:16271 from the `src/ontology/bridge/uberon-bridge-to-dmba.owl` file, after that file has been refreshed from its original online location.

matentzn: My personal preference would be to literally freeze the bridge in the makefile by commenting it out, but I guess I will try to put the screws on the upstream issue instead. Thanks for linking it!

gouttegd: To be clear the fix must happen in [this repository](https://github.com/brain-bican/developing_mouse_brain_atlas_ontology/issues/1); I've linked the issue in the old repository only because it has a proper description of the problem, contrary to the new one that just states what the fix must be.

matentzn:
Superceded by #3300

PR #3299, opened 2024-06-14
Do not merge: Test release of Uberon with ODK 1.5.1

matentzn:
All updated files according to `make uberon`.


matentzn:
No longer needed

PR #3360, opened 2024-09-12
Fix incompatibility with ZFA's "head".

### LINKED ISSUE #379: What is a head?
In ZFA, a head is analagous to a mammalian head+neck.

This causes problems when we bring in spatial disjointness (see issue #378 )

screen shot 2013-11-12 at 5 11 33 pm

The def of head is pretty vague:

```
def: "The head is the anterior-most division of the body [GO]." [GO:0060322, Wikipedia:Head]
```

What's a division?

We introduced a grouping class to accommodate this:
- [UBERON:0007811](http://purl.obolibrary.org/obo/UBERON_0007811) ! craniocervical region   [DEF: "The anteriormost subdivision of the body that includes the head, jaws, pharyngeal region and the neck (if present). In vertebrates this is the subdivision that includes the cervical vertebrae."]
  - [UBERON:0000033](http://purl.obolibrary.org/obo/UBERON_0000033) ! head [DEF: "The head is the anterior-most division of the body [GO]."]
  - [UBERON:0000974](http://purl.obolibrary.org/obo/UBERON_0000974) ! neck [DEF: "Subdivision of body proper, which consists of a maximal set of diverse subclasses of organ and organ part spatially associated with the cervical vertebral coumn, it is partially surrounded by skin of neck. Examples: There is only one neck[FMA]."]

Should ZFA head be mapped to this parent class?

For now, only consider vertebrates - let's not get into the insect head problem...

Note: objections have been raised in the past about the term "craniocervical region". But I think we need some kind of logical structure like this regardless of what we call it.

One option is to call the parent "head" and the child "tetrapod head", but this is ugly, and would make the ontology confusing for mammalian users who expect the head to be... the head.

gouttegd:
In ZFA, the "head" (ZFA:0001114) corresponds to what Uberon calls the "craniocervical region" (UBERON:0007811), comprising the head proper and the neck. So we remap ZFA:0001114 to UBERON:0007811, rather than UBERON:0000033).

Also, the axiom stating that 'pharyngeal arch' (UBERON:0002539) is part of the 'embryonic head' (UBERON:0008816) is bogus, as some pharyngeal arches are located in the neck rather than the head. So we create a new term 'embryonic craniocervical region' to be the embryonic counterpart of 'craniocervical region' (as 'embryonic head' is the embryonic counterpart of 'head'), and rewrite the aforementioned axiom to state that 'pharyngeal arch' is part of the 'embryonic craniocervical region'.

Combined together, these two changes fix about 80 unsatisfiable classes in the merge product of Uberon and ZFA.

closes #379


aleixpuigb:
#gogoeditdiff

aleixpuigb:
I am wondering if it would be benefitial to add a GCI `('part of' some embryo) and ('part of' some 'craniocervical region') SubClassOf 'part of' some 'embryonic craniocervical region'`. Otherwise I will approve this PR.

gouttegd:
> I am wondering if it would be benefitial to add a GCI ('part of' some embryo) and ('part of' some 'craniocervical region') SubClassOf 'part of' some 'embryonic craniocervical region'.

Sounds sensible, but unsure about how beneficial it would be.

Unfortunately the effects of adding such a GCI wouldn’t be shown by `gogoeditdiff`, because we’d need `robot materialize` (instead of just `robot reason`) to infer `'part of' some 'embryonic craniocervical region'` parents, and we cannot use `robot materialize` in GitHub actions because it’s too memory-intensive with Uberon.

PR #3363, opened 2024-09-18
Overhaul taxon subsets

### LINKED ISSUE #3362: Overhauling “taxon subsets”
Currently, each Uberon release comes with a handful of “taxon subsets”:

* `amniote-basic`,
* `euarchontoglires-basic`,
* `human-view`,
* `mouse-view`,
* `xenopus-view`.

Those subsets are generated in two different ways.

a) For the `-basic` subsets, the key part of their generation is OWLTools’ `--make-species-subset` command. Briefly, what this command does, when generating a subset for a taxon _T_, is to assert `owl:Thing in_taxon some T`, and then to exclude all classes that are unsatisfiable because of that assertion, so that the resulting ontology only contains classes that are not known to be invalid for taxon _T_. (After that step, this is merely standard, ROBOT-mediated reasoning and filtering of external terms.)

b) For the `-view` subsets, the approach is quite similar, only the implementation is different. To create a subset for a taxon _T_, instead of using the dedicated `--make-species-subset` command, we merge into Uberon a small ontology (found in `src/ontology/contexts/context-T.owl`) containing an axiom asserting that Uberon’s `anatomical structure` is part of the taxon _T_ (`UBERON:0000061 SubClassOf: part_of some T`), then we run OWLTools’ reasoning command over the merge product and instruct it to ignore all unsatisfiable classes (we have to use OWLTools here because ROBOT’s `reason` command does not allow, as far as I know, to simply ignore unsatisfiable classes – `robot reason` always errors out if the ontology it is trying to reason over is found to contain unsatisfiable classes). Basically it’s a “manual” way of doing more or less the same thing than `--make-species-subset` does.

I don’t know why we have those two methods to generate taxon subsets. Possibly OWLTools’ `--make-species-subsets` did not exist at the time the `-view` subsets were first created, or whoever created them did not know that command existed. In any case, I think it’s needlessly confusing to have two types of taxon subsets (the “-view” subsets and the “-basic” subsets), so those two methods should be consolidated into a single one – ideally one that does not require OWLTools, which we are trying to phase out of our pipelines.

@dosumis suggested at some point that instead of creating the subsets directly, we could have a command (implemented as a ROBOT plugin) that merely _tags_ classes as belonging to a given subset (by adding an annotation like `oboInOwl:inSubset <http://purl.obolibrary.org/obo/uberon/core#human_subset>`), after which the subsets themselves could be created by simply extracting the classes carrying the tag.

gouttegd:
This PR updates the way we are building “taxon subsets”.

As explained in #3362, we currently have, for reasons unknown (to me at least), two slightly different methods to create taxon subsets: one producing the `-view` subsets (`human-view`, `mouse-view`, `xenopus-view`) and one producing the `-basic` subsets (`amniote-basic`, `euarchotonglires-basic`). Both methods rely on the use of OWLTools.

This PR replaces both methods by a single one (so that all taxon subsets are produced in the same way) that relies on a new command in Uberon’s custom ROBOT plugin.

The PR does not change _which_ taxon subsets are produced and released by default (the five subsets aforementioned: `human`, `mouse`, `xenopus`, `amniote`, and `euarchotonglires`). 

More subsets can be produced on demand, all that is needed is to define a `TAXON_ID_subsetname` Make variable pointing to the desired NCBITaxon ID.

For example, to create a subset for, say, insects, one can do:

```sh
sh run.sh make subsets/insect-view.owl TAXON_ID_insect=NCBITaxon:50557
```

The PR also adds a possibility to create, not a subset directly, but a small component containing only `oboInOwl:inSubset` annotations to “tag” classes that belong to a taxon subset. For example:

```sh
sh run.sh make subsets/human-tags.ofn
```

would create a `human-tags.ofn` component containing, for all Uberon classes that belong to the human subset, `oboInOwl:inSubset <http://purl.obolibrary.org/obo/uberon/core#human_subset>` annotation assertion axioms. Such a component can then be merged with the main ontology for downstream use (e.g., extracting all the classes of the subset).

closes #3362

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -144,10 +144,6 @@ POSTPROCESS_ADDITIONS = subsets/human-tags.ofn \
 uberon.owl: $(POSTPROCESS_SRC) $(POSTPROCESS_ADDITIONS)
 	$(ROBOT) merge $(foreach prereq,$^,-i $(prereq)) -o $@
 
-uberon.json.gz: uberon.json
-	gzip -c $< > $@.tmp && mv $@.tmp $@
-.PRECIOUS: uberon.json.gz
-
 
 # ----------------------------------------
 # MIRRORS
@@ -1513,6 +1509,10 @@ refresh-external-resources:
 # Everything that does not belong to any of the sections above.
 # May include both actually useful stuff and stuff that nobody
 # remembers why it was written in the first place.
+
+uberon.json.gz: uberon.json
+	gzip -c $< > $@.tmp && mv $@.tmp $@
+.PRECIOUS: uberon.json.gz
 
 TEMPLATESDIR=templates
 
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -142,7 +142,7 @@ $(POSTPROCESS_SRC): $(OWLSRC) $(BRIDGEDIR)/uberon-bridge-to-bfo.owl $(DEVELOPS_F
 POSTPROCESS_ADDITIONS = subsets/human-tags.ofn \
 			subsets/mouse-tags.ofn
 uberon.owl: $(POSTPROCESS_SRC) $(POSTPROCESS_ADDITIONS)
-	$(ROBOT) merge $(foreach prereq,$^,-i $(prereq)) -o $@
+	$(ROBOT) merge -i $< $(foreach add,$(POSTPROCESS_ADDITIONS),-i $(add)) -o $@
 
 
 # ----------------------------------------
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -9,6 +9,7 @@ TAXMODSDIR =           taxmods
 BRI=                   true
 
 OWLSRC =               $(TMPDIR)/uberon-edit.owl
+POSTPROCESS_SRC =      $(TMPDIR)/uberon.owl
 CATALOG_DYNAMIC =      catalog-dynamic.xml
 OWLTOOLS_NO_CAT=       OWLTOOLS_MEMORY=$(OWLTOOLS_MEMORY) owltools
 OWLTOOLS_CAT_DYNAMIC=  OWLTOOLS_MEMORY=$(OWLTOOLS_MEMORY) owltools --catalog-xml $(CATALOG_DYNAMIC)
@@ -127,14 +128,21 @@ MATERIALIZE=
 else
 MATERIALIZE = materialize -T $(CONFIGDIR)/basic_properties.txt -r elk
 endif
-uberon.owl: $(OWLSRC) $(BRIDGEDIR)/uberon-bridge-to-bfo.owl $(DEVELOPS_FROM_CHAIN)
+$(POSTPROCESS_SRC): $(OWLSRC) $(BRIDGEDIR)/uberon-bridge-to-bfo.owl $(DEVELOPS_FROM_CHAIN)
 	$(ROBOT) merge -i $(OWLSRC) -i $(BRIDGEDIR)/uberon-bridge-to-bfo.owl \
 	               -i $(DEVELOPS_FROM_CHAIN) \
 	         relax $(MATERIALIZE) \
 	         reason -r elk --exclude-duplicate-axioms true \
 	                       --equivalent-classes-allowed asserted-only \
 	         unmerge -i $(DEVELOPS_FROM_CHAIN) \
-	         annotate -O $(URIBASE)/$@ -V $(RELEASE)/$@ -o $@
+	         annotate -O $(URIBASE)/uberon.owl -V $(RELEASE)/uberon.owl -o $@
+
+# Step 3: Postprocessing. We merge some files that are derived from the
+# previous step.
+POSTPROCESS_ADDITIONS = subsets/human-tags.ofn \
+			subsets/mouse-tags.ofn
+uberon.owl: $(POSTPROCESS_SRC) $(POSTPROCESS_ADDITIONS)
+	$(ROBOT) merge $(foreach prereq,$^,-i $(prereq)) -o $@
 
 uberon.json.gz: uberon.json
 	gzip -c $< > $@.tmp && mv $@.tmp $@
@@ -728,7 +736,7 @@ extra-full-bridge-checks: $(foreach ao, $(EXTRA_FULL_CHECK_AO_LIST), $(REPORTDIR
 
 # A quick bridge check uses only uberon plus taxon constraints plus
 # bridging axioms, *not* the axioms in the source ontology itself.
-$(REPORTDIR)/quick-bridge-check-%.txt: uberon.owl \
+$(REPORTDIR)/quick-bridge-check-%.txt: $(POSTPROCESS_SRC) \
 				       $(COMPONENTSDIR)/external-disjoints.obo \
 				       $(TMPDIR)/taxslim-disjoint-over-in-taxon.owl \
 				       $(TMPDIR)/bridges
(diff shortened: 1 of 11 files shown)


src/ontology/uberon.Makefile
```diff
@@ -934,40 +934,59 @@ subsets/life-stages-core.owl: uberon.owl
...
+#            taxon T, assert 'R SubClassOf: in_taxon some T' and exclude
+#            all classes that are unsatisfiable because of that
+#            assertion.
+# - precise: alternative strategy; given a root class R and a taxon T,
+#            iterate through all subclasses C of R and include classes
+#            for which the expression 'C and in_taxon some T' is
+#            satisfiable; about 3-5 times slower than 'default'.
+TAXON_SUBSET_STRATEGY = default
+
+TAXON_SUBSET_ROOTS = UBERON:0001062 UBERON:0000000
```
matentzn, line 958: Does this second root exist? I don't think I have ever seen it

src/ontology/uberon.Makefile
```diff
@@ -934,40 +934,59 @@ subsets/life-stages-core.owl: uberon.owl
...
+	$(ROBOT) uberon:create-species-subset --input $< \
+		                              --taxon $(TAXON_ID_$*) \
+		                              --strategy $(TAXON_SUBSET_STRATEGY) \
+		                              --reasoner ELK \
+		                              $(foreach root,$(TAXON_SUBSET_ROOTS),--root $(root)) \
+		 reason --reasoner ELK --equivalent-classes-allowed all \
+		        --exclude-tautologies structural \
 		 relax \
 		 remove --axioms equivalent \
 		 relax \
```
matentzn, line 972: Can't imagine what this second relax does but well since it's there..

src/ontology/uberon.Makefile
```diff
@@ -934,40 +934,59 @@ subsets/life-stages-core.owl: uberon.owl
...
+	$(ROBOT) uberon:create-species-subset --input $< \
+		                              --taxon $(TAXON_ID_$*) \
+		                              --strategy $(TAXON_SUBSET_STRATEGY) \
+		                              --reasoner ELK \
+		                              $(foreach root,$(TAXON_SUBSET_ROOTS),--root $(root)) \
+		                              --no-remove \
+		                              --prefix 'uberon: http://purl.obolibrary.org/obo/uberon/core#' \
+		                              --subset-name uberon:$*_subset \
+		                              --only-tag-in UBERON: --only-tag-in GO: \
+		                              --write-tags-to $@
```
matentzn, line 989: I personally have a very weak preference to avoid Java code however well designed because of the lack of people being able to maintain it, but this is not a review comment now, just an FYI that you should ignore. I am assuming the Java code does some things beyond the capability of a simple SPARQL query here.

matentzn (approved):
I have reviewed the code changes, they look great, - I would like to offer a single word of caution: renaming files, even subset files, may break existing pipelines somewhere on the deep web unless we also add a purl redirect to the OBO purl config.

src/ontology/uberon.Makefile
gouttegd: Yes it does. It’s '[processual entity](http://purl.obolibrary.org/obo/UBERON_0000000)', which is the parent of '[life cycle](http://purl.obolibrary.org/obo/UBERON_0000104)' and all the '[life cycle stage](http://purl.obolibrary.org/obo/UBERON_0000105)' terms.

src/ontology/uberon.Makefile
gouttegd: I personally have a very strong preference for a well-designed, small piece of code (in any language) that does one thing and does it well, over a complicated pipeline that chains two or three tools and/or needs several intermediate representations of the ontology in order to achieve something. :)

gouttegd:
> renaming files, even subset files, may break existing pipelines somewhere on the deep web unless we also add a purl redirect to the OBO purl config.

I know. But precisely, this is to be handled at the PURL level, which is here exactly for this purpose. There is no reason for us to refrain from renaming files if it brings better consistency (in this case, by having all taxon subsets consistently named `something-view.owl` instead of having some subsets being named `something-basic`).

dosumis:
Hi @gouttegd - great to see this.  

I'm especially excited to see this:

> The PR also adds a possibility to create, not a subset directly, but a small component containing only oboInOwl:inSubset annotations to “tag” classes that belong to a taxon subset. For example:
>
> `sh run.sh make subsets/human-tags.ofn`
>
> would create a human-tags.ofn component containing, for all Uberon classes that belong to the human subset, oboInOwl:inSubset <http://purl.obolibrary.org/obo/uberon/core#human_subset> annotation assertion axioms. Such a component can then be merged with the main ontology for downstream use (e.g., extracting all the classes of the subset).

I would really like the tags to in incorporated into the release files.  We can use them straight away in our autosuggest pipelines to boost for species relevant terms.

src/ontology/uberon.Makefile
matentzn: Fair enough :P Lets do the dance on a case by case basis. In this case here I agree!

gouttegd:
> I would really like the tags to in incorporated into the release files.

I wasn’t sure whether this was your preferred option, so for now the idea was to merely produce the `-tags` file and leave downstream users of Uberon merge them with `uberon.owl` if they wanted.

But it certainly can be done directly upstream if preferred.

Do we want _all_ release artefacts to include those tags (e.g. `uberon.owl`, `oberon-basic.owl`, but also all the organ-specific subsets such as `nephron-minimal`, `sensory-minimal`, etc)? Or just _one_ supplementary artefact containing the taxon subset tags (e.g. something like `uberon-with-subset-tags.owl`)?

Based on your comment I assume the former (tags included in all release artefacts), in which case we’ll need a new intermediate file (upstream of `uberon.owl`) from which the subset tags can be generated before we produce the final `uberon.owl` with the subset tags included.

gouttegd:
PR updated to include the `oboInOwl:inSubset` tags directly into the release products.

For now, we include only the tags for the human subset and the mouse subset. To add tags for another subset, it would simply be a matter of adding another tag file to the `POSTPROCESS_ADDITIONS` Make variable, for example:

```Make
POSTPROCESS_ADDITIONS = subsets/human-tags.ofn \
                        subsets/mouse-tags.ofn \
                        subsets/drosophila-tags.ofn
```

Some considerations:

1) The tags are added as part of the release pipeline, and are _not_ imported from the `-edit` file. This means the tags will _not_ be visible when working on the ontology from Protégé. This is on purpose. The tags are automatically computed and are not supposed to be manually edited, but there would be no way to enforce that if they were visible from Protégé.

2) The tags are added to the main `uberon.owl` artefact, and from there are propagated to all files that are derived from this artefact. They are, however, explicitly excluded from Composite-Metazoan (and its progenitor Collected-Metazoan). Including them into Composite-Metazoan would produce a somewhat weird result in that **Uberon** terms would be tagged as belonging to a given taxon subset, but terms coming from the taxon-specific anatomy ontologies would not be tagged. For example, UBERON’s `artery` would be tagged as belonging to the mouse subset, but not EMAPA’s `superior vesicle artery`, even though that term is (obviously, as coming from EMAPA) definitely applicable to mice. I believe this would be confusing.

3) The additions of the mouse and human tags are adding about ~3MB to the main `uberon.owl` file (from 88MB to 91MB). Not sure whether this is a big concern or not, but I think it should be noted, especially if we want to add more taxon subset tags to the main release file.

matentzn (approved):
assuming that uberon.owl was swapped out to `POST_PROCESS` in all the relevant places in the makefile, this looks good to me!

cmungall (approved):

gouttegd:
> assuming that uberon.owl was swapped out to POST_PROCESS in all the relevant places in the makefile

Only two places, in fact:

1) As the input of the Composite-Metazoan pipeline, because as I have said in my last comment, we (at least I) do not want the taxon subset tags to be included in Composite-Metazoan, so it’s fine to build Composite-Metazoan from the `POSTPROCESS_SRC` intermediate file (which is the same thing as the final `uberon.owl` minus the taxon subset tags);

2) As the input of the “bridge checks”, because the taxon subset tags are completely unnecessary for those checks (they would cause no harm, but would serve no purpose either).

All other custom pipelines that were dependent on `uberon.owl` are _still_ dependent on `uberon.owl`, meaning those custom pipelines now work on a version of the ontology that includes the taxon subset tags. Notably, this means that the “system subsets” (e.g. `circulatory-minimal`, `digestive-minimal`, etc.) do contains those tags.

gouttegd:
Merging here.

@dosumis: (1) If you want more taxon subset tags than just human and mouse to be included in the main product, just say so and we can add new subsets anytime, it’s one line to add in the custom Makefile.

(2) I assume you want to see the same thing in CL?

PR #3372, opened 2024-09-23
tracheal mucosa: making label, def, logical def all consistent.

cmungall:


matentzn:
Thank you @KrishnaTO, I have confirmed and concur.

PR #3374, opened 2024-09-23
Adding gross anatomical part. Fixes #2921

### LINKED ISSUE #2921: NTRs: gross anatomical part, animal gross anatomical part
Add a generic GAP, as discussed in:

 - https://github.com/OBOFoundry/COB/issues/45

Uberon would also make a metazoan GAP

See:

 - https://github.com/geneontology/go-ontology/issues/22994

cmungall:


cmungall:
image

cmungall:
This is turning out to be harder than I expected

gouttegd:
Regarding the last QC failure, I believe you have run into what should be considered a bug in the ODK. I have opened [an issue with the details](https://github.com/INCATools/ontology-development-kit/issues/1107) on the ODK repo.

A quick (but crude) workaround until the issue is fixed in the ODK would be to edit the standard `Makefile` to change the definition of `$(SRCMERGED)` from

```Makefile
SRCMERGED     = $(TMPDIR)/merged-$(SRC)
```

into

```Makefile
SRCMERGED     = $(TMPDIR)/merged-uberon-edit.ofn
```

to force the intermediate file on which the ROBOT `report` command is run to be written in OFN rather than OBO format, thereby avoiding OBO serialisation quirks.

PR #3377, opened 2024-10-08
Generate the release diff with the OAK output

aleixpuigb:
Fixes issues with reaching the limit number of characters in the release notes.


gouttegd (changes requested):
No objection to the generation of another diff with OAK, but since we already have a `src/ontology/diffs` directory with a small `Makefile` to generate the existing diffs, I’d rather have the new rule added to that Makefile than directly to the `uberon.Makefile`. Especially since `src/ontology/diffs/Makefile` already contains a rule to download the latest released OBO version of the ontology, so there’s no need to duplicate that in `uberon.Makefile`.

gouttegd (approved):

aleixpuigb:
Thank you @gouttegd!

PR #3387, opened 2024-10-24
Peyer's patch epithelium name update

JABelfiore:
Name change from Peyer's patch epithelium to follicle associated epithelium as this is the more common term. Peyer's patch epithelium is now a synonym.


dosumis:
@Jasmine - looks good, but there might be a wrinkle  - there are immune follicles in the appendix in humans and in the cecum of mice.  FAE may be used for the covering of these, in which case Peyer's patch epithelium would be a subclass.  See https://pmc.ncbi.nlm.nih.gov/articles/PMC10033573/figure/Fig1/.   

Can you ask some Gut Cell Atlas experts what they would recommend?

AvolaAmg (approved):

AvolaAmg:
#gogoeditdiff

PR #3397, opened 2024-10-25
new term aorta-gonad-mesonephros

JABelfiore:
New term added


src/ontology/uberon-edit.obo
```diff
@@ -191,6 +191,7 @@ property_value: doap-SVNRepository "https://obo.svn.sourceforge.net/svnroot/obo/
 property_value: foaf-homepage "http://uberon.org" xsd:anyURI
 property_value: has_ontology_root_term UBERON:0000104
 property_value: has_ontology_root_term UBERON:0001062
+owl-axioms: Prefix(owl:=<http://www.w3.org/2002/07/owl#>)\nPrefix(rdf:=<http://www.w3.org/1999/02/22-rdf-syntax-ns#>)\nPrefix(xml:=<http://www.w3.org/XML/1998/namespace>)\nPrefix(xsd:=<http://www.w3.org/2001/XMLSchema#>)\nPrefix(rdfs:=<http://www.w3.org/2000/01/rdf-schema#>)\n\n\nOntology(\nDeclaration(AnnotationProperty(<http://www.geneontology.org/formats/oboInOwl#hasExactSynonym>))\n\n\nAnnotationAssertion(<http://www.geneontology.org/formats/oboInOwl#hasExactSynonym> <http://purl.obolibrary.org/obo/UBERON_8700000> \"\")\n)
```
gouttegd, line 194: @JABelfiore This line here is what is making the QC fail.

Protégé will automatically add this `owl-axioms` line whenever you create (intentionally or not) a OWL axiom that cannot be represented in the OBO format.

In Uberon, we explicitly check, as part of the QC suite, that no such line is ever inserted, because there should never be any need for an editor to create an axiom that is outside of the subset of OWL that is compatible with the OBO format.

In this instance, the problematic axiom is

```
AnnotationAssertion(oboInOwl:hasExactSynonym UBERON:8700000 "")
```

That is, you tried to add an _empty_ exact synonym to that new “aorta-gonad-mesonephros” term (UBERON:8700000). The OBO format cannot handle that (in the OBO format a synonym cannot be the empty string), so Protégé wrote the axiom in OFN syntax in a newly added `owl-axioms` line -- thereby triggering the QC failure.

How to fix: In Protégé, either replace that empty synonym by an actual synonym for “aorta-gonad-mesonephros”, or if you do not have any synonym for it, remove the empty synonym entirely. When you’ll save the ontology again, the `owl-axioms` line should no longer be there.

gouttegd (changes requested):

src/ontology/uberon-edit.obo
JABelfiore: @gouttegd amazing, will do that now. Thanks for looking into this for me!

PR #3399, opened 2024-11-01
Uberon release 2024-10-31

aleixpuigb:
DO NOT MERGE!

### DIFF
--- a/src/ontology/bridge/cl-bridge-to-caro.owl
+++ b/src/ontology/bridge/cl-bridge-to-caro.owl
@@ -8,7 +8,7 @@
      xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
      xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#">
     <owl:Ontology rdf:about="http://purl.obolibrary.org/obo/uberon/bridge/cl-bridge-to-caro.owl">
-        <owl:versionIRI rdf:resource="http://purl.obolibrary.org/obo/uberon/releases/2024-10-31/bridge/cl-bridge-to-caro.owl"/>
+        <owl:versionIRI rdf:resource="http://purl.obolibrary.org/obo/uberon/releases/2024-11-01/bridge/cl-bridge-to-caro.owl"/>
     </owl:Ontology>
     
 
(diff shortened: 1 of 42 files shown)


aleixpuigb:
Many terms are being classified as SubclassOf '[protein-containing material entity](http://purl.obolibrary.org/obo/PR_000050567)', which has been obsoleted in GO. Probably is best to fix this before running the release, what do you think? @dosumis @gouttegd @cmungall

dosumis:
It's obsolete in PRO https://www.ebi.ac.uk/ols4/ontologies/pr/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FPR_000050567

So should be possible to fix by fixing import.

@gouttegd  - Assume we are now using base imports so we shouldn't be getting this indirectly?

gouttegd:
@dosumis We are using base imports, but that didn’t help here because the faulty axioms seem to come directly from PR itself. I just tried refreshing my local PR mirror here, and it does not contain the deprecation axiom for PR:000050567, and to the contrary it contains many **SubClassOf** axioms that refers to PR:000050567 as the superclass, as if the entity had never been deprecated. Maybe something wrong happened on the PR side, I don’t know.

gouttegd:
> the faulty axioms seem to come directly from PR itself.

Correction: the faulty axioms come from [the OBO PR slim](https://github.com/obophenotype/pro_obo_slim), not from PR itself -- we do not import PR directly as it is too big.

The [last release of the OBO PR slim](https://github.com/obophenotype/pro_obo_slim/releases/tag/v2024-10-24) does contain the faulty PR:000050567 axioms, while the current version of PR available on http://purl.obolibrary.org/obo/pr.owl (dated 8 August) does not, so I am not sure what happened. Possibly the last OBO PR slim was somehow made from an older version of PR, I don’t know.

gouttegd:
I’ve made a PR to [refresh the OBO PR slim](https://github.com/obophenotype/pro_obo_slim/pull/50). Once merged, refreshing the imports once again in Uberon should clear the issue.

aleixpuigb:
Thank you @gouttegd, I will close this PR and refresh the imports. Once it is done I will run again the release.

PR #3400, opened 2024-11-02
Refresh imports after obo PR slim update

aleixpuigb:
Refresh imports after obo PR slim update


gouttegd (approved):
OK, this time PR:000050567 (the deprecated `protein-containing material entity`) is not there. All good.

PR #3415, opened 2024-11-10
Update foreign SSSOM sets from the ontology.

gouttegd:
This PR amends the rules that refresh the externally maintained SSSOM mapping sets (the FBbt set and the Biomappings set) to include a step that check those sets against the current version of the ontology to

(1) remove any mapping to a Uberon entity that does not exist or is deprecated;
(2) add any missing label, using the labels found in the ontology.

This necessitates, as a preliminary:

(1) making sure we are using ODK 1.5.3, as previous versions do not include a recent enough version of SSSOM-Java;
(2) removing a bogus empty `xml:base` attribute in the XML catalog, that prevents `sssom-cli` from reading it.

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -1257,7 +1257,8 @@ ifeq ($(strip $(IMP)),true)
 $(MAPPINGDIR)/fbbt.sssom.tsv: .FORCE
 	wget "http://purl.obolibrary.org/obo/fbbt/fbbt.sssom.tsv" -O - | \
 		sssom-cli --prefix-map-from-input \
-		          --rule 'object==UBERON:* -> include()' \
+		          --include 'object==UBERON:*' \
+		          --update-from-ontology=$(SRC):object,label,existence \
 		          --output $@
 
 # CL mapping set. We simply fetch it as it is from CL. It already
@@ -1276,7 +1277,8 @@ $(MAPPINGDIR)/biomappings.sssom.tsv: $(TMPDIR)/biomappings.sssom.tsv \
 	sssom-cli --input $< --prefix 'UBERON=http://purl.obolibrary.org/obo/UBERON_' \
 		  --rule '!(subject==UBERON:* || object==UBERON:*) -> stop()' \
 		  --rule 'object==UBERON:* -> invert()' \
-		  --include-all | \
+		  --include-all \
+		  --update-from-ontology=$(SRC):subject,label,existence | \
 	sssom-cli --mangle-iris obo --output $@
 
 $(TMPDIR)/biomappings.sssom.tsv:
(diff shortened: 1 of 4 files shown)


src/ontology/uberon.Makefile
```diff
@@ -1276,7 +1277,8 @@ $(MAPPINGDIR)/biomappings.sssom.tsv: $(TMPDIR)/biomappings.sssom.tsv \
 	sssom-cli --input $< --prefix 'UBERON=http://purl.obolibrary.org/obo/UBERON_' \
 		  --rule '!(subject==UBERON:* || object==UBERON:*) -> stop()' \
 		  --rule 'object==UBERON:* -> invert()' \
-		  --include-all | \
+		  --include-all \
+		  --update-from-ontology=$(SRC):subject,label,existence | \
```
matentzn, line 1281: Nothing here looks particularly wrong or strange, but maybe a super shart explainer of this new option. I read https://incenp.org/dvlpt/sssom-java/sssom-cli.html, very nice;

> If the subject or the object of a mapping does not exist in the ontology or is deprecated, the mapping is removed from the set.


In order for "existence" to be fulfilled, what are the exact conditions?

1. An OWL declaration exists
2. The ID is used in an axiom
3. The ID is used as the object of an xref assertion

Thanks!

matentzn (approved):

src/ontology/uberon.Makefile
gouttegd: > In order for "existence" to be fulfilled, what are the exact conditions?
>
> 1. An OWL declaration exists
> 2. The ID is used in an axiom

The IRI of the entity (in this case, the mapping’s `subject_id`) must be in the ontology’s signature. This means that the IRI must either:

1. be the subject of an OWL declaration axiom, or
2. be used in at least one axiom **that is enough to allow inferring the type of entity** (e.g. a _SubClassOf_ axiom to infer that the referred entity is a class).

(In fact condition (1) is merely a special case of condition (2): a declaration axiom is precisely “an axiom that is enough to allow inferring the type of the entity”.)

gouttegd: So for example, if the only axiom the ontology contains about entity `EX:0001` is a label annotation like this:

```
AnnotationAssertion(rdfs:label EX:0001 "my class")
```

then we don’t know what `EX:0001` is (class, property, individual?) and so it is not considered part of the signature (at least this is how the OWLAPI sees it). Therefore, from SSSOM-CLI’s point of view, that entity would be considered _not_ to exist, and under the `existence option a mapping with EX:0001 would be removed.

matentzn: Just as an irrelevant aside:

> (In fact condition (1) is merely a special case of condition (2): a declaration axiom is precisely “an axiom that is enough to allow inferring the type of the entity”.)

didn't used to be true. Formally, yes, but not practically. Declarations were injected and write-time if they were missing.. `x.signature()` methods behave correctly though!

PR #3416, opened 2024-11-11
Remove injections about RO:has_skeleton.

gouttegd:
In RO, 'has skeleton' (RO:0002551) is domain-restricted to 'anatomical structure' (UBERON:0000061) and range-restricted to 'material anatomical entity' (UBERON:0000465).

Currently, in Uberon we are overriding those restrictions by injecting tigher constraints: a domain restriction to 'organism subdivision' (UBERON:0000475) and a range restriction to 'subdivision of skeleton' (UBERON:0010912).

A ticket to upstream those tighter restrictions to RO has gone nowhere in more than one year
(https://github.com/oborel/obo-relations/issues/700).

This PR removes the Uberon-injected restrictions and leaves it to RO to decide whether they should adopt them or not.

Of note, the only inferred axiom that those tighter restrictions give us is this one:

> 'nose' SubClassOf: 'organism subdivision'

So we don't lose much by removing them until they are (maybe) added back directly by RO.

Related to #2453


dosumis (approved):
Looks good.  The idea of the tighter domain and range is for consistency checks, but I agree they belong in RO.

PR #3418, opened 2024-11-13
Remove Anita from CODEOWNERS

matentzn:
As per discussion with @anitacaron we remove her from codeowners so she is not assigned to everything automatically.

Anyone wanting to take up that huge burden please put your hands up.


anitacaron (approved):
Thanks! 🙏🏽

PR #3429, opened 2024-11-18
Add ID range for the NDI Cloud project.

gouttegd:
This PR assigns the ID range 8,910,000--8,919,999 to members of the [NDI](https://github.com/VH-Lab/NDI-matlab) Cloud project, so that they can participate in creating the terms they need to cover some aspects of arthropod neuroanatomy.

As discussed in Uberon call of 2024-11-18.

### DIFF
--- a/src/ontology/uberon-idranges.owl
+++ b/src/ontology/uberon-idranges.owl
@@ -262,3 +262,9 @@ Datatype: idrange:35
         allocatedto: "Tiago Lubiana"
     EquivalentTo:
         xsd:integer[>= 8900000, <= 8909999]
+
+Datatype: idrange:40
+    Annotations:
+        allocatedto: "NDI Cloud"
+    EquivalentTo:
+        xsd:integer[>= 8910000, <= 8919999]


src/ontology/uberon-idranges.owl
```diff
@@ -261,4 +261,10 @@ Datatype: idrange:35
     Annotations:
         allocatedto: "Tiago Lubiana"
     EquivalentTo:
-        xsd:integer[>= 8900000, <= 8909999]
\ No newline at end of file
+        xsd:integer[>= 8900000, <= 8909999]
+
+Datatype: idrange:40
```
aleixpuigb, line 266: It needs indentation for this line.

gouttegd: I don’t mind adding indentation, but it should not be required. AFAIK with or without indentation the file is a valid Manchester file, and neither Protégé nor ROBOT has any trouble parsing it.

aleixpuigb (approved):

PR #3435, opened 2024-11-25
Uberon release v2024-11-25

aleixpuigb:


subsets/life-stages-composite.obo
```diff
@@ -1451,12 +1467,12 @@ is_a: BFO:0000004 ! independent continuant
 is_a: CHEBI:138675 ! gas molecular entity
 is_a: CHEBI:35107 ! azane
 is_a: CHEBI:37176 ! mononuclear parent hydride
-relationship: has_role CHEBI:228364 ! NMR chemical shift reference compound
-relationship: has_role CHEBI:50910 ! neurotoxin
-relationship: has_role CHEBI:59740 ! nucleophilic reagent
-relationship: has_role CHEBI:75771 ! mouse metabolite
-relationship: has_role CHEBI:77941 ! EC 3.5.1.4 (amidase) inhibitor
-relationship: has_role CHEBI:78433 ! refrigerant
+relationship: RO:0000087 CHEBI:228364 ! has role NMR chemical shift reference compound
```
gouttegd, line 1470: For information: I believe those changes, and all similar changes in that file, are a normal and expected consequence of 61e7f2f. We no longer automatically create “shorthands” (such as `has_role`) for every single RO relation, so in the OBO release products the relations appear with their CURIEs instead.

rays22 (approved):
Looks OK.

subsets/life-stages-composite.obo
gouttegd: Likewise for all the similar changes in `uberon.obo`.

aleixpuigb:
Is the creation of the file [uberon-lastbuild.obo](src/ontology/diffs/uberon-lastbuild.obo) expected? I didn't see in previous Uberon releases. I am trying to find where does it come from.

gouttegd:
@aleixpuigb It’s a consequence of #3377. The _creation_ of the file is normal, but it is probably best to avoid committing it to the repository. You can remove it, and then avoid re-committing it in the future, you can add the following line to the `.gitignore` file:

```
uberon-lastbuild.obo
```

(There is already a [similar line](https://github.com/obophenotype/uberon/blob/a77f1aa827900cac9cd62d5e02708760ff86d60a/.gitignore#L115) to avoid committing `uberon-lastbuild.owl`, but the OBO version was somehow forgotten.)

aleixpuigb:
Thank you, I will do that.

gouttegd (approved):
Looks all good to me.

PR #3439, opened 2024-11-28
Uncouple bridge generation from mappings refresh.

### LINKED ISSUE #3438: Tech: Version IRIs in bridge files
Right now the bridges are refreshed through SOP rather than piggy-bagging on any of the regular process link import refresh or release.

**Problem:** As discussed in #3434, the fact that bridges are not tied to releases makes version IRIs a bit weird because they may not correspond to any existing release, which means the version IRIs wont resolve, defeating the purpose of versioned PURLs.

@gouttegd suggested due to the fact:

> The problem is not that it is incomplete, it is that the bridge version IRIs are potentially wrong. Whenever the Uberon release happens on a different day than the day the bridges were last re-generated, the bridges end up with a version IRI that has a different date than the date of the Uberon release itself. This happened for example last August.

to drop the altogether.

My argument against this was that it is useful for downstream applications consuming the bridges to know _when they were last generated_.

**Proposal:** Without thinking, my first intuition was to couple bridge files to the release. However, this is not quite as trivial, as the accurate generation of bridge files requires updating the imports, which we have deliberately decoupled from the release process to reduce points of failures. 

Before making things complicated: I do not really care about version IRIs. My simple proposal is to replace `ontology-version` with `dc-modified` (http://purl.org/dc/terms/modified), and replace the header in bridges.dispatch by:

```
[__default]
ontology-iri: http://purl.obolibrary.org/obo/uberon/bridge/%filename.owl
dc-maintained: %date
```

That will satisfy my desire to look "how old a bridge is" and I hope @gouttegd wish to not include wrong version IRIs..

gouttegd:
We change the `refresh-external-resources` target so that it still refreshes (1) the imports, (2) the "local" imports (in fact the local copies of the foreign ontologies used to build Composite Metazoan), (3) the externally maintained mapping sets, (4) the externally maintained bridges; **but** it does not re-generate the bridge files that are dependent on the refreshed mappings.

The mappings-derived bridges are now expected to be built at release time (and also at QC time). We build them under `MIR=false`, so they will only be built using the locally available mappings (from the last time they were refreshed), without triggering the download of any remote resource.

Editors/maintainers can still locally trigger the building of the bridge files at any time by explicitly invoking the `refresh-bridges` target, if they want/need to do so.

This does not change anything for whoever is running the refresh pipeline nowadays: it’s still only a matter of invoking `refresh-external-resources`. The only difference is that this won't trigger a rebuild of the bridge files.

Two benefits: the “refresh imports” PRs will no longer be polluted by needless changes in the bridge files, in which only the version IRI has changed; the bridge files will always be generated (at least) at the same time as the release, so they will always have a correct version IRI.

Minor cosmetic changes also in this commit:

* `CUSTOM_BRIDGES` is renamed to `EXTERN_BRIDGES`, which reflects better the true nature of those bridges (it does not really matter that they are “custom”, what matters is that they are not generated here).

* We use `MIR=false`, rather than `IMP=false`, to prevent the downloading of any remote resource. This is for consistency with other parts of the Makefile.

closes #3438

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -1249,8 +1249,8 @@ EXTERNAL_SSSOM_PROVIDERS = fbbt cl biomappings
 # All the sets coming from the above ontologies.
 EXTERNAL_SSSOM_SETS = $(foreach provider, $(EXTERNAL_SSSOM_PROVIDERS), $(MAPPINGDIR)/$(provider).sssom.tsv)
 
-# We only refresh external resources under IMP=true
-ifeq ($(strip $(IMP)),true)
+# We only refresh external resources under MIR=true
+ifeq ($(strip $(MIR)),true)
 
 # FBbt mapping set. We filter it to only keep UBERON-related mappings
 # (CL-related mappings are already bundled with the CL set).
@@ -1296,7 +1296,7 @@ endif
 ifeq ($(BRI),true)
 
 # Those bridges are generated separately (see below).
-CUSTOM_BRIDGES = $(BRIDGEDIR)/uberon-bridge-to-mba.owl \
+EXTERN_BRIDGES = $(BRIDGEDIR)/uberon-bridge-to-mba.owl \
 		 $(BRIDGEDIR)/uberon-bridge-to-dmba.owl
 
 # 1. Prepare the ruleset file.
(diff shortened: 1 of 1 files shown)


matentzn (approved):
I scanned it 12 times but cant find any obvious issues! LGTM!

PR #3440, opened 2024-11-28
Make the SSSOM/T ruleset future-proof.

gouttegd:
There is one particular bit of syntax that is currently used in the SSSOM/T ruleset that I would like to disallow in the upcoming version of SSSOM/Java. Specifically, the dereferencing of a variable enclosed in angled brackets, as in:

```
  "... and (<%TAXREL> some NCBITaxon:????)"
```

It should either be `%TAXREL` (no angled brackets, preferred form as it is more readable) or `<%{TAXREL}>` (angled brackets + variable name enclosed in curly brackets — cumbersome, but more consistent throughout SSSOM/T).

The first form is already allowed with current versions of SSSOM-Java, so we can switch to it immediately without breaking anything. And this will ensure nothing will also break when the upcoming SSSOM-Java will land in the ODK.


matentzn (approved):
LGTM based on your explanations!

PR #3443, opened 2024-12-06
Build the “composite“ life stages ontology directly in Uberon

gouttegd:
(Draft PR as this is a work that depends on:

* an overhaul of the “Species-Specific Life Stages Ontology“ project (https://github.com/obophenotype/developmental-stage-ontologies/pull/104);
* an updated version of Uberon’s ROBOT plugin.)

This PR changes the way Uberon interacts with the “Species-Specific Life Stages Ontology” (SSLSO) project. Roughly, instead of relying on that project to provide us with a “pre-composited” version of the life stage ontologies, we do everything here in Uberon. All SSLSO has to do is to provide us with the mappings between their terms and the corresponding taxon-neutral terms in Uberon.

There are several reasons for such a change, the most important being that it keeps all the logic to create the “composite” ontologies in the same place, here in Uberon. Having the SSLSO perform its own compositing leads to a lot of duplicated code, a lot of unnecessary back-and-forth between Uberon and SSLSO (Uberon generating the bridges with FBdv and WBls, which are then fetched by SSLSO to produce `ssso-merged-uberon`, which is then fetched by Uberon to produce `composite-metazoan`), and a risk that the two composite pipelines (the one in SSLSO and the one in Uberon) are not kept in sync and therefore behave slightly differently (which is exactly the case currently).


src/ontology/bridge/collected-lifestages-hdr.owl
matentzn, line 1: noting the "-hdr" in the file name which at least at first glance does not match what is in the makefile

src/ontology/uberon.Makefile
```diff
@@ -1101,11 +1097,24 @@ COLLECTED_vertebrate_SOURCES =       $(COLLECTED_tetrapod_SOURCES) \
...
+				     $(IMPORTDIR)/local-fbdv.owl \
+				     $(IMPORTDIR)/local-wbls.owl \
+				     $(IMPORTDIR)/local-mmusdv.owl \
+				     $(IMPORTDIR)/local-hsapdv.owl \
+				     $(IMPORTDIR)/local-sslso.owl \
+				     $(BRIDGEDIR)/uberon-bridge-to-fbdv.owl \
+				     $(BRIDGEDIR)/uberon-bridge-to-wbls.owl \
+				     $(BRIDGEDIR)/uberon-bridge-to-mmusdv.owl \
+				     $(BRIDGEDIR)/uberon-bridge-to-hsapdv.owl \
+				     $(BRIDGEDIR)/uberon-bridge-to-sslso.owl
```
matentzn, line 1116: Not actionable, just informative: I am noting that this seems to be missing a very important organism, Xenopus: https://www.ebi.ac.uk/ols4/ontologies/xao/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FXAO_1000000?lang=en

src/ontology/uberon.Makefile
```diff
@@ -1101,11 +1097,24 @@ COLLECTED_vertebrate_SOURCES =       $(COLLECTED_tetrapod_SOURCES) \
...
 				     $(COLLECTED_drosophila_SOURCES) \
 				     $(COLLECTED_worm_SOURCES) \
-				     $(IMPORTDIR)/local-ssso.owl \
+				     $(IMPORTDIR)/local-sslso.owl \
+				     $(BRIDGEDIR)/uberon-bridge-to-sslso.owl \
 				     $(IMPORTDIR)/local-ceph.owl \
 				     $(IMPORTDIR)/local-cteno.owl \
 				     $(IMPORTDIR)/local-poro.owl
 
+COLLECTED_lifestages_SOURCES =       $(SUBSETDIR)/life-stages-minimal.owl \
```
matentzn, line 1106: I am not 100% sure about the line between life-stages-minimal.owl and local-sslso.owl, but if you are confident about it I am as well.

src/ontology/uberon.Makefile
```diff
@@ -1132,39 +1141,51 @@ collected-metazoan.owl: $(TMPDIR)/collected-metazoan.owl
...
 		                              -l UBERON=10 -l CL=9 \
 		                              -d UBERON=10 -d CL=9 \
 		 reason -r ELK --equivalent-classes-allowed all \
 		 relax \
 		 reduce -r ELK -o $@
 
+# Step 2a: The "tax-merges.tsv" file used in the rule above is automatically
+# derived from the list of species in config/taxa.yaml.
+$(TMPDIR)/tax-merges.tsv: $(SCRIPTSDIR)/taxa.py config/taxa.yaml
+	python3 $(SCRIPTSDIR)/taxa.py make-merge-table > $@
```
matentzn, line 1175: Optional: This is not a breaking comment but I have made the experience that to be able to read makefiles unobtrusively, it is nice if all file dependencies are passed explicitly to the python script. Else, oftencase, you have like dozens of file dependencies in the header of a python script which violates SRP when the file moves location, and also when a goal contains multiple processes (its then unclear which make dependency relates to which script).

matentzn (dismissed):
I have not reviewed src/scripts/taxa.py, and made a few optional comments for your own pleasure; non of it is binding.

src/ontology/bridge/collected-lifestages-hdr.owl
gouttegd: That is indeed an oversight, thanks.

src/ontology/uberon.Makefile
gouttegd: Err, `life-stages-minimal.owl` and `local-sslso.owl` have virtually nothing to do with each other.

`life-stages-minimal.owl` is the subset of Uberon that contains only Uberon’s life stage terms (UBERON:0000104, UBERON:0000105 and their descendants).

`local-sslso.owl` is our local mirror of the life stage ontologies repository, containing all the Bgee-maintained taxon-specific life stage terms (all the BtauDv, GgorDv, PpanDv, etc.).

The “line” between the two is clear: `life-stages-minimal.owl` is a **Uberon** subset that only contains Uberon terms; `local-sslso.owl` is a product of the life stage ontologies repositories that only contains terms from that repo.

I guess the confusion might stem from the fact that the Uberon subset is itself called **life-stages**-minimal (which might give the impression it is a product from the life stage ontologies rather than a product of Uberon); this is unfortunate, but this has always been the name of that subset, I didn’t change it.

src/ontology/bridge/collected-lifestages-hdr.owl
matentzn: Ha sorry

src/ontology/uberon.Makefile
matentzn: Ha sorry

src/ontology/uberon.Makefile
gouttegd: So you’d prefer something like

```
python3 $(SCRIPTSDIR)/taxa.py make-merge-table config/taxa.yaml > $@
```

?

No objection to that.

src/ontology/uberon.Makefile
gouttegd: Double-checking now, but it looks like XAO has indeed slipped through the cracks. It used to be provided by `ssso-merged`, so now we need to bring it ourselves like we do for, e.g., FBdv and WBls.

gouttegd: XAO is a tricky case because, contrary to FBdv, WBls, and other similar species-specific ontologies, XAO contains both anatomical terms and life-stage terms, so we need to make a special case for this one (forcefully extracting a life-stage-only version of XAO).

gouttegd: Done.

matentzn:
Looks great, thank you!

gouttegd:
I am still working on a (hopefully quick) last update: Moving the source of truth for the HsapDv and MmusDv mappings from Uberon to SSLSO, now that SSLSO provides a SSSOM mapping set.

Currently those mappings come _both_ from Uberon and from SSLSO, and there are already some discrepancies between the Uberon-maintained mappings and the SSLSO-maintained ones (with the Uberon mappings pointing to some HsapDv/MmusDv terms that have been obsoleted). Better to let SSLSO be the sole source of truth.

Then it’s only a matter of updating the docs. Overall I plan for everything to be done and over in time for the next release that @aleixpuigb  is planning for early next week.

matentzn (approved):
Awesssssome:

I reviewed

https://github.com/obophenotype/uberon/pull/3443/commits/832f254a57656c1458876da5692dcbe52d2c252b (looks great)

and 

https://github.com/obophenotype/uberon/pull/3443/commits/d57e4d582f3f7020ba5da9b6315e08095a737033

Which is _super_great!

All good to me!

PR #3426, opened 2024-11-14
Replace `supplies` and `drains` relations by more specific relations

gouttegd:
RO now provides '[vessel supplies blood to](http://purl.obolibrary.org/obo/RO_0020101)' (RO:0020101) as a more specific variant of '[supplies](http://purl.obolibrary.org/obo/RO_0002178)' (RO:0002178). The new relation is domain-restricted to '[blood vessel](http://purl.obolibrary.org/obo/UBERON_0001981)' (UBERON:0001981). In this PR, we use the new relation wherever '[supplies](http://purl.obolibrary.org/obo/RO_0002178)' was used to qualify a blood vessel.

Likewise for '[vessel drains blood from](http://purl.obolibrary.org/obo/RO_0020102)' (RO:0020102), which is a more specific variant of '[drains](http://purl.obolibrary.org/obo/RO_0002179)' (RO:0002179).

Recently introduced patterns for arteries and veins (in #3288) already use the more specific relations, so this improves Uberon’s consistency. More importantly, no longer using '[supplies](http://purl.obolibrary.org/obo/RO_0002178)' and '[drains](http://purl.obolibrary.org/obo/RO_0002179)' wherever a blood vessel is involved means that we no longer need to inject our own domain restrictions to those relations, so we can remove those injections as well (#2453).

The new relations are range-restricted in RO to (material) '[anatomical structure](http://purl.obolibrary.org/obo/UBERON_0000061)', which causes two side-effects:

* '[supraorbital artery](http://purl.obolibrary.org/obo/UBERON_0015160)' (UBERON:0015160) is unsatisfiable because it '[supplies blood to](http://purl.obolibrary.org/obo/RO_0020101)' the '[frontal sinus](http://purl.obolibrary.org/obo/UBERON_0001760)' (UBERON:0001760), which is an immaterial entity. We fix that by changing the target of the relation to '[mucosa of frontal sinus](http://purl.obolibrary.org/obo/UBERON_0005027)' (UBERON:0005027).

[...]

Issue #3288, opened 2024-06-04
Blood vasculature design pattern - VCCF


aleixpuigb:
To support the addition of new terms from the [HRA VCCF](https://github.com/hubmapconsortium/hra-vccf), we need a new design pattern. The following logical axioms will be added to the blood vessels:
- SubclassOf
- part of
- connecting branch of (mainly for arteries/arterioles)
- tributary of (for veins/venules)
- vessel supplies blood to (mainly arteries/arterioles)
- vessel drains blood from (mainly for veins/venules)

To test the design pattern, [55 blood vessels](https://docs.google.com/spreadsheets/d/1QrXvmsiCPqFQy6ZMHJQiE2rySrhmbQ1TE7GgDsg9ZFk/edit#gid=2033210470) that are missing in Uberon, will be added. Some of them are from FMA, and for those, the FMA term label will be added as an exact synonym and the FMA term ID as a dbxref.

PR #3450, opened 2024-12-16
Replace supplies and drains relations by more specific relations

gouttegd:
(This PR supersedes #3426; it is almost identical to it, except that the fix to the unsatisfiability of “intercostal vein“ and “intercostal artery” is to redefine those terms to say they drain/supply blood from/to the intercostal muscles, instead of introducing a new term like “intercostal element“.)

RO now provides '[vessel supplies blood to](http://purl.obolibrary.org/obo/RO_0020101)' (RO:0020101) as a more specific variant of '[supplies](http://purl.obolibrary.org/obo/RO_0002178)' (RO:0002178). The new relation is domain-restricted to '[blood vessel](http://purl.obolibrary.org/obo/UBERON_0001981)' (UBERON:0001981). In this PR, we use the new relation wherever '[supplies](http://purl.obolibrary.org/obo/RO_0002178)' was used to qualify a blood vessel.

Likewise for '[vessel drains blood from](http://purl.obolibrary.org/obo/RO_0020102)' (RO:0020102), which is a more specific variant of '[drains](http://purl.obolibrary.org/obo/RO_0002179)' (RO:0002179).

Recently introduced patterns for arteries and veins (in https://github.com/obophenotype/uberon/issues/3288) already use the more specific relations, so this improves Uberon’s consistency. More importantly, no longer using '[supplies](http://purl.obolibrary.org/obo/RO_0002178)' and '[drains](http://purl.obolibrary.org/obo/RO_0002179)' wherever a blood vessel is involved means that we no longer need to inject our own domain restrictions to those relations, so we can remove those injections as well (https://github.com/obophenotype/uberon/issues/2453).

The new relations are range-restricted in RO to (material) '[anatomical structure](http://purl.obolibrary.org/obo/UBERON_0000061)', which causes two side-effects:

* '[supraorbital artery](http://purl.obolibrary.org/obo/UBERON_0015160)' (UBERON:0015160) is unsatisfiable because it '[supplies blood to](http://purl.obolibrary.org/obo/RO_0020101)' the '[frontal sinus](http://purl.obolibrary.org/obo/UBERON_0001760)' (UBERON:0001760), which is an immaterial entity. We fix that by changing the target of the relation to '[mucosa of frontal sinus](http://purl.obolibrary.org/obo/UBERON_0005027)' (UBERON:0005027).

* '[intercostal vein](http://purl.obolibrary.org/obo/UBERON_0012197)' (UBERON:0012197) and '[intercostal artery](http://purl.obolibrary.org/obo/UBERON_0005612)' (UBERON:0005612) are unsatisfiable because they drain and supply blood from and to the '[intercostal space](http://purl.obolibrary.org/obo/UBERON_0012198)' (UBERON:0012198), an immaterial entity. As discussed in the Uberon call of December 16th, 2024, we fix that by redefining those terms as draining/supplying blood from/to '[intercostal muscle](http://purl.obolibrary.org/obo/UBERON_0001111)' instead of '[intercostal space](http://purl.obolibrary.org/obo/UBERON_0012198)'. This causes a slight loss of information (there are more things in the intercostal space than just the intercostal muscles) but this was deemed (by Uberon editors in the aforementioned call) to be good enough.


gouttegd:
Cancelled in favour of the original proposal in #3426.

PR #3461, opened 2025-01-13
Refresh imports for Uberon release

aleixpuigb:


balhoff:
The error above is almost exactly the example I used in the taxon constraints explainer! 😁 https://oboacademy.github.io/obook/explanation/taxon-constraints-explainer/#using-taxon-restrictions-for-quality-control

aleixpuigb:
@emquardokus this conflict is caused by refreshing the HRA subset. The addition of [arrector pili muscle of vibrissa](http://purl.obolibrary.org/obo/UBERON_0011940) in one of the ASCT+B tables results in the addition of `'present in taxon' some 'Homo sapiens'`:

image

This term is part_of some [strand of vibrissa hair](http://purl.obolibrary.org/obo/UBERON_0006378) that contains `'never in taxon' some Homo sapiens'.

As we need an Uberon release and fixing this will take some time (it needs to be fixed externally to Uberon), I think the best course of action for now is close this PR and refresh the imports in a new one without refreshing the HRA subset.

cc @gouttegd @dosumis

dosumis:
Pretty clearly an error in the ASCT+B tables (unless someone has discovered that humans have whiskers)

gouttegd:
@aleixpuigb Since it seems that it is clearly an error in the HRA subset, what you can also do is simply edit the refreshed HRA subset file to remove the offending axiom.

We normally don't recommend this kind of fix (it's only a band-aid as the real fix must happen upstream), but this is an instance where it is actually justified.

gouttegd (approved):
Looks all good to me.

src/ontology/imports/merged_import.owl
```diff
@@ -153265,30 +154006,20 @@ AnnotationAssertion(rdfs:comment <http://purl.obolibrary.org/obo/PR_000050216> "
 AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/PR_000050216> "receptor-type tyrosine-protein phosphatase C isoform CD45R")
 SubClassOf(<http://purl.obolibrary.org/obo/PR_000050216> <http://purl.obolibrary.org/obo/PR_000001006>)
 
-# Class: <http://purl.obolibrary.org/obo/PR_000050567> (obsolete protein-containing material entity)
+# Class: <http://purl.obolibrary.org/obo/PR_000050567> (protein-containing material entity)
```
aleixpuigb, line 154009: I am conducting the release (I am reviewing the diff file) and realised that many terms were adding SubclassOf [PR_000050567](http://purl.obolibrary.org/obo/PR_000050567). I am not sure why it has reverted this obsoletion.

gouttegd: Last time this happened it came from the PR slim import. Might be worth checking if anything changed over there recently...

(In a train right now, can't do anything.)

aleixpuigb: Yes it comes from there, I will try to investigate
image
Thank you!

gouttegd: The non-obsoleted PR:000050567 was apparently re-introduced in the PR slim yesterday as part of [this commit](https://github.com/obophenotype/pro_obo_slim/commit/b7415d8d62b9fece2cadf4a59cdcef0c2de81f3e).

Not entirely sure how it could have happened, but my wild guess is that there might have been a glitch when the upstream `pr.owl.gz` file was being downloaded, which may have caused the mirror _not_ to be updated, so the slim may have been re-generated using an old mirror that dated from before PR:000050567 has been fixed.

gouttegd: If my wild guess is true, it’s an issue that needs to be addressed in the PR slim repository. The process of re-generating the slim must be made to be more robust to this kind of glitches.

gouttegd: For now, I’d suggest

(1) re-generating the PR slim after making sure you forcefully delete the local PR mirror (everything in the `mirror` directory);
(2) checking that PR:000050567 is obsoleted as it should in the re-generated slim;
(3) push the changes to the PR slim repo;
(4) refresh the imports in Uberon once again.

aleixpuigb: Thank you for looking into this Damien.

PR #3465, opened 2025-01-22
Make sure the SSSOM-derived cross-reference use correct OBO prefixes.

gouttegd:
Most of the prefixes used by the taxon-specific life stage ontologies (SSLSO) are not intrinsically known by ROBOT (prefixes such as BtauDv, GgalDv, PpanDv, etc.).

As a result, when the mappings from SSLSO are transformed into cross-references for injection into the ontology (when we create the `mappings.owl` component), the SSLSO term IDs are shortened using the only suitable prefix that ROBOT knows about, the `obo:` prefix.

This means that a mapping to, say, <http://purl.obolibrary.org/obo/BtauDv_0000078> gets turned into a cross-reference to `obo:BtauDv_0000078` -- which is, strictly speaking, correct, but _not_ the format most people would expect for OBO-style cross-references (which would be `BtauDv:0000078`).

The fix is simply to make sure the prefixes are correctly declared.


matentzn (approved):
Great, thank you!

PR #3466, opened 2025-01-23
Update bridging relations in cross-species bridges

### LINKED ISSUE #3378: Add taxon constraints to imported species-specific terms in Composite-Metazoan?
Uberon itself makes use of taxon constraints to express the fact that we know that some terms are applicable/not applicable to a given taxon.

However, AFAIK most (if not all) of the taxon-specific ontologies that we merge to build Composite-Metazoan do _not_ make use of taxon constraints. For example, none of the terms in FBbt are formally restricted to NCBITaxon:7227 (_D. melanogaster_) -- this is deemed unnecessary since the entirety of FBbt, by definition, is intended to apply to only that taxon.

This means that in Composite-Metazoan, we have a mixed situation where the “core” terms (coming from Uberon itself) may have taxon constraints, but the taxon-specific terms (imported from the taxon-specific ontologies), somewhat paradoxically, do not have any constraints (apart from the ones they may inherit from the core terms, but those would be less strict than they could be).

For example, in Uberon `circulatory system dorsal vessel` (UBERON:0015231) is restricted to _Arthropoda_; therefore, all FBbt terms attached to that term (such as `embryonic/larval dorsal vessel (Drosophila)` (FBbt:00001673)) inherit that restriction, but they are _not_ formally restricted further to _D. melanogaster_ only, even though we know (by the mere fact that they are FBbt terms) that they are intended to apply only to _D. melanogaster_.

So maybe we could update the bridge files that connect Uberon terms with taxon-specific terms to automatically inject appropriate `in_taxon` relationships to the taxon-specific terms?

### LINKED ISSUE #3409: What relation should link a life stage term to its taxon-specific counterpart?
(This question was first raised in the [developmental stage ontologies](https://github.com/obophenotype/developmental-stage-ontologies/pull/102#discussion_r1799308335) repo, but as a side remark on a PR. It deserves its own discussion.)

Part of the steps that are needed to build Composite Metazoan is to create “bridging axioms” that link the taxon-neutral terms in Uberon to their taxon-specific counterparts in the the taxon-specific ontologies (FBbt, FBdv, WBbt, XAO, etc.).

For almost all terms, that link is of the form:

> {taxon-specific term} **EquivalentTo:** {Uberon term} _and_ ('part of' _some_ {taxon ID})

But for the terms that represent life stages (specifically '[life cycle](http://purl.obolibrary.org/obo/UBERON_0000104)' and '[life cycle stage](http://purl.obolibrary.org/obo/UBERON_0000105)' and their descendants), the links can use two distinct relations depending on where the bridging axioms are generated.

For the bridging axioms that are generated in the [life stage ontologies](https://github.com/obophenotype/developmental-stage-ontologies/) repository, they are of the same form as above (in particular, they use the same `part of` relation).

But for the bridging axioms that are generated in Uberon (typically, those are the bridges that link to externally maintained, taxon-specific life stage ontologies like FBdv or WBls), the links are of the form

> {taxon-specific term} **EquivalentTo:** {Uberon term} _and ('occurs in' _some_ {taxon ID})

where `occurs in` is [BFO:0000066](http://purl.obolibrary.org/obo/BFO_0000066).

It cannot be good that the relations used in those bridging axioms vary depending on the species (e.g. `occurs in` for _Drosophila melanogaster_, `part of` for _Gorilla gorilla_). Whatever relation is used, it should be the same for all species.

gouttegd:
This PR changes the axioms that make up the cross-species bridge ontologies.

Currently, foreign (taxon-specific) terms are bridged to their Uberon counterparts by a single equivalence axiom that looks like this:

> FBbt:00000002 EquivalentTo: UBERON:6000002 and ('part of' some NCBITaxon:7227)

or like this (if the term represents an occurrent, such as a developmental stage)

> FBdv:00007012 EquivalentTo: UBERON:0000105 and ('occurs in' some NCBITaxon:7227)

However, this has only been intended as a _temporary_ situation. The [original intention]( https://github.com/obophenotype/uberon/issues/2428#issuecomment-1103321518) was for each foreign term to be bridged by _a pair_ of axioms as follows:

> FBbt:00000002 EquivalentTo: UBERON:6000002 and ('in taxon' some NCBITaxon:7227)
> FBbt:00000002 SubClassOf: 'part of' some NCBITaxon:7227

if the term represents a continuant, or

> FBbt:00007012 EquivalentTo: UBERON:0000105 and ('in taxon' some NCBITaxon:7227)
> FBbt:00007012 SubClassOf: 'occurs in' some NCBITaxon:7227

if the term represents an occurrent.

The two-axioms form had been left aside in favour of the single-axiom form solely for technical reasons, because the bridging pipeline at the time was too difficult to update to accommodate for such a change. This is no longer true, so this PR now implements the two-axioms form.

The Composite Metazoan pipeline is updated accordingly to “unfold“ over 'in taxon' rather than over 'part of' or 'occurs in'.

closes #3409 
closes #3378


matentzn:
This needs a @cmungall glance, I am not a good reviewer for this; I can help find someone that is.

gouttegd:
@matentzn I did explicitly ask @cmungall for a review; you were automatically asked for one only because you are a “code owner”.

But again, this merely (and finally!) implements what was asked more than two years ago.

cmungall:
The PR makes complete logical sense, and this was indeed the original
intent.

However, representing this unilaterally in Uberon could cause confusion.
Queries that work for UBERON may fail for GO or CL.

But even assuming we can move forward in unison, I now think there is value
in using different relations for different use cases. Granted this is not
the underlying basis for RO, so there are valid cases either way. Would be
good to gather opinions for others. I can expand on with some examples
later.

One of my concerns is increasing raggedness in the part-of lattice. User
sees 'flexor pollicis longus muscle' as a part of Human in OLS. They open
up Human in OLS expecting to see all human parts. They get confused when
there is just some tiny collection of terms.

gouttegd:
> Queries that work for UBERON may fail for GO or CL.

Not sure I follow you. The relations updated by this PR would not go into UBERON itself -- they would only go into the bridges (and from there into the collected/composite products).

Or are you referring to queries on Ubergraph (which does use the Uberon-provided bridges)? If so, note that this PR also updates the CL bridges - since they are maintained here in Uberon.

> But even assuming we can move forward in unison, I now think there is value
in using different relations for different use cases.

I don’t think it is realistic for Uberon to provide several versions of the bridges or of the collected/composite ontologies, each using slightly different axioms to accommodate for potentially different use cases.

If people have use cases that require different bridging axioms that the one we provide in the pre-built bridges, they now have the possibility of downloading the Uberon SSSOM mapping set and generate their own bridges.

cmungall:
On Fri, Jan 24, 2025 at 9:28 AM Damien Goutte-Gattat <
***@***.***> wrote:

> Queries that work for UBERON may fail for GO or CL.
>
> Not sure I follow you. The relations updated by this PR would not go into
> UBERON itself -- they would only go into the bridges (and from there into
> the collected/composite products).
>
> Or are you referring to queries on Ubergraph (which does use the
> Uberon-provided bridges)? If so, note that this PR also updates the CL
> bridges - since they are maintained here in Uberon.
>
Good point, I forgot about CL

But the broader issue is we give users the expectations that they can
traverse anatomy to taxon via part-of in one anatomy/product, they might
reasonably expect they can do it in another (e.g GO-CC)


> But even assuming we can move forward in unison, I now think there is value
> in using different relations for different use cases.
>
> I don’t think it is realistic for Uberon to provide several versions of
> the bridges or of the collected/composite ontologies, each using slightly
> different axioms to accommodate for potentially different use cases.
>
> If people have use cases that require different bridging axioms that the
> one we provide in the pre-built bridges, they now have the possibility of
> downloading the Uberon SSSOM mapping set and generate their own bridges.
>

Agreed, and sorry I wasn't clear. I didn't mean to suggest different
versions, that would be complicated.

I just meant that I think it's fine if in-taxon is the anatomy->taxon
relation in all cases, and part-of is anatomy->anatomy. The two kinds of
traversals represent different use cases. Even if from a BFO/COB/OBO
Realist perspective NCBITaxon is an anatomy ontology (just one that
subclasses the 'whole organism' class)

gouttegd:
> Agreed, and sorry I wasn't clear. I didn't mean to suggest different
versions, that would be complicated.
> 
> I just meant that I think it's fine if in-taxon is the anatomy->taxon
relation in all cases, and part-of is anatomy->anatomy. The two kinds of
traversals represent different use cases. 

OK, it’s clearer now, thanks.

But then I’m afraid I still don’t understand your concern with this PR.

Currently (without this PR), in-taxon is not used in the bridge files at all. All links between a Uberon term and its taxon-specific counterpart use part-of (or occurs-in, for terms representing developmental stages). So, part-of is both for anatomy-to-anatomy (e.g. from cell type to tissue, from tissue to organ, etc.) _and_ for anatomy-to-taxon (e.g., FBbt’s ovary is a Uberon ovary that is part-of some _Drosophila melanogaster_).

The point of this PR is precisely to no longer use part-of (or, again, occurs-in for developmental stages) for anatomy-to-taxon (in cross-species bridging equivalence axioms), but only in-taxon (e.g., FBbt’s ovary is a Uberon ovary that is in-taxon some _Drosophila melanogaster_) -- to me this seems closer to what you actually want, compared to the current situation.

cmungall:
AHA it all makes sense now, thanks for your patience, I should have read more closely. Approving now!

cmungall (approved):

PR #3469, opened 2025-01-29
Integrate ChirDv into the composite pipeline.

gouttegd:
Now that SSLSO includes the ChirDv (goat life stages) component, we must make sure the bridging/composite pipeline is ready to deal with that component: that is, we must generate bridging axioms between the Uberon life stages terms and their ChirDv counterparts, and unfold them into class expressions when building Composite Metazoan.

Otherwise, the ChirDv terms would still end up in the collected/composite products (since they are bundled with SSLSO), but would not be connected to Uberon’s hierarchy.

### DIFF
--- a/src/ontology/config/taxa.yaml
+++ b/src/ontology/config/taxa.yaml
@@ -223,3 +223,8 @@ species:
     bridging:
       - prefix: TnigDv
         name: sslso
+  - taxon_id: NCBITaxon:9925
+    label: goat
+    bridging:
+      - prefix: ChirDv
+        name: sslso


matentzn (approved):
🎉 wow... we dare to... Add a new species! That must be the first in a looong time :D GREAT thanks for taking the initiative!

gouttegd:
Well, this is more or less imposed by SSLSO. Whether we want a new species in Composite Metazoan or not, goats are now represented in SSLSO, and since SSLSO is an integral part of Composite Metazoan, goat terms _will_ end up in Composite Metazoan anyway. So better to make sure they are properly integrated into Uberon’s hierarchy than to let them float around unattached to anything.

(The alternative would be to forcibly strip ChirDv terms out of SSLSO prior to merging SSLSO in, but that would actually be _more_ work than properly integrating those terms.)

PR #3477, opened 2025-02-12
remove incorrct subClassOf axiom on thoracic drg

tgbugs:
UBERON:0002835 ! thoracic dorsal root ganglion is not a subClassOf UBERON:0000961 ! thoracic ganglion because a thoracic dorsal root ganglion is not a paravertebral ganglion

closes https://github.com/obophenotype/uberon/issues/3475


tgbugs:
@cmungall it doesn't look like the incorrect asserted subClassOf (`UBERON:0000961 ! thoracic ganglion`) came from neurolex or from NIFSTD, or maybe it did and it was corrected on our end at some point. It is present in the first commit 621c5a6285866779e00c05f54ab4304b74f42552 so can't say more than that.

PR #3496, opened 2025-03-24
Include seeAlso annotations in mapping-derived xrefs.

gouttegd:
When generating the cross-references from the SSSOM-maintained mappings, annotate them with any seeAlso annotations that may be associated with the mappings.

The intention is that, if a mapping editor has a reference to back up a mapping statement, that reference can be stored in the SSSOM `see_also` field, and once there it should be preserved when the mapping is transformed into a classic cross-reference.

(This is the same thing as the [corresponding CL PR](https://github.com/obophenotype/cell-ontology/pull/3022)).


matentzn (approved):
Looks good to me!

PR #3511, opened 2025-04-24
Fix issue #3003: Updated cardiac septum definition

### LINKED ISSUE #3003: review definition of cardiac septum and its child terms
**Uberon term**
cardiac septum (UBERON:0002099)

current definition is too narrow for the child terms 

The definition is  "The thin membranous structure between the two heart atria or the thick muscular structure between the two heart ventricles" but child terms include septa between the atria and ventricles and outflow tract septa.

**Suggested revision of textual definition**
Revised definition assuming the child terms are correct (that is this term is intended to cover all septa between parts of the heart not just those between the atria or ventricles)

"The thin membranous structure between parts of the heart, including the atria, ventricles, and outflow tract."

cmungall:
## Summary
- Updated definition of cardiac septum (UBERON:0002099) to include all septa between parts of the heart, specifically including the outflow tract and better accommodating all child terms
- The previous definition was too narrow and only mentioned atria and ventricles, but child terms include atrioventricular septum and outflow tract septum

Fixes #3003

Signed-off-by: @dragon-ai-agent


cmungall:
This is the real @cmungall here,We're teaching @dragon-ai-agent how to do this. I just want to note that it did actually check the existing children as @sbello suggested. We're working on ways to have it explain it's working more.

Here is the full claude code headless trace: https://gist.github.com/cmungall/5dec10f6d8ce35d4b41099a30ae9db43

gouttegd:
I would _strongly_ suggest that your AI-based workflow includes a final _re-serialisation step_ in which you run the -edit file through `robot convert` or any other tool that would ensure the file is serialised exactly like Robot or Protégé would serialise it.

Otherwise the serialisation glitches caused by your tool are going to be automatically fixed the next time someones makes a (unrelated) edit in Protégé, which is going to be causing a lot of spurious diffs all over the place.

Issue #3512, opened 2025-04-29
missing insect mesothoracic leg, and insect metathoracic leg


ANiknejad:
- insect mesothoracic leg

- insect metathoracic leg

while it exist UBERON:6004663 ! insect prothoracic leg (synonym: "first leg" EXACT [], synonym: "T1 leg" EXACT []) , both T2 and T3 legs are not present in Uberon, as far as I can see, do you confirm @gouttegd ?

Proposition:
- insect mesothoracic leg
def: "Leg of the mesothoracic segment." []
synonym: "middle leg" EXACT []
synonym: "midleg" EXACT []
synonym: "second leg" EXACT []
synonym: "T2 leg" EXACT []
is_a: UBERON:0004121 ! ectoderm-derived structure
is_a: UBERON:0005895 ! insect leg
relationship: part_of UBERON:6003021 ! insect adult mesothoracic segment OR UBERON:6000018 ! insect mesothoracic segment ????
xref: FBbt:00004685

- insect metathoracic leg
def: "Leg of the metathoracic segment." []
synonym: "hindleg" EXACT []
synonym: "T3 leg" EXACT []
synonym: "third leg" EXACT []
is_a: UBERON:0004121 ! ectoderm-derived structure
is_a: UBERON:0005895 ! insect leg
relationship: part_of UBERON:6000019 ! insect metathoracic segment
xref: FBbt:00004707

Please note that it does not exist an UBERON 'insect adult metathoracic segment'

Kimball's Biology Pages
https://www.biology-pages.info/I/Insects.html

gouttegd:
> both T2 and T3 legs are not present in Uberon, as far as I can see, do you confirm @gouttegd ?

Yes.

And I agree that it does not make much sense for Uberon to have a term for the prothoracic leg specifically but not for the meso- and metathoracic legs.

It is questionable, however, whether Uberon needs a term for the leg of each segment at all. So rather than create the terms for the meso- and metathoracic legs, another option would be to, on the contrary, _obsolete_ Uberon’s “insect prothoracic leg” so that the “cut-off” point between Uberon and the insect-specific anatomy ontologies would be just below UBERON:0005895 (“insect leg”).

Given that “insect prothoracic leg” does already exist, though, I am mildly inclined to leave it as it is and to create the missing “insect mesothoracic leg“ and “insect metathoracic leg” terms as requested, though I’d welcome the opinion of other Uberon editors.

ANiknejad:
[...]

Issue #3376, opened 2024-10-03
Generalize all selected insect-X terms to arthropod


cmungall:
All UBERON:6X terms were pseudogeneralized from a subset of FBbt deemed generally useful (primarily those reference in GO).

Many of these are in fact applicable to arthropods. At first I thought it might be easiest to first universally generalize these and then selectively restrict, but we have a bit of a mix, and it may be better to do this more carefully on an as needed basis.

I also think we should introduce a new TC annotation "applicable outside of", such that for any term labeled "arthropod X" we can say "exists some T where T is not insecta where this is applicable" and have the reasoner tell us if we have accidentally placed an arthropod term under an insect term.

But for now I think we just need to work through these selectively following the patterns below on an as-needed basis. These could later be yamlified. 

## Insect in label, logic says arthropod, term applies to arthropods

This is the most straightforward category:

```yaml
id: UBERON:6005096
name: insect stomatogastric nervous system
def: "A group of small, interconnected ganglia situated posterior to and between the two brain hemispheres and associated with the pharynx, esophagus and aorta." [FlyBase:FBrf0089570, PMID:12966498]
synonym: "stomodaeal nervous system" RELATED [FlyBase:FBrf0111704]
synonym: "stomodeal nervous system" RELATED [FlyBase:FBrf0111704]
is_a: UBERON:0011216 ! organ system subdivision
relationship: in_taxon NCBITaxon:6656 ! Arthropoda
relationship: part_of UBERON:0001017 ! central nervous system
```

The label says insect but neither the text def nor the logic restricts to insects. (and in fact the definition doesn't outright rule out non-arthropods). In this case we think generalizing to arthropods is a good idea.

I think it's safe to relabel this to arthropod.

## label says insect, logic is largely unrestricted, term should remain insect

```yaml
id: UBERON:6000006
name: insect head segment
def: "Any segment that is part of some insect head." [UBERON:cjm]
intersection_of: UBERON:0000914 ! organismal segment
intersection_of: part_of UBERON:0000033 ! head
relationship: never_in_taxon NCBITaxon:7742 ! Vertebrata <vertebrates>
```

It's not clear why this didn't get a TC. As it is, it's kind of odd.

when talking about head homology in arthropods we get into murky territory (let's not discuss in this issue). I think the most conservative thing to do is to add an insect TC, but we have to be careful we don't restrict any existentially dependent children (see proposal for new TC annotation above)

Action: add a TC to insect

[...]

PR #3514, opened 2025-04-30
Add insect meso- and metathoracic legs.

gouttegd:
This PR adds the terms requested in #3512 (“insect mesothoracic leg” and “insect metathoracic leg”).

Along the way, it also amends the existing terms “insect leg” and “insect prothoracic leg” to:

* remove the '[develops from](http://purl.obolibrary.org/obo/RO_0002202)' relationships with the terms for the imaginal discs, because it is not true for all insects that the legs develops from imaginal discs (for example, in _Lepidoptera_ the adult leg develops from the larval leg, which itself does not develop from any imaginal disc);
* make them part of stage-neutral '[insect thoracic segment](http://purl.obolibrary.org/obo/UBERON_6000016)' rather than the adult-specific '[insect adult thoracic segment](http://purl.obolibrary.org/obo/UBERON_6003019)', because legs may also exist in larvae;
* add an explicit taxon constraint (not strictly necessary but for consistency, as recommended in #3376).

**Completely unrelated**: the PR includes some completely unrelated formatting changes. This is the (expected) consequence of edits recently made by “Claude”, as explained [here](https://github.com/obophenotype/uberon/pull/3511#issuecomment-2826999537).


aleixpuigb:
#gogoeditdiff

aleixpuigb (approved):
Everything looks great!

Issue #3027, opened 2023-08-18
facial skeleton terms


helenamachado:
**Preferred term label:**
Nasal notch

**Synonyms**
nasoincisve notch

**Definition (free text, please give PubMed ID)**
Anatomical point of cranium between nasal process of premaxilla and nasal bone, at the anteriormost end of the suture.

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
'part of' some premaxilla
'part of' some 'nasal bone'

ref.: https://books.google.com/books?id=0xIbF8NHU5kC&pg=PA589&lpg=PA589&dq=Incisura+nasoincisiva&source=bl&ots=hHpRp-2hv3&sig=ACfU3U2ZNUJmK-0bE-Ndi7hvSdF6NsoPuA&hl=en&sa=X&ved=2ahUKEwjFhYnzyL6AAxVAODQIHUoeBswQ6AF6BAgoEAM#v=onepage&q=Incisura%20nasoincisiva&f=false
https://www.imaios.com/en/vet-anatomy/anatomical-structure/naso-incisive-notch-11073769892

-----------------------------------------------------

**Preferred term label:**
Staphylon

**Definition (free text, please give PubMed ID)**
anatomical point of cranium; the most aboral point of the horizontal part of the palate in the median plane

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
'anatomical point'
'part of' some 'palate bone'

ref.: ISBN: 9780873659505

-----------------------------------------------------

**Preferred term label:**
Facial crest

**Synonyms**
crista facialis

**Definition (free text, please give PubMed ID)**
 ridge or similar projection rising above the surface of the lateral side of the cranium that crosses the zygomatic bone, extends to the maxilla and end on the facial tuberble

**Parent term (use https://www.ebi.ac.uk/ols/ontologies/uberon)**
crest and 'part of' some 'facial skeleton'
'part of' some 'temporal process of zygomatic bone'
'part of' some maxilla 

**Your nano-attribution (ORCID)**
https://orcid.org/0000-0001-8415-4848

PR #3516, opened 2025-05-02
Issue 3027

meghalithic:
creating the following terms: facial crest, nasal notch, and staphylon


src/ontology/uberon-edit.obo
```diff
@@ -184,6 +184,7 @@ property_value: doap-SVNRepository "https://obo.svn.sourceforge.net/svnroot/obo/
...
...s(<http://purl.obolibrary.org/obo/UBERON_0007681>))\nDeclaration(Class(<http://purl.obolibrary.org/obo/UBERON_4200133>))\nDeclaration(Class(<http://purl.obolibrary.org/obo/UBERON_4200230>))\nDeclaration(Class(<http://purl.obolibrary.org/obo/UBERON_7500131>))\nDeclaration(Class(<http://purl.obolibrary.org/obo/UBERON_8500006>))\nDeclaration(ObjectProperty(<http://purl.obolibrary.org/obo/BFO_0000050>))\n\n############################\n#   Classes\n############################\n\n# Class: <http://purl.obolibrary.org/obo/UBERON_7500131> (<http://purl.obolibrary.org/obo/UBERON_7500131>)\n\nSubClassOf(<http://purl.obolibrary.org/obo/UBERON_7500131> ObjectIntersectionOf(<http://purl.obolibrary.org/obo/UBERON_4200133> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/BFO_0000050> <http://purl.obolibrary.org/obo/UBERON_0007681>)))\n\n# Class: <http://purl.obolibrary.org/obo/UBERON_8500006> (<http://purl.obolibrary.org/obo/UBERON_8500006>)\n\nSubClassOf(<http://purl.obolibrary.org/obo/UBERON_8500006> ObjectIntersectionOf(<http://purl.obolibrary.org/obo/UBERON_4200230> ObjectSomeValuesFrom(<http://purl.obolibrary.org/obo/BFO_0000050> <http://purl.obolibrary.org/obo/UBERON_0001474>)))\n\n\n)
```
gouttegd, line 187: You are trying to insert the following axiom:

> 'facial crest' **SubClassOf:** 'crest' **and** ('part of' **some** 'facial neural crest')

Such an axiom cannot be represented in “pure” OBO format, hence the introduction of this `owl-axioms` tag.

If you intended to say a 'facial crest' is both a 'crest' and something that is 'part of' some 'facial neural crest', you should do so in two independent **SubClassOf** axioms, there is no need for an ObjectIntersectionOf.

If you intended to say that any 'crest' that is 'part of' some 'facial neural crest' is a 'facial crest', then what you need is a **EquivalentTo** axiom, not a **SubClassOf** axiom. Such an axiom _is_ representable in “pure” OBO.

src/ontology/uberon-edit.obo
```diff
@@ -224249,6 +224260,34 @@ is_a: UBERON:0006983 ! anatomical point
...
+id: UBERON:8500007
+name: staphylon
+def: "Anatomical point of cranium; the most aboral point of the horizontal part of the palate in the median plane." [isbn:9780873659505]
+is_a: UBERON:0006983 ! anatomical point
+relationship: part_of UBERON:0012071 ! palate bone
+property_value: dc-contributor "https://orcid.org/0000-0001-8415-4848" xsd:string
+
+[Term]
+id: UBERON:8500008
+name: Facial crest
```
gouttegd, line 224282: What’s the difference with UBERON:7500131 (“facial crest”) introduced above? It looks like you created the term twice.

gouttegd (changes requested):
It looks like you inadvertently created two “facial crest” terms.

And the first one has logical axioms that can’t be represented in OBO.

src/ontology/uberon-edit.obo
meghalithic: @gouttegd do you have any guidance for when EquivalentTo should be used ('crest' and 'part of' some 'facial neural crest') or SubClaseOf ('crest', 'part of' some 'facial neural crest')? The facial crest is not a crest can pop up anywhere (as seems to be facial neural crest), it is specifically on the lateral side between the maxilla and zygomatic arch.

src/ontology/uberon-edit.obo
meghalithic: That is a mistake! I will deprecate UBERON:850008

src/ontology/uberon-edit.obo
gouttegd: There are some pages on the OBOOK about the issue (for example [here](https://oboacademy.github.io/obook/explanation/logical-axiomatization/#equivalent-class-logical-definitions) and [here](https://oboacademy.github.io/obook/explanation/subClassOf-vs-equivalentTo/)), though it’s nowhere near as complete as it should be IMHO.

But the most important rule of thumb, I believe, is: _When in doubt, use SubClassOf_. :) Whenever you are unsure which of SubClassOf or EquivalentTo is more appropriate, it can never be wrong to use SubClassOf (the worst that can happen is that you will make a statement that won’t be as “strong” as it should be, but that will be correct all the same).

gouttegd: However, independently of the choice between SubClassOf and EquivalentTo, there may also be a problem with the term involved: are you sure this 'facial crest' is 'part of' some 'facial neural crest'?

'facial neural crest' is a part of the neural crest, an embryonic structure. That does not seem to fit well with the text definition.

Of note, the original term requested had proposed a logical definition involving 'facial skeleton', not 'facial neural crest'.

src/ontology/uberon-edit.obo
```diff
@@ -224249,6 +224261,32 @@ is_a: UBERON:0006983 ! anatomical point
...
+id: UBERON:8500007
+name: staphylon
+def: "Anatomical point of cranium; the most aboral point of the horizontal part of the palate in the median plane." [isbn:9780873659505]
+is_a: UBERON:0006983 ! anatomical point
+relationship: part_of UBERON:0012071 ! palate bone
+property_value: dc-contributor "https://orcid.org/0000-0001-8415-4848" xsd:string
+
+[Term]
+id: UBERON:8500008
+name: DEPRECATED Facial crest
```
gouttegd, line 224285: Given that the term has never made it into a public release (it did not even make it into the master branch!), it would have been perfectly acceptable to simply delete it without leaving a trace.

gouttegd:
@meghalithic There are a lot of minor conflicts, likely due to the fact that the first commit of the PR was done nearly two years ago. Do you want me to resolve them for you?

src/ontology/uberon-edit.obo
```diff
@@ -223972,6 +223984,32 @@ is_a: UBERON:0006983 ! anatomical point
 relationship: dc-contributor https://orcid.org/0000-0001-8415-4848 ! Helena Machado
 relationship: part_of UBERON:0001697 ! orbit of skull
 
+[Term]
+id: UBERON:8500006
+name: nasal notch
+def: "Anatomical point of cranium between nasal process of premaxilla and nasal bone, at the anteriormost end of the suture." [isbn:9783830410690]
+synonym: "nasoincisve notch" EXACT []
```
gouttegd, line 223991:
```suggestion
synonym: "nasoincisive notch" EXACT []
```

src/ontology/uberon-edit.obo
```diff
@@ -223972,6 +223984,32 @@ is_a: UBERON:0006983 ! anatomical point
...
+intersection_of: UBERON:4200230 ! surface of bone
+intersection_of: part_of UBERON:0001474 ! bone element
```
gouttegd, line 223993: There must be a mistake here, as I don’t see how the “nasal notch” can be defined as a “surface of bone that is part of a bone element”.

In fact the [original new term request](https://github.com/obophenotype/uberon/issues/3027) did not suggest a logical definition for “nasal notch”, just the relationships

> 'part of' some premaxilla

and

> 'part of' some 'nasal bone'

Given the text definition (“Anatomical point of cranium…”), I’d suggest removing the logical definition and add a simple `is_a: UBERON:0006983 ! anatomical point`.

gouttegd (changes requested):
The conflicts have been fixed (which makes the changes much easier to review now that they are no longer interspersed with unrelated changes).

The logical definition of 'nasal notch' is problematic, see below.

src/ontology/uberon-edit.obo
gouttegd: Yes, SubClassOf, as you have now done, thanks!

src/ontology/uberon-edit.obo
```diff
@@ -928,7 +928,7 @@ intersection_of: has_quality PATO:0002236 ! aliform
 relationship: capable_of_part_of GO:0060361 ! flight
 relationship: in_taxon NCBITaxon:33213 ! Bilateria
 relationship: never_in_taxon NCBITaxon:118072 {source="bgee"} ! Coelacanthimorpha
-relationship: never_in_taxon NCBITaxon:186634 {source="bgee", seeAlso="Wikipedia:Flying_fish"} ! Otomorpha
+relationship: never_in_taxon NCBITaxon:186634 {seeAlso="Wikipedia:Flying_fish", source="bgee"} ! Otomorpha
```
gouttegd, line 931: Can you check that you are using the latest version of Protégé (5.6.5)?

Those changes (caused by the fact that the way qualifiers are sorted when being written has changed in recent OWLAPI versions) should not be here.

src/ontology/uberon-edit.obo
```diff
@@ -223972,6 +223984,31 @@ is_a: UBERON:0006983 ! anatomical point
...
+id: UBERON:8500007
+name: staphylon
+def: "Anatomical point of cranium; the most aboral point of the horizontal part of the palate in the median plane." [isbn:9780873659505]
+is_a: UBERON:0006983 ! anatomical point
+relationship: part_of UBERON:0012071 ! palate bone
+property_value: dc-contributor "https://orcid.org/0000-0001-8415-4848" xsd:string
+
+[Term]
+id: UBERON:8500008
+name: DEPRECATED Facial crest
```
gouttegd, line 224007: Please change that `DEPRECATED` to `obsolete`. Or remove that term entirely – as I said before, it is fine to remove it rather than keeping it as an obsolete term, since it was first added in this very PR.

gouttegd (changes requested):

src/ontology/uberon-edit.obo
meghalithic: oh shoot, I am not. If I re-download it will it fix the issues?

src/ontology/uberon-edit.obo
```diff
@@ -223972,6 +223984,31 @@ is_a: UBERON:0006983 ! anatomical point
...
+def: "Anatomical point of cranium; the most aboral point of the horizontal part of the palate in the median plane." [isbn:9780873659505]
+is_a: UBERON:0006983 ! anatomical point
+relationship: part_of UBERON:0012071 ! palate bone
+property_value: dc-contributor "https://orcid.org/0000-0001-8415-4848" xsd:string
+
+[Term]
+id: UBERON:8500008
+name: obsolete Facial crest
+comment: accidentally duplicated term
+relationship: seeAlso UBERON:7500131 ! facial crest
```
gouttegd, line 224009: This relationship should be removed. Obsolete terms should have no relationships at all.

gouttegd (changes requested):

src/ontology/uberon-edit.obo
matentzn: Agreed, I removed it.

PR #3518, opened 2025-05-06
Remove the dynamic XML catalog.

### LINKED ISSUE #3504: Check whether the dynamic catalog is still needed, and remove if not
Unless I am very much much mistaken, the “dynamic XML catalog” (`src/ontology/catalog-dynamic.xml`) is no longer used anywhere by anything.

The only place where that file is referenced is in the definition of the `OWLTOOLS_CAT_DYNAMIC` variable:

```Makefile
CATALOG_DYNAMIC =      catalog-dynamic.xml
OWLTOOLS_CAT_DYNAMIC=  OWLTOOLS_MEMORY=$(OWLTOOLS_MEMORY) owltools --catalog-xml $(CATALOG_DYNAMIC)
```

but that variable itself is no longer used since the refactoring of the custom Makefile in #3028.

gouttegd:
The "dynamic XML catalog" (catalog-dynamic.xml) is a remnant of some old, Owwltools-dependent workflows that have been replaced when the custom Makefile was overhauled in #3028.

This catalog is no longer needed by any existing workflow, so we can remove it along with the (also unused) variables that referred to it, the rule that triggered its generation, and the script that actually generated it.

closes #3504


matentzn (approved):
🧹

Issue #3495, opened 2025-03-18
epithelium and lamina propria for GI tract


jychien:
We are currently in the process of reviewing GI datasets in CELLxGENE. The mucosa is composed of the epithelium and lamina propria:

![Image](https://github.com/user-attachments/assets/afe5d9fc-bb04-4df8-8e45-f3fe736c30e6)

There current are UBERON terms that are used to describe these layers for specific portions of the GI track, examples:

- [jejunal epithelium](http://purl.obolibrary.org/obo/UBERON_0000400)
- [jejunum lamina propria](http://purl.obolibrary.org/obo/UBERON_8600034)

To enable consistency and ability to search for the specific layer along all portions of the large intestine, we were hoping to have new terms added:

- **Tissues that do not have an epithelium term**: ascending colon, descending colon, sigmoid colon and transverse colon
- **Tissues that do not have a lamina propria term**: ascending colon, descending colon, sigmoid colon, transverse colon, stomach, caecum, rectum

We also have some questions regarding the parent terms of existing terms. But wanted to start this ticket to see what your thoughts are on these new term requests. Much appreciated.

dosumis:
@dragon-ai-agent please add a set of epithelium terms, each defined as part of one of the following gut regions: ascending colon, descending colon, sigmoid colon and transverse colon.  All should have equivalent class definitions with 'intestinal epithelium' in the genus position.  All should have a textual definition following the pattern 'A epithelial layer that lines the {gut region}', where {gut region} is replaces with the name of the gut region it is part_of.

cmungall:
Hi @jychien, I've created a PR (#3541) with the epithelium terms for the colon regions as requested. I've added terms for ascending colon epithelium, transverse colon epithelium, descending colon epithelium, and sigmoid colon epithelium. Each term has an equivalent class definition with 'intestinal epithelium' in the genus position and textual definitions following the pattern 'An epithelial layer that lines the {gut region}'.

The PR is available at: https://github.com/obophenotype/uberon/pull/3541

@dragon-ai-agent

cmungall:
I've added a set of epithelium terms for the requested colon regions:

- epithelium of ascending colon (UBERON:7770001)
- epithelium of transverse colon (UBERON:7770002)
- epithelium of descending colon (UBERON:7770003)
- epithelium of sigmoid colon (UBERON:7770004)

Each term has:
- A definition following the pattern 'An epithelial layer that lines the {gut region}'
[...]

PR #3541, opened 2025-05-21
Add epithelium terms for colon regions

cmungall:
## Summary
- Added four new epithelium terms for colon regions: ascending, transverse, descending, and sigmoid colon epithelia
- Each term has an equivalent class definition with 'intestinal epithelium' in the genus position
- Each term has a textual definition following the pattern 'An epithelial layer that lines the {gut region}'
- Added appropriate synonyms for each term

Addresses #3495 - epithelium and lamina propria for GI tract

@dragon-ai-agent


dosumis:
#gogoeditdiff

PR #3558, opened 2025-06-11
Update the Uberon ROBOT plugin.

gouttegd:
The Uberon plugin has been moved from my personal GitHub space (https://github.com/gouttegd/) to the “obophenotype” organisation (https://github.com/obophenotype/). So in this PR we update the location we are downloading the plugin from, and get the latest version.


matentzn (approved):
Gr8

PR #3563, opened 2025-06-16
Add more automatic reviewers.

gouttegd:
This PR update the `reviewers.yml` workflow so that @cmungall is not the only person asked to review changes affecting logical definitions. As discussed in the Uberon editors call on June 16th, 2025, @dosumis, @aleixpuigb, and @Caroline-99 should also be automatically assigned as reviewers on PRs that include changes to logical definitions, so as to reduce the review bottleneck.

This cannot be done simply by adding more reviewers to the `requested_reviewers` API call though, because doing so would cause the action to fail if the author of the PR is _any_ of the reviewers that we want to assign.

Instead, we refactor the workflow so that the “Assign reviewer” step is in a distinct job, in which we call the `requested_reviewers` API endpoint as many times as we have reviewers to assign, so that if one of them happens to be the PR's author, only that specific call will fail.

### DIFF
--- a/.github/workflows/reviewers.yml
+++ b/.github/workflows/reviewers.yml
@@ -7,41 +7,51 @@ on:
       - src/ontology/uberon-edit.obo
 
 jobs:
-  assign-reviewer:
+  check:
     runs-on: ubuntu-latest
-
+    outputs:
+      intersection_of_found: ${{ steps.check_intersection_of.outputs.intersection_of_found }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
         with:
-          fetch-depth: 0  # Fetch the entire history for all branches
-      
+          fetch-depth: 0
+
       - name: Fetch base branch
         run: |
-          git fetch origin ${{ github.base_ref }}  # Fetch base branch explicitly
-      
+          git fetch origin ${{ github.base_ref }}
+
       - name: Check if equivalent class axiom was edited
         id: check_intersection_of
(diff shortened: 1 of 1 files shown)


.github/workflows/reviewers.yml
```diff
@@ -7,41 +7,51 @@ on:
...
 
-      - name: Block PR by requesting changes
-        if: env.intersection_of_found == 'true'
+
+  assign-reviewer:
+    needs: check
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        reviewer: [ "cmungall", "dosumis", "aleixpuigb", "Caroline-99" ]
```
matentzn, line 47: Not sure I agree with this set of reviewers - unless everyone has had special training in logical reasoning, this defeats the purpose of this specific check. I know @aleixpuigb and @Caroline-99 are both highly capable curators and engineers, but have they been trained to really understand the responsibility and dangers associated with changes affecting the logical definitions of high-level classes? If so, then consider my comment void - I am of course _not_ blocking this PR if they are aware and trained in this matter.

I would suggest to add @gouttegd @balhoff and myself to this list as we have at least some level of training here.

gouttegd: > Not sure I agree with this set of reviewers

The list was proposed by @dosumis in the Uberon call of June 16th, and approved there.

Happy to add you (and @balhoff if he doesn’t object) to the list.

PR #3565, opened 2025-06-18
Cross-reference 'anatomical structure' to FAO:0000001 'fungal structure'

jseager7:
I'm part of the team maintaining the [Pathogen-Host Interaction Phenotype Ontology](https://github.com/PHI-base/phipo/tree/master) (PHIPO). We were looking to use [design patterns](https://github.com/obophenotype/upheno/tree/master/src/patterns) from uPheno to describe fungal anatomical entity phenotypes, using terms from the [Fungal Anatomy Ontology](https://github.com/obophenotype/fungal-anatomy-ontology) (FAO).

However, uPheno's anatomical design patterns are currently restricted to UBERON:0001062 'anatomical entity' ([example here](https://github.com/obophenotype/upheno/blob/7ea2ee9e680b84f52e59cba8202282a9471dde92/src/patterns/dosdp-dev/abnormallyDecreasedNumberOfAnatomicalEntity.yaml)), and FAO isn't cross-referenced to 'anatomical entity', which prevents using these patterns for fungal anatomical entities.

This pull request adds an xref from UBERON:0001062 'anatomical entity' to FAO:0000001 'fungal structure', which should hopefully resolve this.

```
xref: FAO:0000001
```

@matentzn I'm not sure if anything else is needed, like adding FAO as an imported ontology to Uberon.

Related issues:
* https://github.com/obophenotype/upheno/issues/987
* https://github.com/obophenotype/fungal-anatomy-ontology/issues/52


gouttegd (dismissed):
No problem here, but two remarks:

(1) Please consider whether Uberon’s “anatomical entity” is really the most appropriate term to map FAO’s “fungal structure” to (see inline comment).

(2) Adding this cross-reference will _not_ resolve your issue about not being able to use uPheno’s design patterns. As @matentzn said in https://github.com/obophenotype/fungal-anatomy-ontology/issues/52, the cross-reference on Uberon’s side is merely for documentation purpose – it has no actual significance. What you really need to do is, **on your side**, to import `UBERON:0001062` and make your `FAO:0000001` a subclass of it.

jseager7:
> What you really need to do is, on your side, to import UBERON:0001062 and make your FAO:0000001 a subclass of it.

Thanks for the clarification. I'll be making pull requests to do this in FAO as part of this task.

> Given the name and definition of FAO:0000001 (“fungal structure“, “an anatomical structure that forms all or part of a fungus”), I wonder if it should be mapped with UBERON:0000061 (“anatomical structure“) rather than UBERON:0001062 (“anatomical entity”).

I don't have any objections to this because I agree that it looks like 'anatomical structure' is a better mapping. I'm happy to amend the pull request if that's what's needed.

To explain: I think I chose 'anatomical entity' because I assumed that's what we'd need to subclass in FAO to 'enable' (so to speak) the uPheno pattern correctly, but if the cross-reference in Uberon is only for documentation purposes, like you said, then I'm happy for it for cross-reference to 'anatomical structure' instead.

jseager7:
Since there were no objections to cross-referencing to 'anatomical structure' I've gone ahead and done it.

@gouttegd I'm happy for this to be merged if you are.

PR #3566, opened 2025-06-19
Added new vasculature terms-4

ar-ibrahim:
- Related to issue https://github.com/obophenotype/uberon/issues/3457

- Added 17 new terms


aleixpuigb:
#gogoeditdiff

aleixpuigb (changes requested):
Definitions are good. The partonomy needs to be revised since the vessels supply or drain the structure, but it's not part of it.
The format od the date should be revised.

aleixpuigb (approved):

Issue #3564, opened 2025-06-17
Automatic ID allocation for AI-created PRs


gouttegd:
## Introduction

The use of AI agents (such as Dragon AI) that automatically open pull requests to change the ontology creates a particular problem regarding the allocation of term IDs.

As soon as the agent has more than one PRs open at the same time, and if it creates new terms in those PRs, it runs into the risk of generating conflicting IDs across the PRs.

This is a problem that theoretically also exists for human editors, but for human editors it is mitigated in two ways:

1) All editors have their own ID range, so even if two editors create each a PR at around the same, they cannot cause ID conflicts.

2) Protégé (used by most if not all human editors) can keep track of used IDs independently of Git (so, even across PRs), so even if one single editor is working on different PRs simultaneously, conflicts should still be avoided.

With an AI agent, both those mitigations are unavailable: all PRs from an agent will use the same ID range, and all invocations of the agent happen in their own context (their own branch), where the agent is not aware (cannot be aware) of what is happening in the other contexts.

For now, people using the agent have been working around the issue by basically keeping track _themselves_ of which IDs have been used in presently open PRs, and explicitly telling the agent not to use these IDs ([example here](https://github.com/obophenotype/uberon/issues/3495#issuecomment-2896204404)). Needless to say, this is hardly a satisfying solution.

So, in this ticket I want to discuss a better solution.

## Principle

The basic idea is the one proposed by @balhoff in [this ROBOT ticket](https://github.com/ontodev/robot/issues/763).

1) Let the agent create new terms using _temporary_ IDs picked in a _distinguishable_ namespace.

2) Once the PR is ready (all checks passed, all reviewers happy with the changes, etc.), before it is merged (or just after it has been merged; see later for a discussion about the options), temporary IDs are automatically replaced by newly minted _definitive_ IDs.

An implementation of the proposed `mint` command is available in the KGCL plugin for ROBOT, which is itself available in the ODK.

[...]

PR #3567, opened 2025-06-20
Support temporary IDs

gouttegd:
This PR implements the idea proposed in #3564 for allowing AI agents (or in fact any editor, AI or not) to use “temporary” term IDs that are later replaced by definitive IDs.

It allocates two new ID ranges:

* the `Temporary IDs` range (UBERON:99NNNNN): any ID picked within that range is treated as a temporary ID, slated for later replacement by a definitive ID;
* the `Automation` range (UBERON:12NNNNN): this is a “normal” ID range (IDs picked within that range are not treated differently from any other ID), but is reserved for automated process – notably, for the process that replaces temporary IDs by definitive ones.

It also adds a new GitHub Action workflow, `allocate-definitive-ids.yml`, which finds all temporary IDs (all IDs in the `UBERON:99NNNNN` range) in the -edit file and replaces them by newly allocated IDs from the `Automation` range. That workflow can be used in two different ways, corresponding to the two options outlined at the end of the #3564 ticket:

(A) Manual trigger: When a PR contains temporary IDs, a Uberon maintainer should manually trigger the `allocate-definitive-ids.yml` workflow on the PR’s underlying branch, _just before_ merging the PR into the master branch. This will automatically add a commit to the PR, in which temporary IDs are replaced by definitive ones. The PR should then be merged _as soon as possible_ once that commit has been added.

(B) Automatic trigger: If a PR containing temporary IDs is merged into the master branch (with the temporary IDs still present), the `allocate-definitive-ids.yml` workflow will automatically be triggered and push a new commit to replace the temporary IDs by definitive ones, directly on the master branch. This can act as a “fall-back” action to ensure that temporary IDs are never left in the master branch, in case a Uberon maintainer forgets to manually trigger the ID replacement workflow on a PR.

### DIFF
--- a/src/ontology/uberon-idranges.owl
+++ b/src/ontology/uberon-idranges.owl
@@ -270,3 +270,15 @@ Datatype: idrange:41
         allocatedto: "Arwa Ibrahim"
     EquivalentTo:
         xsd:integer[>= 8920000, < 8930000]
+
+Datatype: idrange:42
+    Annotations:
+        allocatedto: "Temporary IDs"
+    EquivalentTo:
+        xsd:integer[>= 9900000, < 10000000]
+
+Datatype: idrange:43
+    Annotations:
+        allocatedto: "Automation"
+    EquivalentTo:
+        xsd:integer[>= 1200000, < 1300000]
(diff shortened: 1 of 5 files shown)


src/ontology/uberon-idranges.owl
```diff
@@ -17,260 +16,269 @@ AnnotationProperty: allocatedto:
 
 AnnotationProperty: idprefix:
 
-AnnotationProperty: iddigits: 
+AnnotationProperty: iddigits:
```
gouttegd, line 19: Reviewers: all the changes in the `uberon-idranges.owl` files are due to the fact that the file has been re-serialized as a side-effect of the addition of the two new ranges. Those serialisation-induced changes are inconsequential.

cmungall (approved):
This is very well thought out

PR #3576, opened 2025-06-30
Use ODK 1.6 in diff.yml workflow.

### LINKED ISSUE #3575: GOGOEDITDIFF failing
Generating the diff file using the gogoeditdiff comment fails (example in issue https://github.com/obophenotype/uberon/pull/3573):
https://github.com/obophenotype/uberon/actions/runs/15970929389/job/45042014552
Not sure if it could be related to https://github.com/obophenotype/uberon/issues/3155 
@gouttegd @matentzn do you have an idea why does it fail? Thanks

gouttegd:
The `diff.yml` workflow, which powers the “gogoeditdiff” feature, is still using the ODK image version 1.5.4. This leads to a systematic failure, because at least one of the rules invoked by this workflow (`make uberon-base.owl`) is dependent upon the new ROBOT option `--include-subproperties`, which is not available in the version of ROBOT provided with ODK 1.5.4.

We update the workflow to ensure it uses version 1.6 of the ODK.

closes #3575

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -61,7 +61,7 @@ jobs:
     needs: [branch_status]
     if: ${{ github.event.issue.pull_request }}
     runs-on: ubuntu-latest
-    container: obolibrary/odkfull:v1.5.4
+    container: obolibrary/odkfull:v1.6
     steps:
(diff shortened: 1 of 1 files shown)


aleixpuigb (approved):
Thank you Damien!

PR #3577, opened 2025-07-02
Improve the check for changes to logical definitions.

### LINKED ISSUE #3568: Improved check for PRs that modify equivalence axioms
In #2020, it was requested that any change to a logical definition must be explicitly approved by some “senior” editor, typically @cmungall or @dosumis .

The way this was done, in #3425, is to detect if a given PR is adding/removing/modifying a `intersection_of:` line in `uberon-edit.obo`, and if so, to automatically assign @cmungall to review the PR.

However, this is not sustainable as @cmungall cannot afford to personally review every such PR.

Following several discussions both in Uberon meetings and on Slack, I propose an improved check that would work as follows:

(1) Determine how many logical definitions are added/removed/modified (_N_); if zero, nothing more to do.

(2) For every impacted class, determine: (i) the number of “removed subclasses” _Nr_ (classes that used to be subclasses of the impacted class, but no longer are under the modified logical definition), and conversely (ii) the number of “added subclasses” _Na_ (classes that were not previously subclasses of the impacted class, but now are under the modified logical definition).

(3) Post a comment on the PR that could look like this:

```
Number of modified logical definitions: 2

Impacted classes:

## Example Class 1 (UBERON:NNNNNNN)
2 removed inferred subclasses:
* Example Class 2 (UBERON:NNNNNNN)
* Example Class 3 (UBERON:NNNNNNN)

1 added inferred subclasses:
* Example Class 4 (UBERON:NNNNNNN)

## Example class 5 (UBERON:NNNNNNN)
7 added inferred subclasses:
* ...
```

(4) If either: (i) the number of modified logical definitions is higher than 10, or (ii) there is an impacted class for which the number of removed subclasses or the number of added classes is higher than 10, then @cmungall and other senior editors are assigned to review.

Overall, the idea is that the report posted in step 3 should be enough to give both the author of the PR and any reviewer some warning about the “hidden” consequences of the logical changes, without systematically looping @cmungall in. **But** if we detect that the changes are on a somewhat “large scale” (> 10 modified logical defs or > 10 changed inferences), then we do call the senior editors to have a closer look.

gouttegd:
This PR replaces the existing GA workflow that checks for changes to logical definitions by an improved check as envisioned in #3568.

The new check is “improved” in that:

* it relies on a ROBOT plugin to detect changes to logical definitions, rather than using `grep` to detect whether the diff contains `(-|+)intersection_of:` lines; this notably means that it does _not_ depend on the format of the `-edit` file (the check would work just as well if the `-edit` file was in any other format supported by the OWLAPI);
* it can additionally asses the “scale” of the changes (how many logical definitions are altered, and how many inferred _SubClassOf_ axioms are impacted).

If changes to logical definitions are detected, the new action will always post a report highlighting the consequences of the changes.

In addition, if the changes are “large” (more than 10 logical definitions changed, or at least one change that causes more than 10 inferred _SubClassOf_ axioms to change), the PR will be blocked (by requesting changes) and @cmungall and @dosumis will be assigned as reviewers.

closes #3568


gouttegd:
Note: the fact that the new check failed on this very PR is expected and normal: this is because in the base commit of this PR, the repository is still configured to use version 0.4.0 of the Uberon ROBOT plugin, which does not yet contain the new `inference-diff` command. That particular problem will automatically go away the moment this PR is merged.

PR #3580, opened 2025-07-04
adding copilot

cmungall:
See https://github.com/ai4curation/aidocs/issues/3


cmungall:
> Consider using the robot-version input 

I am not sure this works with copilot mode. TODO check

PR #3587, opened 2025-07-24
Replace obsolete SSSOM/T-OWL rules.

gouttegd:
The `mappings-to-xrefs.rules` SSSOM/T-OWL ruleset (used to generate the `mappings.owl` component, containing xref representations of foreign mappings) is still using some deprecated, pre-1.0 SSSOM/T-OWL features:

* the `check_object_existence()` preprocessing function;
* the `%subject_curie` placeholder.

Those features will no longer be supported in a future SSSOM-Java version and should be replaced by their SSSOM/T-OWL 1.0 equivalents, which is what we do here.

### DIFF
--- a/src/scripts/mappings-to-xrefs.rules
+++ b/src/scripts/mappings-to-xrefs.rules
@@ -30,12 +30,12 @@ subject==UBERON:* -> invert();
 !object==UBERON:* -> stop();
 
 # Ignore any mapping to an inexistent or obsolete UBERON class.
-predicate==* -> check_object_existence();
+!exists(%{object_id}) -> stop();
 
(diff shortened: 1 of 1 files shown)


gouttegd:
> what does the short do in "%{subject_id|short}"

It shortens (or “CURIE-fies”, if you will) the identifier.

`%{subject_id}` will inject the identifier itself (e.g. `http://purl.obolibrary.org/obo/FBbt_00005157`); `%{subject_id|short}` will inject the CURIE (`FBbt:00005157`). Since we are creating OBO-style cross-references here, we want the CURIE, not the full-length form.

PR #3598, opened 2025-08-15
Uberon release 2025-08-15

aleixpuigb:
The large size of changes in the ontology is caused by the removal of `hasOBONamespace](http://www.geneontology.org/formats/oboInOwl#hasOBONamespace) "uberon" ` #3584


aleixpuigb:
I can see merged PRs reflected in the diff dile

PR #3652, opened 2026-01-20
Move disjoint axioms from edit file to component

### LINKED ISSUE #3651: Newly introduced disjointness axioms cause OBO serialisation issue
PR #3374 introduced the two following `disjoint_from` tags:

```
[Term]
id: UBERON:0000001
disjoint_from: GO:0005623 ! cell
disjoint_from: GO:0110165 ! cellular anatomical entity
```

Nothing wrong with that addition _per se_, except that, whenever the edit file is written back by any OWLAPI-based application (e.g. Protégé or ROBOT), those tags will become:

```
[Term]
id: GO:0005623
disjoint_from: UBERON:0000001

[Term]
id: GO:0110165
disjoint_from: UBERON:0000001
```

Semantically, from a OWL point of view, this does not change anything: both forms represent the following disjointness axioms:

```
DisjointClasses(GO:0005623 UBERON:0000001)
DisjointClasses(GO:0110165 UBERON:0000001)
```

because the order of the operands of a _DisjointClasses_ axiom is not significant. And since it is not significant, there is no way for the OBO serializer to know that it should attach the corresponding tags to the `UBERON:0000001` term frame, rather than creating new term frames for `GO:0005623` and `GO:0110165` – when it has to translate `DisjointClasses(GO:0005623 UBERON:0000001)` into OBO, the OBO serializer simply adds the tag to the frame for the operand that comes first in lexicographical order (so in this case, `GO:0005623` rather than `UBERON:0000001`, since _G < U_).

But from a OBO point of view, the two forms are _not_ equivalent, because in the second form we now have two new term frames that do not have a label. And missing labels is an **error**.

Possible solutions:

(A) Finally admit that the OBO format is not worth the trouble, and switch the edit file to OWL Functional Syntax. I am mentioning that for completeness, I know there is no hope of that happening in my lifetime.

(B) Move the disjointness axioms to an separate component in OWL Functional Syntax.

(C) Add labels to the disjoint GO terms.

Since (A) will never happen, the second best solution is (B) in my opinion.

aleixpuigb:
Disjoint axioms have been moved from uberon-edit.obo to disjoint_union_over.owl
Fixes #3651


matentzn:
> Note that, to fix https://github.com/obophenotype/uberon/issues/3651, all that was needed was to move the disjointness axioms between a Uberon class and a non-Uberon class. Disjointness axioms between two Uberon classes do not pose any issue whatsoever, and could very well remain in the edit file.

@aleixpuigb how long would it take you do move the Uberon disjointness axioms back? My reasoning is here that we will continue to curate disjointness axioms in uberon with Protege, and it will be hard teaching people how if we move these into a separate file.

PR #3656, opened 2026-01-27
add artery and vein related terms requested by HuBMAP

zhengj2007:
issue: https://github.com/obophenotype/cell-ontology/issues/3471
Added 40 vein specific terms to template: artery_and_arteriole_pattern.tsv
Added 104 artery specific terms to template: vein_and_venule_pattern.tsv


zhengj2007:
I will make some edits in the template.

PR #3681, opened 2026-03-23
Add HOMBA Uberon bridge

ubyndr:
This PR adds `uberon-bridge-to-homba.owl` and registers it in the Uberon bridge configuration.

This is needed to support HOMBA, the Harmonized Ontology of Mammalian Brain Anatomy, in the Uberon bridge pipeline.

Changes:
- add `src/ontology/bridge/uberon-bridge-to-homba.owl` with HOMBA-Uberon bridge content
- register the bridge in `src/ontology/bridge/bridges.dispatch`
- add the bridge entry to `src/ontology/catalog-v001.xml`

This introduces the HOMBA bridge and the required wiring so it can be resolved and used by the existing bridge tooling.


src/ontology/bridge/bridges.dispatch
matentzn, line 1: bridges dispatch file is only for those bridges directly derived from the SSSOM file. Is that the case here?

src/ontology/bridge/uberon-bridge-to-homba.owl
matentzn, line 1: How is this file maintained? Manually? Is it sourced from elsewhere? There are some externally maintained breidges in uberon.Makefile - this would need to be integrated there.

matentzn (changes requested):

cmungall (commented):
more docs would be good. What is a .dispatch file?

Not registered
https://bioregistry.io/registry/homba

who is homba?

dont want to block important work so merge when ready but we need it to be more maintainable

dosumis:
We should definitely register in bioregistry.

Source is new structuregraph file from Allen.  They now have a link for this.  Hiw should we indicate source?

cmungall:
> Source is new structuregraph file from Allen. They now have a link for this. Hiw should we indicate source?

No strong opinions. Maybe for now let's just start a gh issue and gather thoughts there. I have many questions re HOMBA..

dosumis:
This is current state of release:  https://alleninstitute.github.io/CCF-MAP/docs/HOMBA_ontology_v1.html

They have promised a link where we can get the latest JSON.  I suspect that the best we will get is a DOI wrapping GitHub releases (neither available yet).  If I could get them to maintain the PURL system they set up and then let rot, this would all be much simpler.

PR #3690, opened 2026-04-23
Update Claude and GitHub AI automation config

cmungall:
## Summary
- replace the legacy `ai-agent` workflow with a GitHub App + Claude Code OAuth based setup
- add a `claude-code-review` workflow for PR review automation
- allow `dragon-ai-agent` to be summoned by assignment, following the dismech pattern
- add a Stop hook in `.claude/settings.json` so automated runs are checked for actual GitHub-side deliverables before stopping

## Notes
- this keeps the existing `.github/ai-controllers.json` allowlist unchanged
- workflow and JSON syntax were validated locally
- the new automation still needs the `ai4c-agent` app installed on this repo and the matching secrets configured

## Required repo setup
- `AI4C_AGENT_APP_ID`
- `AI4C_AGENT_PRIVATE_KEY`
- `CLAUDE_CODE_OAUTH_TOKEN`

@dragon-ai-agent

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,14 +1,27 @@
 {
   "permissions": {
     "allow": [
-        "Bash",
-        "Edit",
-        "MultiEdit",
-        "NotebookEdit",
-        "FileEdit",
-        "WebFetch",
-        "WebSearch",
-        "Write"
+      "Bash",
+      "Edit",
+      "MultiEdit",
+      "NotebookEdit",
+      "FileEdit",
+      "WebFetch",
+      "WebSearch",
+      "Write"
+    ]
+  },
+  "hooks": {
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "agent",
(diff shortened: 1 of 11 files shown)


cmungall:
Addressed the two substantive  review items in the latest push.

What changed:
- Pinned the ODK container in  to  to match the repo's existing pinning pattern.
- Aligned  in  to  so the two new Claude workflows use the same setup action version.
- Kept , but changed the dispatch logic so synchronize runs do **not** honor a stale PR-body  mention.

Reasoning on the synchronize point:
- I want pushes to an open PR to rerun the workflow so workflow changes validate themselves on the PR branch.
- A pure  fix would only cancel duplicate runs **after** they were already dispatched.
- The better fix is to avoid redispatch in the first place: body mentions are honored on  and , comments are honored on comment events, and assignments still work explicitly, but  now acts as a validation pass rather than a fresh summon from old PR text.

So the workflow keeps the self-validation behavior we wanted, without repeatedly re-summoning the agent on every push to a mentioned PR.

@dragon-ai-agent

cmungall:
Addressed the two substantive `ai4c-agent` review items in the latest push.

What changed:
- Pinned the ODK container in `ai-agent.yml` to `obolibrary/odkfull:v1.6` to match the repo's existing pinning pattern.
- Aligned `astral-sh/setup-uv` in `ai-agent.yml` to `@v7` so the two new Claude workflows use the same setup action version.
- Kept `pull_request.synchronize`, but changed the dispatch logic so synchronize runs do **not** honor a stale PR-body `@dragon-ai-agent please ...` mention.

Reasoning on the synchronize point:
- I want pushes to an open PR to rerun the workflow so workflow changes validate themselves on the PR branch.
- A pure `concurrency` fix would only cancel duplicate runs **after** they were already dispatched.
- The better fix is to avoid redispatch in the first place: body mentions are honored on `opened` and `edited`, comments are honored on comment events, and assignments still work explicitly, but `synchronize` now acts as a validation pass rather than a fresh summon from old PR text.

So the workflow keeps the self-validation behavior we wanted, without repeatedly re-summoning the agent on every push to a mentioned PR.

@dragon-ai-agent

PR #3697, opened 2026-04-27
Add bulk NTR ROBOT template generation workflow

dosumis:
Four-stage pipeline for generating UBERON new term request ROBOT templates from HRA ASCTB unmapped term tables:

Stage 1 (generate_template.py): reads xlsx/csv input, classifies parent IDs (UBERON/FMA/ASCTB-TEMP), assigns UBERON:99xxxxx provisional IDs, writes initial ROBOT template TSV + error and candidate reports.

Stage 2 (group_terms_by_parent.py): groups template rows by parent and writes per-group JSON files for parallel subagent processing.

Stage 3 (ntr-term-researcher agent): resolves FMA/ASCTB-TEMP parents via OLS4, checks for existing UBERON matches, writes Aristotelian definitions from Wikipedia, resolves is_a vs part_of relationship types.

Stage 4 (merge_definitions.py): merges subagent outputs back into the template; appends confirmed/possible OLS4 matches to candidates report.

Template columns: ID, LABEL, Definition, def_xref (definition annotation), is_a, part_of, In_subset, Date, Contributor, Present_in_taxon, Wikipedia_image (foaf:depiction), xref (direct oboInOwl:hasDbXref for Wikipedia article URL + FMA ID).

Supporting agents/skills:
- ntr-term-researcher: Stage 3 subagent spec
- ontology-term-lookup: OLS4 structured search
- fetch-wiki-info: Wikidata + Wikipedia lookup
- .mcp.json: ols4, artl-mcp, playwright MCP servers


.claude/agents/ontology-term-lookup.md
```diff
@@ -0,0 +1,102 @@
+---
+name: ontology-term-lookup
```
cmungall, line 2: is this necessary if you already gave the MCP? Seems redundant.

However, it wouldn't be a bad idea to add some instructions (probably in CLAUDE.md) that tells it *not* to use this if looking up directly in uberon when it should just use the edit file

cmungall: also if we did want to do this, skill >> agent.

.claude/skills/fetch-wiki-info/SKILL.md
```diff
@@ -0,0 +1,145 @@
+---
+name: fetch-wiki-info
```
cmungall: seems pointless clutter, every agent knows how to search wikipedia

.claude/skills/fetch-wiki-info/SKILL.md
```diff
@@ -0,0 +1,145 @@
...
+description: Fetch structured and descriptive information from Wikidata and Wikipedia for any topic
+argument-hint: "[search term]"
+allowed-tools: mcp__playwright__browser_run_code
```
cmungall: wait... I assume this skill was cooked up by an agent. Using playwright to access something that has an API...

bulk_ntr_workflow/CLAUDE.md
```diff
@@ -0,0 +1,184 @@
+# NTR Workflow: UBERON New Term Request ROBOT Template Generator
```
cmungall, line 1: so this CLAUDE.md won't be read at startup, so the intent is a human

```
cd buld_ntr_workflow
claude
<claude add some terms...>
```

if so there should be a README for humans

but why not just make this a skill?

.mcp.json
```diff
@@ -0,0 +1,21 @@
...
+      "command": "uvx",
+      "args": ["artl-mcp"],
+      "tools": ["*"]
+    },
+    "ols4": {
+      "type": "http",
+      "url": "http://www.ebi.ac.uk/ols4/api/mcp",
+      "tools": ["*"]
+    },
+    "playwright": {
```
cmungall, line 13: sure about this?

.claude/skills/fetch-wiki-info/SKILL.md
dosumis: Changing to use wikipedia API.  Used playwright because webfetch blocked.

bulk_ntr_workflow/CLAUDE.md
dosumis: Added README.md.  For now I will keep this as something run locally but may migrate to regular workflow to run from tickets in future.  Not sure about skill because has quite a bit of orchestration.

PR #3710, opened 2026-05-21
Update Biomappings

### LINKED ISSUE #3709: Update Biomappings ingestion
The Biomappings SSSOM artifact at https://w3id.org/biopragmatics/biomappings/sssom/biomappings.sssom.tsv now contains full metadata as a header. It would be nice to update the following section accordingly (I would also like to model other repositories on the workflow here, or develop a nice abstraction over this to make it easier to import SSSOM files into ODK repos, in general)

https://github.com/obophenotype/uberon/blob/9b71d7f349d6947378383b15b5c85436dff60ee7/src/ontology/uberon.Makefile#L1267-L1288

cthoyt:
Closes #3709 

This PR does two things:

1. Biomappings now has full inline metadata, so this PR removes vestigial artifact processing from the Makefile 
2. Update to newer version of Biomappings. For example, https://github.com/biopragmatics/biomappings/pull/266 also adds some more curations, which increases the amount of metadata available. This is why every line changes in the diff, since some new columns got added.

### DIFF
--- a/src/mappings/biomappings.sssom.tsv
+++ b/src/mappings/biomappings.sssom.tsv
@@ -6,285 +6,290 @@
 #  mesh: http://id.nlm.nih.gov/mesh/
 #  orcid: https://orcid.org/
 #mapping_set_id: https://w3id.org/biopragmatics/biomappings/sssom/biomappings.sssom.tsv
-#mapping_set_version: 0.5.0-dev-528eae3f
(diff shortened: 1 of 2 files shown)


matentzn (approved):

matentzn:
Thank you @cthoyt!

PR #3745, opened 2026-07-28
Review as ai4c-reviewer, not ai4c-agent

cmungall:
Switches `.github/workflows/claude-code-review.yml` from the `ai4c-agent` app to the separate `ai4c-reviewer` app (app id `4388239`, bot user `308941448`), matching geneontology/go-ontology#32365.

## Why

The review workflow authenticated as `ai4c-agent` — the same app that *opens* PRs in this repo. An author approving its own work isn't a review, and under required-approval branch protection it would be actively harmful: agent PRs could self-clear the gate. Reviews now come from a distinct identity.

`allowed_bots` **deliberately still lists `ai4c-agent`**. That input names the PR *author* to accept, not the reviewer identity, and reviewing the editing agent's PRs is the main reason this workflow exists. I clarified the comment there, since "the ai4c GitHub App" is now ambiguous between two apps.

## ⚠️ This is a hard cutover — setup required before merge

I checked this repo's secrets. Currently present:

```
AI4C_AGENT_APP_ID, AI4C_AGENT_PRIVATE_KEY, ANTHROPIC_API_KEY,
CLAUDE_CODE_OAUTH_TOKEN, GH_TOKEN, ID_ALLOCATION_TOKEN,
ONTOBOT_TOKEN, PAT_FOR_PR, TAVILY_API_KEY
```

No `AI4C_REVIEWER_*`. So on merge, review stops working until:

- [ ] `ai4c-reviewer` is installed on `obophenotype/uberon` with Contents **read**, Pull requests **write**, Issues **read**
- [ ] Secret `AI4C_REVIEWER_APP_ID` added (`4388239`)
- [ ] Secret `AI4C_REVIEWER_PRIVATE_KEY` added (full PEM)

(I can only see repo-level secrets — if these are configured at the `obophenotype` org level, they may already resolve and there's nothing to do.)

To make that failure legible rather than mysterious, a preflight step now checks both secrets are non-empty and fails with the actual cause, the fact that the reviewer app is separate from the agent, and the app id. `create-github-app-token` on an empty `app-id` otherwise fails in a way that reads as a token fault. Tested both branches locally: missing secrets exit 1 with the annotation, present secrets exit 0.

## Verification

- YAML parses; step order confirmed (preflight sits before token generation)
- Preflight guard executed directly under bash in both states
- No `ai4c-token` / `AI4C_AGENT_*` references remain in the file
- `actionlint` not run (not installed locally) — no Actions-schema check

## Deliberately not changed

The review prompt and criteria are untouched — this PR is only the identity switch. Two things I noticed while in here but left alone, happy to do either as a follow-up:

1. **The report is written twice.** `track_progress: true` posts the agent's final response as a tracking comment, and the prompt separately requires a checklist and summary in the `gh pr review` body. Curators get two copies that drift as the second is condensed. This was fixed in go-ontology by making the review body a three-line stub (verdict, counts, link) with the full report living in the tracking comment.
2. **`ANTHROPIC_API_KEY` isn't passed as a fallback.** It exists in this repo. Without it, an exhausted `CLAUDE_CODE_OAUTH_TOKEN` silently no-ops review on every open PR at once.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,3 +1,14 @@
+# AI review of pull requests.
+#
+# Authenticates as the `ai4c-reviewer` GitHub App, NOT `ai4c-agent`. The two are
+# deliberately separate identities: ai4c-agent authors PRs, so having it also
+# approve them would make the approval meaningless. Reviews now come from
+# ai4c-reviewer[bot] (app id 4388239, bot user 308941448).
+#
+# Required secrets:
+# - AI4C_REVIEWER_APP_ID
+# - AI4C_REVIEWER_PRIVATE_KEY
+# - CLAUDE_CODE_OAUTH_TOKEN
 name: Claude Code Review
 
 on:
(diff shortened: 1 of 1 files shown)


cmungall:
Setup is now complete, which resolves the one blocking finding.

`AI4C_REVIEWER_APP_ID` and `AI4C_REVIEWER_PRIVATE_KEY` are present on this repository and the `ai4c-reviewer` app is installed. The evidence is this PR's own review: run [30371566148](https://github.com/obophenotype/uberon/actions/runs/30371566148) completed successfully through the preflight check, token generation, and the review step, and the review above was posted by `ai4c-reviewer[bot]` — not `ai4c-agent[bot]`. That end-to-end path is only reachable with the app installed and both secrets valid, so the review's existence disproves its own blocking condition.

Worth noting the preflight guard earned its place before that: the [first run](https://github.com/obophenotype/uberon/actions/runs/30371566148) on this branch failed at step 3 with

> ai4c-reviewer not configured — Missing secret(s): AI4C_REVIEWER_APP_ID AI4C_REVIEWER_PRIVATE_KEY. This workflow authenticates as the ai4c-reviewer GitHub App, which is separate from ai4c-agent. Install the app on this repository and add the secrets; app id is 4388239.

rather than the opaque token error that would otherwise have surfaced.

The two deferred follow-ups (single report instead of two; `ANTHROPIC_API_KEY` fallback) and the reviewer's suggestion to preflight `CLAUDE_CODE_OAUTH_TOKEN` as well are all still worth doing, and both fixes already exist in geneontology/go-ontology#32365 to port across. Keeping them out of this PR so the identity switch stays reviewable on its own.

PR #3751, opened 2026-08-09
Prune non-relevant taxa from taxon-specific subsets.

### LINKED ISSUE #3748: Taxon-specific ontology generation
Hello Uberon team,

I would like to generate taxon-specific ontologies (for several species but my tests are done with human).

The online human-view contains GCAs that are specific to Rodentia. For example: 
```
relationship: part_of UBERON:0001883 {gci_relation="part_of", gci_filler="NCBITaxon:9989", source="MP"} ! olfactory tubercle
```
I also generated the ontology using the `precise` parameter (by modifying `uberon.Makefile`), but the result is the same.

So, I planned to implement the taxon-specific ontology generation using OWLAPI, but first I wanted to check the ELK inferences in Protégé.
Using ELK and the DL Query tab, I tested the following expressions: `'island of Calleja' and ('in taxon' some 'Homo sapiens')` or`'island of Calleja' and ('part of' some Primates)`.
	
However, I do not obtain 'nucleus accumbens' as inferred superclass, even though the ontology contains the following GCA:
```
relationship: part_of UBERON:0001882 {gci_relation="part_of", gci_filler="NCBITaxon:9443", source="MP"} ! nucleus accumbens
```
    
My questions are:

1. Is it expected that the human-view still contains rodent-specific GCAs?
2. What is the recommended approach for generating taxon-specific ontologies?

Thank you for your help.

Best regards.

gouttegd:
The method that generates the taxon-specific subsets (or "views") leaves the entire NCBITaxon:* tree untouched. For example, a taxon-specific subset still contains the same NCBITaxon:* tree than the entire Uberon ontology.

As a result, general class axioms (GCAs) that are completely irrelevant to the taxon for which the subset is generated are still present in the subset.

This PR updates the pipeline that creates the taxon-specific subsets so that the NCBITaxon:* is pruned of all the non-relevant taxa (all the taxa that are neither ancestors nor descendants of the taxon for which the subset is generated).

Partially fixes #3748

### DIFF
--- a/src/ontology/uberon.Makefile
+++ b/src/ontology/uberon.Makefile
@@ -892,6 +892,7 @@ subsets/%-view.owl subsets/%-tags.ofn: $(POSTPROCESS_SRC) | all_robot_plugins
 		                              --taxon $(TAXON_ID_$*) \
 		                              --strategy $(TAXON_SUBSET_STRATEGY) \
 		                              --reasoner ELK \
+		                              --prune-taxa \
 		                              $(foreach root,$(TAXON_SUBSET_ROOTS),--root $(root)) \
 		                              --prefix 'uberon: http://purl.obolibrary.org/obo/uberon/core#' \
 		                              --subset-name uberon:$*_subset \
--- a/src/ontology/Makefile
+++ b/src/ontology/Makefile
@@ -10,7 +10,7 @@
 # More information: https://github.com/INCATools/ontology-development-kit/
 
 # Fingerprint of the configuration file when this Makefile was last generated
-CONFIG_HASH=                91414c1a5f963a19baa027e134f3e31fdf6ee4b115f70a6cdeec15180d073d89
+CONFIG_HASH=                d12abab42ccf603187deaa68c3c631ab20a1469ceffe0d5e4092d71c12b9dbf3
 
 
 # ----------------------------------------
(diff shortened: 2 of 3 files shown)


gouttegd:
> No obvious CI, auth, or workflow regression

You have to love the confidence of the large crap generator that can say that even before the CI workflow has finished to run…

src/ontology/uberon.Makefile
```diff
@@ -892,6 +892,7 @@ subsets/%-view.owl subsets/%-tags.ofn: $(POSTPROCESS_SRC) | all_robot_plugins
 		                              --taxon $(TAXON_ID_$*) \
 		                              --strategy $(TAXON_SUBSET_STRATEGY) \
 		                              --reasoner ELK \
+		                              --prune-taxa \
```
matentzn, line 895: Are  never_in_taxon axioms pointing at removed taxa dropped as well?

matentzn (approved):
I assume you checked the diff on subsets/human-view.owl for example manually to convince yourself the change works :)

src/ontology/uberon.Makefile
gouttegd: No. This is because the axioms involving `never_in_taxon` are annotation assertion axioms, not logical axioms (remember that `never_in_taxon` is an _annotation property_, not an _object property_), so the OWLAPI’s `OWLEntityRemover` will not remove any such axiom even if it happens to point to an entity to remove (the annotation will remain, with a IRI value pointing to something that exists outside of the ontology).

I could add code behind the `--prune-taxa` option to explicitly also remove the annotation assertion axioms, but I think keeping those annotations is in fact a good thing.

PR #3757, opened 2026-08-21
Replace 'only in taxon' by 'in taxon' in bridges.

gouttegd:
The bridging axioms in the bridges to taxon-specific ontologies normally use RO:0002162 ('in taxon'), but some bridges also include a custom bridging axiom between Uberon’s 'anatomical entity' and the corresponding term in the foreign ontology. Those custom axioms are set forth in the `src/ontology/bridges/bridges.dispatch` file, and they currently use (for historical reasons) RO:0002160 ('only in taxon').

The RO:0002160 property is obsolete (or [about to be](https://github.com/oborel/obo-relations/pull/892)), so this PR replaces it with RO:0002162 (as for all other bridging axioms).

### DIFF
--- a/src/ontology/bridge/bridges.dispatch
+++ b/src/ontology/bridge/bridges.dispatch
@@ -34,7 +34,7 @@ file: cl-bridge-to-dhba.owl
 
 [ehdaa2-uberon]
 file: uberon-bridge-to-ehdaa2.owl
-add-axiom: (UBERON:0001062 and (BFO:0000050 some NCBITaxon:9606)) SubClassOf: RO:0002160 some NCBITaxon:9606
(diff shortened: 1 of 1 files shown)


balhoff:
@ai4c-reviewer is incorrect; the point is that 'only in taxon' is not logically stronger than 'in taxon'.

matentzn (approved):

gouttegd:
> RO:0002160 ('only in taxon') is logically stronger (universal restriction) than RO:0002162 ('in taxon', existential).

In other words, and for all the claims to the contrary, LLMs are nothing more than stochastic parrots. That one saw the word “only” in the label of RO:0002160, established a link with the keyword `only` used in Manchester syntax to express universal restrictions, and from there decided that RO:0002160 was a universal restriction. Completely ignoring that this makes actually no sense, since (i) a property is not in itself an universal restriction or an existential restriction, and (ii) the label of a property has no logical meaning.

PhD-level intelligence, they said…

Issue #3769, opened 2026-09-14
Missing has_part relations to CL: systematic scan of term definitions (33 evidence-backed candidates)


mellybelly:
Following the pattern of #2904 and #2960, this is a proposal list of `has_part` relations from Uberon structures to CL cell types that are asserted in a term's own textual definition but missing as an axiom.

No ontology changes are being proposed in a PR here — this ticket is the report, so the list can be reviewed (or cherry-picked) before anything is added.

## How the candidates were found

1. Parsed all 15,621 non-obsolete `[Term]` stanzas in `src/ontology/uberon-edit.obo`.
2. Indexed every non-obsolete CL label and synonym from `src/ontology/imports/local-cl.owl` (3,219 classes, 5,413 label/synonym strings).
3. Matched multi-word CL names inside each term's textual definition, discarding pairs where the `has_part` already exists.
   → **113 candidate subjects**.
4. Triaged against the anti-patterns described in #2963 and against the 247 `has_part CL:*` axioms already in the edit file.
5. Found one reference per assertion and verified every PMID against Europe PMC.

## Proposed axioms (33)

Each would carry a `{source="PMID:…"}` qualifier, matching the evidence-backed sub-pattern already used for e.g. `islet of Langerhans` and `olfactory epithelium`.

**Retina**
- [ ] UBERON:0001791 inner nuclear layer of retina → CL:0000636 Mueller cell — PMID:16839797
- [ ] UBERON:0001792 ganglionic layer of retina → CL:0000740 retinal ganglion cell — PMID:23083731

**Cerebral cortex** (matches the existing `cortical layer III` / `cortical layer V` axioms)
- [ ] UBERON:0005391 cortical layer II → CL:0000598 pyramidal neuron — PMID:25622573
- [ ] UBERON:0005395 cortical layer VI → CL:0000598 pyramidal neuron — PMID:25622573

**Hippocampus**
- [ ] UBERON:0002313 hippocampus pyramidal layer → CL:1001571 hippocampal pyramidal neuron — PMID:2687721
- [ ] UBERON:0014548 pyramidal layer of CA1 → CL:1001571 hippocampal pyramidal neuron — PMID:2687721
- [ ] UBERON:0014549 pyramidal layer of CA2 → CL:1001571 hippocampal pyramidal neuron — PMID:2687721
- [ ] UBERON:0014550 pyramidal layer of CA3 → CL:1001571 hippocampal pyramidal neuron — PMID:2687721
[...]

Issue #2963, opened 2023-07-07
Add design pattern docs for has-part to CL


cmungall:
We should have a DP to document has-part to CL. This would primarily be a documentative DP as it would not have an equivalence axiom.

We would document anti-patterns. The primary antipattern is making a has-part between a *broad* structure and a *specific* cell type. These are likely to lead to taxonomic incoherencies. E.g. "digestive tract subClassOf has-part some goblet cell" which would be false as DT encompasses almost all metazoa.

There are two sub-patterns:

1. trivially true reciprocals (e.g. smooth muscle tissue of bronchiole has-part bronchiolar smooth muscle cell)
2. evidence-backed knowledge (e.g. lung alveoli have type 1 and type 2 pneumocytes) 

You can see examples of both here:

 * #2960

I don't know if we want to combine the sub-patterns in one yaml or have two (if we move to linkml for DPs then we would group them).

There are some nuanced OWL/logic issues when making these kinds of assertions. I think the first is notable but likely not a major issue.

1. when the subject is a "portion" term like epithelium, there is a danger of making incorrect assertions. E.g if Foo-epithelium has Foo1 and Foo2 cells, it may be possible to slice out our find a natural (possible small) stretch of Foo epithelium that lacks either Foo1 or Foo2 cells. But I don't think this is a major concern. We just say that to be true Foo epithelium you need both Foo1 and Foo2 cells, our excised slices above are just connected epithelial cells derived from Foo.
2. coreference issues
3. lack of entailments using EL. Let's say I add an axiom "brain has-part some neuron" to uberon. Then later on a CL editor adds a trivial class "brain neuron". My original uberon assertion is correct but confusingly incomplete. We would like to update it to "brain has-part some brain neuron" but we don't (I think??) get this entailment with Elk?

The other thing we should bear in mind when making these is the principle of "reasonable completeness". See [open world considered harmful](https://douroucouli.wordpress.com/2020/09/04/the-open-world-assumption-considered-harmful/). It could be confusing for users if they are led to believe they can "slice open" e.g. an alveolus and see all the cell types necessarily present, but when slicing open other structures they are curiously empty (by slice open here I mean a query `<structure> has-part some ?x` which can be done in ubergraph or with OAK). For these kinds of queries to be complete we would likely need to combine taxonomic variability GCIs with the has-part pattern

cmungall:
[...]

PR #3770, opened 2026-09-14
docs: curation report for the has_part-to-CL definition-mining scan

mellybelly:
Report for #3769.

**Documentation only — this PR changes no ontology files.** Nothing under `src/` is touched, so no axioms are added, removed or altered by merging it.

## What this adds

| File | Change |
|---|---|
| `docs/reports/has_part_to_cl_scan_2026-09.md` | new, 265 lines |
| `mkdocs.yaml` | +2 lines, a nav entry under "Uberon Internal Docs" → "Curation reports" |

## Why

A scan of all 15,621 non-obsolete terms in `uberon-edit.obo` turned up 113 terms whose textual definition asserts cellular composition but which carry no corresponding `has_part` axiom to a CL class. Triaging those against the anti-patterns in #2963 took a fair amount of case-by-case reasoning, and most of that reasoning is about candidates that should *not* be added. That is the part worth writing down: without it, the next person to run a similar scan re-derives the same ~80 rejections from scratch.

## What the report contains

- **Method**, with enough detail to reproduce the scan using only a `robot` jar, plus the input snapshot it was run against (`uberon-edit.obo` at `f060b3e`, `cl-base.owl` release `2025-07-30`).
- **33 proposed axioms**, grouped by organ system, each with a PMID verified against Europe PMC, and a citation table for all 18 references.
- **The ~80 rejected candidates, with the reason for each**: broad structure paired with a specific cell type; matches that name a cell *part* rather than a whole cell (axon tracts, dendritic fields, synaptic structures); structures that surround rather than contain the cell; subjects already covered by `composed_primarily_of` or by a superclass; and the deliberate `overlaps ... {notes="soma"}` modelling on `outer nuclear layer of retina`, which is the right treatment for a layer holding somata rather than whole cells.
- **10 deferred candidates**, each with the specific open question — e.g. `glomerular epithelium` also subsumes the podocyte-free parietal epithelium; there is no CL class for a spiny stellate cell or a hilar mossy cell; Paneth cells are absent in several mammals, so a taxon GCI may be preferable to a plain assertion.
- **Why the UBERON→UBERON variant is not proposed**: 788 subjects / 1,359 pairs, but the curator SOP prefers `part_of` for anatomy-to-anatomy links and the match quality is poor (`mouth` → `digestive tract`, `tube` → `anatomical structure`).

## Checks

- `mkdocs.yaml` parses and the new nav entry resolves; the one relative link in the report (`../uberon-editor-sop.md`) points at `docs/uberon-editor-sop.md`. `mkdocs` is not installed in this environment, so the site build itself was not run locally.
- No files under `src/` in the diff — `git diff --name-only master... -- src/` is empty.

## Follow-up, not part of this PR

The 33 proposed axioms are implemented on [`claude/uberon-has-part-cl-report`'s sibling branch `claude/uberon-missing-has-part-jlis59`](https://github.com/obophenotype/uberon/tree/claude/uberon-missing-has-part-jlis59) (+33 lines in `uberon-edit.obo`, round-trip stable, ELK-clean, no unsatisfiable classes). That branch has no PR open; happy to raise one, or to trim it to whichever subset survives review of #3769.

@ai4c-agent

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018SafyR4sVmrWYzfV91tJ1p

---
_Generated by [Claude Code](https://claude.ai/code/session_018SafyR4sVmrWYzfV91tJ1p)_


matentzn (commented):
What precisely is the rationale to have this kind of information in git rather than simply as GitHub issues?

cmungall:
> What precisely is the rationale to have this kind of information in git rather than simply as GitHub issues?

I think the 3 of us + @dosumis should have a conversation about the general strategy re issues vs docs

...but in this case it's not so important we can leave this as is for now and just start doing the work that is outlined, IMO it all makes sense, let's start making PRs!

